Issue #36, opened 2023-03-29
VIM Client


wsxiaoys:
It'll be nice if we could support completion in vim.

Maybe starts with an omni func: https://vim.fandom.com/wiki/Omni_completion

wsxiaoys:
Related: https://github.com/github/copilot.vim

tricktux:
It would be amazing if you support `neovim`

[This](https://github.com/hrsh7th/nvim-cmp) is the most popular code completion engine on `neovim`

There's also [copilot.lua](https://github.com/zbirenbaum/copilot.lua)

Thanks for the excellent work!

nzlov:
Try [cmp-tabby](https://github.com/nzlov/cmp-tabby)

wsxiaoys:
> Try [cmp-tabby](https://github.com/nzlov/cmp-tabby). But it's Implemented very simple, i need full doc of tabby.
> 
> When callback event has error:
> 
> ```
> curl -X POST http://192.168.1.109:5000/v1/events -H 'Content-Type: application/json' --data '{
>     "type":"completion",
>     "completion_id":"cmpl-GUyycQ60ls5nq23vLELW5RJMntRhU",
>     "choice_index":0
> }'
> "invalid event"
> ```

Thank you for your contribution! @icycodes has already provided a Neovim-compatible extension on https://github.com/TabbyML/tabby/tree/main/clients/vim that you might want to try out.

wsxiaoys:
I am closing this issue for now. If you encounter any further issues related to the Vim extension, please feel free to open a new issue

PR #63, opened 2023-04-08
Add VIM client.

icycodes:
#36  

A first version support VIM 9.0+, NeoVIM is not supported at the moment.   

* Use `+job` feature to start a Node process for http requests. (Could be implemented in VIM script, but use Node make it easier for now. Have not yet decided whether to reuse these code with vscode client or not.)  
* Use `+textprop` feature to show inline code suggestion. 

---

Tested on: 
```
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Feb 07 2023 11:35:56)
macOS version - x86_64
Included patches: 1-1276
Compiled by Homebrew
Huge version with MacVim GUI.
```


wsxiaoys:
LGTM

Issue #52, opened 2023-04-06
VSCode Extension should handle unsupported language


wsxiaoys:
When a languageId is not presented in openai specs (http://localhost:5000/openai.json), it should omit the language type to use the fallback setting (unknown).

PR #75, opened 2023-04-10
Fix vscode extension handle unsupported language.

icycodes:
#52 

Hard coded supported language.  
Do we need an API like `/v1/supported_languages` ?


wsxiaoys:
You don’t have to hard code it. It can be inferred fromhttp://localhost:5000/openai.json

PR #84, opened 2023-04-11
fix link to vim extension

neowit:
original version of link to vim extension in README file pointed to vscode.


wsxiaoys:
Thanks for the PR, nice catch

neowit:
Thank you for including vim support!

Issue #66, opened 2023-04-08
Apple Silicon Support ( w/ CPU inference)


wsxiaoys:
Currently, we only support inference on CUDA. However, it would be advantageous to expand our support to include M1/M2, as many developers work on Mac laptops.

We'd like to revisited the feature once ggml became stable

References
1. https://github.com/ggerganov/ggml
2. https://github.com/NolanoOrg/cformers

Please reply with a 👍 if you want this feature.

cwwmbm:
Definitely +1 on the M1 support!

wsxiaoys:
related: https://github.com/mlc-ai/mlc-llm

wsxiaoys:
Hello!

Thank you for your message. I wanted to let you know that I have just added the very first version of Apple Silicon support. If you're interested, I would appreciate it if you could give it a try by following these steps:

1. Download the `tabby_aarch64-apple-darwin` file from this link: https://github.com/TabbyML/tabby/suites/13205372288/artifacts/719099228.
2. Execute the following commands in your terminal:

```bash
# Download the model locally.
tabby download --model TabbyML/J-350M

# Serve the model.
PORT=8080
tabby serve --model TabbyML/J-350M --port $PORT
```

After completing these steps, you can configure the API endpoint to http://localhost:8080/ in the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=TabbyML.vscode-tabby).

I would be very interested in hearing about your experience with the new feature and learning from your feedback. Thank you in advance!

wsxiaoys:
FYI, just added automatic release process at: https://github.com/TabbyML/tabby/releases/tag/latest

chanansh:
followed the instruction above. The server is running and `Listening at 0.0.0.0:8080`
the vscode extension is installed and by default configured to endpoint http://localhost:8080
however, I don't get any sweet suggestions :-(

wsxiaoys:
Hi @chanansh 

Could you please share some examples of bad suggestions that you have received?

chanansh:
I don't get suggestions at all. That's the issue.
When I logged in to the server via browser it seems to work. I guess the client in the vscode extension is faulty (p.s. there is no validation "connected to server successfully"). Or,.maybe the inference is too slow.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

wsxiaoys:
[...]

PR #86, opened 2023-04-11
M1 Mac support

glennfu:
I have this project now working on my M1 Mac well enough that I can visit `http://127.0.0.1:8081/#/default/completions_v1_completions_post` and do a successful test post of the binarysearch completions example. Skimming the logs, the only clear error remaining is this:
```
2023-04-11 12:12:00,170 DEBG 'dagu_scheduler' stdout output:
2023/04/11 12:12:00 Failed to start DAG: fork/exec ./tabby/tools/analytic/main.sh: no such file or directory
```

Some notes:
* My strategy was to avoid Docker since the benefit of Metal would be lost unless I ran this natively
* I added a few instruction changes to the README.md that I used to get things set up on my local machine
* I had to replace the absolutely path of `/data` with the relative path of `data` in many instances. I don't know how to write that in a backward compatible way.
* I added `torchvision` and `supervisor` to `pyproject.toml`. I experimented with some ways to make it ONLY install on the Mac to not affect the original Dockerfile behavior but couldn't get it to work. I'm a Python novice and today is the first time I even heard of `poetry` so it's probably entirely possible for someone else to fix.

I hope this helps you speed this project along! Like I said, I'm a Python novice, but I'm very interested in this project and eager to assist. Let me know what else I can do to help.


glennfu:
Oh I forgot one more thing: I changed to port 5001 because there's a core macos process that runs on 5000 and conflicts with this.

wsxiaoys:
Thank you for submitting the PR. It provides a great reference point for the mps device support. However, we won't be able to merge it.

To achieve complete support, adding an arm docker image would be better, and we should remove all the fine-tuning stuff as it's unlikely that we'll fine-tune on a Mac client.

Nevertheless, we appreciate your contribution and will mark this PR as draft for referencing purposes.

glennfu:
If the goal is to have performant M1 Mac support, and not just slow emulation, I don't believe it will be possible to achieve that using Docker.

wsxiaoys:
implemented in #66

PR #94, opened 2023-04-12
Updates README.md to include a tabby.toml example

bschrib:


bschrib:
I'll revise this, now that I'm more familiar with this it appears that much more work is needed as the dataset must be processed.

wsxiaoys:
Thank you for your pull request. The reason why I have not documented this feature is because the fine-tuning and retrieval-based finetuning are still under development. Therefore, I would prefer not to merge the PR at this time.

wsxiaoys:
related TAB-17

PR #104, opened 2023-04-14
Update README.md

nzlov:
add docker compose


wsxiaoys:
Thanks for the PR, while you might noticed we provided some compose based orchestration in https://github.com/TabbyML/tabby/tree/main/deployment

We plan to document it once we have a turn-key deployment story for public cloud - e.g AWS

PR #112, opened 2023-04-20
update huggingface_gptneox_convert.py

AkiyamaYummy:
Fix:

1. Using tensor parallelism requires dividing some of the bias by the factor.
2. HF will start multiprocessing first if multi `.bin` files are needed to load, causing the "context has already been set" problem.

### DIFF
--- a/tabby/tools/converter/huggingface_gptneox_convert.py
+++ b/tabby/tools/converter/huggingface_gptneox_convert.py
@@ -2,7 +2,6 @@ import argparse
 import configparser
 import multiprocessing
 import os
-import shutil
 from pathlib import Path
 
 import numpy as np
@@ -139,7 +138,9 @@ def split_and_convert(args):
     ]
 
     huggingface_model_file_list = [
-        hf_file_name for hf_file_name in os.listdir(args.in_file) if hf_file_name.endswith(".bin")
+        hf_file_name
+        for hf_file_name in os.listdir(args.in_file)
+        if hf_file_name.endswith(".bin")
     ]
     if len(huggingface_model_file_list) > 1:
         multiprocessing_context = multiprocessing.get_context()
--- a/tabby/tools/converter/huggingface_gptneox_convert.py
+++ b/tabby/tools/converter/huggingface_gptneox_convert.py
@@ -138,7 +138,9 @@ def split_and_convert(args):
         "mlp.dense_4h_to_h.weight",
     ]
 
-    huggingface_model_file_list = [__fn for __fn in os.listdir(args.in_file) if __fn.endswith(".bin")]
+    huggingface_model_file_list = [
+        hf_file_name for hf_file_name in os.listdir(args.in_file) if hf_file_name.endswith(".bin")
+    ]
     if len(huggingface_model_file_list) > 1:
         multiprocessing_context = multiprocessing.get_context()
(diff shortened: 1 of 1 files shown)


AkiyamaYummy:
@wsxiaoys 

Hi, looking forward to you taking a look.

tabby/tools/converter/huggingface_gptneox_convert.py
```diff
@@ -251,5 +245,7 @@ def split_and_convert(args):
         print("{}: {}".format(key, vars(args)[key]))
     print("========================================")
 
-    shutil.rmtree(args.saved_dir, ignore_errors=True)
-    split_and_convert(args)
+    __dir = os.path.join(args.saved_dir, "%d-gpu" % args.infer_gpu_num)
```
wsxiaoys, line 248: `__dir` -> `output_dir`

tabby/tools/converter/huggingface_gptneox_convert.py
```diff
@@ -145,8 +138,16 @@ def split_and_convert(args):
         "mlp.dense_4h_to_h.weight",
     ]
 
-    torch.multiprocessing.set_start_method("spawn")
-    pool = multiprocessing.Pool(args.processes)
+    huggingface_model_file_list = [__fn for __fn in os.listdir(args.in_file) if __fn.endswith(".bin")]
```
wsxiaoys, line 141: `__fn` > `fn`

tabby/tools/converter/huggingface_gptneox_convert.py
```diff
@@ -145,8 +138,16 @@ def split_and_convert(args):
...
+    if len(huggingface_model_file_list) > 1:
```
wsxiaoys, line 142: will checking `trained_gpu_num > 1`  be equivalent?

wsxiaoys (changes requested):
Thanks for the PR! I added a few nits and questions. Please also fix the lint.

tabby/tools/converter/huggingface_gptneox_convert.py
AkiyamaYummy: fixed

tabby/tools/converter/huggingface_gptneox_convert.py
AkiyamaYummy: fixed

tabby/tools/converter/huggingface_gptneox_convert.py
AkiyamaYummy: I don't think there should be any difference in huggingface model files, whether trained on a single gpu or multiple gpus. The presence of multiple .bin files in the model files is due to the save rule implemented to prevent the model file from becoming too large. ([huggingface's documentation](https://huggingface.co/docs/transformers/v4.28.1/en/main_classes/model#transformers.PreTrainedModel.save_pretrained.max_shard_size))

wsxiaoys:
Please fix pre-commit.

Otherwise LGTM

wsxiaoys (changes requested):
Please fix pre-commit.

Otherwise LGTM

AkiyamaYummy:
> Please fix pre-commit.
> 
> Otherwise LGTM

fixed

Issue #127, opened 2023-05-11
When input characters according to the displayed inline suggestion, the remaining part should keep unchanged.


icycodes:
In the case of when some inline suggestion has shown up, and i do not use `<Tab>` to accept full suggestion. Instead, I input characters according to suggestion, or use `Accept Word`(`<Ctl+→>`) or `Accept Line` in VSCode, the remaining part of inline suggestion should keep unchanged.

---
Please reply with a 👍 if you want this feature.

icycodes:
Completed in #158

PR #128, opened 2023-05-11
VSCode extensions: add completion cache.

icycodes:
#127 

Add a LRU cache to save recently completions. 

* When a completion returned from server and will show up, add it to cache.
* If prompt text range remains unchanged, and inputed characters after prompt text range are matched with any choice of a cached completion, use cached completion instead of making new request.


wsxiaoys:
LGTM

PR #130, opened 2023-05-11
small fix for completion cache

vodkaslime:
Small fix changing `>`  to `>=` to fix the bug that when making `add()` to an already full cache, the `evict()` and then `prepend()` would make the cache size to be `capacity+1`.

Also some small renaming for better readability.

Technically the evicting while loop should be fenced with `capacity != 0`. Now that it is hardcoded as a static value, it's fine so far. In the future when we decide to make it configurable, we need to add the fence in `add()`.

Also, would prefer classic LRU implementation with a hashmap together with a linked list. The current implementation's matching is a full O(n) complexity, and so is `refresh()`. Would be a nice-to-have to reduce the complexity with a custom hash function for `documentId`, `text` and `cursor`. Could leave it as a TODO.

### DIFF
--- a/clients/vscode/src/CompletionCache.ts
+++ b/clients/vscode/src/CompletionCache.ts
@@ -14,25 +14,23 @@ export type CompletionCacheEntry = {
 };
 
 export class CompletionCache {
-  public static cacheSize = 10;
+  public static capacity = 10;
   private cache = new LinkedList<CompletionCacheEntry>();
 
   constructor() {}
 
-  private evict() {
-    while (this.cache.length > CompletionCache.cacheSize) {
-      this.cache.removeTail();
-    }
-  }
-
-  private pop(entry: CompletionCacheEntry) {
+  private refresh(entry: CompletionCacheEntry) {
     this.cache.remove(entry);
     this.cache.prepend(entry);
   }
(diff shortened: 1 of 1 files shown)


icycodes:
Thank you for your PR!
1. That was my mistake. I think simply moving `evict()` after `prepend()` should solve this problem.
2. The renaming looks very good.
3. I believe that if we don't change `>` to `>=` to fix the bug, then this fence for `capacity != 0` is not necessary. We can also add configuration validation `> 0` for capacity if it becomes configurable in the future. And setting `capacity = 0` means not to use cache, which will break feature #127.
4. Certainly, this implementation is O(n), so I hard coded the cache capacity to a small value. The reason why I did not use a hash map LRU cache is that, this cache is not only designed to reduce duplicated completion requests, but to try to solve #127. In this case, text and curse changed, but a cached completion should be returned (after adjustment), so I think it is required to traverse the full list.

I am considering renaming this completion cache to something else to make it clear that it is not a client API request cache, but I am not sure yet.

vodkaslime:
Cool it makes sense. Updated the minimal fix for the bug.

icycodes (approved):

Issue #134, opened 2023-05-18
Vim plugin should not invoke completion request with cursor moving key strokes.


icycodes:
The Vim plugin currently invokes a completion request every time the cursor moves in INSERT mode. This can cause confusion when attempting to simply move the cursor through a word or code block without any intention of editing it.

Like other completion engines, the Vim plugin should not invoke a completion request under these circumstances.

icycodes:
#135 add a very simple rule to workaround
If there is no word (`\W` in regex) after cursor in the current line, schedule a completion request. Otherwise clear current completion and do not schedule a new completion request.

icycodes:
#139 changes scheduler trigger event from `CursorMovedI` to `TextChangedI`, cursor moving key strokes
will not trigger completion request.   
This solution should create a similar behavior to VSCode’s inline completion.

Also looking for standard solution for vim plugin but not found yet.
Github copilot vim plugin could also be triggered if i just move cursor.

PR #135, opened 2023-05-18
Vim plugin: only trigger completion request when no word after cursor in current line.

icycodes:
#134


wsxiaoys:
maybe just don’t generate suggestion if the key stroke is just to move the cursor?

icycodes:
Follow up in #139

PR #143, opened 2023-05-22
implement detect_gpu in tabby.sh

### LINKED ISSUE #140: Detect GPU type in tabby.sh

Detect GPU type in https://github.com/TabbyML/tabby/blob/main/tabby/scripts/tabby.sh , if GPU is older than Pascal, generate a warning and skip the triton

---

Iamshankhadeep:
fixes #140


wsxiaoys:
thanks for the contribution - while this is no longer applicable since we've migrated to an ctranslate2 based implementation, which covers a much wider GPU platforms.

PR #166, opened 2023-05-30
Add prefix / suffix interface

wsxiaoys:


crates/tabby/src/serve/completions.rs
```diff
@@ -11,12 +11,24 @@ mod languages;
...
+    prompt: String,
+
     /// https://code.visualstudio.com/docs/languages/identifiers
     #[schema(example = "python")]
     language: Option<String>,
 
-    #[schema(example = "def fib(n):")]
-    prompt: String,
+    /// When segments are set, the `prompt` is ignored during the inference.
+    segments: Option<Segments>,
```
icycodes, line 22: Filed `segment` is optional but `prompt` is not here. So if we want FIM, it is also required to have a `prompt` field with empty value in request.   
I think using a non-optional `prompt` (prefix) and an optional `suffix` field should be better.

wsxiaoys: as said in comment，when segments present，prompt field is ignored.

icycodes: LGTM

wsxiaoys:
close in favor of #183

PR #209, opened 2023-06-06
refactor: agent configuration interface

icycodes:
TAB-30


clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -0,0 +1,23 @@
+export type AgentConfig = {
+  server?: {
+    endpoint?: string;
+  };
+  logs?: {
+    level?: "debug" | "error" | "silent";
+  };
+  analytics?: {
```
wsxiaoys, line 8: Let's call this anonymousUsageTracking

clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -0,0 +1,23 @@
...
+    enabled?: boolean;
```
wsxiaoys, line 9: `enabled` -> `disabled`, means the anonymousUsageTracking is by default enabled.

wsxiaoys (changes requested):

wsxiaoys:
Otherwise LGTM

PR #216, opened 2023-06-07
fix: vim plugin inline completion display error if starts with newline.

### LINKED ISSUE #159: Vim Plugin: inline completion is not displayed correctly if the completion starts with a  `\n`
Vim plugin inline completion is not displayed correctly if the completion starts with a  `\n`.   
Contents should be displayed in secound line was displayed in current line in this case.

icycodes:
Fix TAB-28
Fix #159


wsxiaoys:
LGTM

PR #227, opened 2023-06-10
feat: add otlp-endpoint for OpenTelemetry support [TAB-67]

wsxiaoys:


crates/tabby/src/main.rs
```diff
@@ -64,3 +73,47 @@ macro_rules! fatal {
         })
     };
 }
+
```
vodkaslime, line 76: Wanna put this logging functionality into a separate file/module?

wsxiaoys: ack

Issue #232, opened 2023-06-12
docker compose v2 


Allesanddro:
On docker compose v2 the provided yml file is not supported.

This would  be the "correct" compose file for v2

```
version: '3.5'
services:
  tabby:
    restart: always
    image: tabbyml/tabby
    command: serve --model TabbyML/SantaCoder-1B --device cuda
    volumes:
      - "$HOME/.tabby:/data"
    ports:
      - 8080:8080
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: 1
              capabilities: [gpu]
```

wsxiaoys:
Hi @Allesanddro ,

Thanks for the feedback, nice catch! Would you mind submit a PR directly? You can generate a PR by uediting https://github.com/TabbyML/tabby/edit/main/website/docs/self-hosting/01-docker.mdx in browser directly.

PR #233, opened 2023-06-12
docs: Update 01-docker.mdx

Allesanddro:
migration from docker compose v1 to v2 

[Migrate to Compose V2](https://docs.docker.com/compose/migrate/)


wsxiaoys:
fix #232

wsxiaoys:
Thanks for the PR @Allesanddro !

PR #250, opened 2023-06-18
Fix incorrect tabby serve command

prologic:


wsxiaoys:
Thanks for contributing this fix @prologic

prologic:
@wsxiaoys No worries! I just installed Tabby locally on my Mac Studio (M1 Max) connecting it up to VSCodium and so far I'm pretty impressed actually. This is pretty damn cool and useful already!

PR #253, opened 2023-06-19
Update the clients/vim/README.md

pauchiner:
Added a section that explains how to install the extension on packer and lazy.


wsxiaoys:
Thanks for the PR @pauchiner !

PR #255, opened 2023-06-22
feat: agent enable per request prefix/suffix max lines configuration

ghthor:
# What Changed

1. bug: fix vim agent startup race condition
1. feat: add per request max lines config for prefix/suffix to vscode/vim
   clients

# Why

## 1. bug: vim agent startup race condition

At some point the nodejs tabby-agent was updated, but the compiled bundle for
the vim client was not updated. Once updating there was a race condition where
the initialize call from the vim client was failing because the agent hadn't
been bound to the StdIO object BEFORE StdIO started processing requests from
stdin.

## 2. feat: add per request max lines config for prefix/suffix to vscode/vim

I initially wanted to configure this amount to test if it had any effect on the
quality of the completions. I think this will be useful to test on different
models and making it easy to change without requiring a recompile of the
tabby-agent seems very useful.


clients/tabby-agent/src/cli.ts
ghthor, line 1: This was the race condition causing the vim client to fail to initialize the agent

wsxiaoys: nice catch! cc @icycodes

wsxiaoys (commented):
Thank you for the PR, @ghthor.

Overall, it looks good to me. However, I would like to bring your attention to a scheduled refactor for the developer options:

1. All options under `developerOptions` will no longer be configurable in IDE/Editor extensions. For example, `developerOptions.suggestionDelay` will be removed from the VSCode settings.

2. The Tabby agent will attempt to read these developer options from `~/.tabby/agent/config.toml`.

By implementing these changes, we aim to reduce the maintenance effort required for each IDE/Editor extension. Only the ability to enable or disable the options will be exposed.

wsxiaoys (approved):

icycodes (approved):
LGTM

ghthor:
> The Tabby agent will attempt to read these developer options from ~/.tabby/agent/config.toml.

I don't see any issue with that. I'd suggest adding a command we can send to the tabby-agent in addition to support for handling the SIGHUP signal that would trigger tabby-agent to re-read from this configuration.

PR #288, opened 2023-07-11
refactor: update yarn workspace struct and build scripts.

icycodes:
TAB-64


package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "private": true,
+  "workspaces": [
+    "clients/tabby-agent",
+    "clients/vscode",
+    "clients/vim"
+  ]
+}
```
icycodes, line 8: Add a workspace package.json in root dir.  
Is it better to move this file to ./clients?

wsxiaoys: in root is fine

PR #297, opened 2023-07-14
feat(vscode): add keybindings for accepting inline completions.

icycodes:
TAB-103

Settings preview:  

Screenshot from 2023-07-14 14-42-41


wsxiaoys:
shall we use tabby style as default？

icycodes:
> shall we use tabby style as default？

Sure

PR #314, opened 2023-07-27
feat(deployment): add deployment and pvc file to deploy the server on openshift

marcoklaassen:
feat(deployment): add deployment and pvc file to deploy the server on openshift

### DIFF
--- a/experimental/deploy/openshift-deployment.yaml
+++ b/experimental/deploy/openshift-deployment.yaml
@@ -1,0 +1,67 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: tabby-data
+spec:
+  accessModes:
+    - ReadWriteOnce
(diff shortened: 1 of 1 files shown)


wsxiaoys:
Hi @marcoklaassen 

I appreciate your effort in contributing to the project. However, at the moment, I would prefer to keep it as an open pull request for reference, rather than merging it into the main branch. This is because we are currently unable to support the deployment process for non-docker environments.

Once again, thank you for your contribution!

marcoklaassen:
Hi @wsxiaoys 

Thanks. No problem. I just went my first steps to deploy is on OpenShift and wanted to document my progress here, so anyone can use it to go forward. But I understand that you can not support any other deployment process. 
So, I'm fine with keeping this as an open pull request. 

Thanks, 

Marco

PR #325, opened 2023-08-02
fix: improve download logging

### LINKED ISSUE #313: Can't connect to tabby
**Describe the bug**
Following the Getting Started, I created a ~/tabby/config.toml file with the provided URL (copy and paste from the web page), then run the docker command for use on a CPU (not CUDA). The Docker image is downloaded and it looks like it is running (no errors shown). When I then try to open the web page in my browser (localhost:8080) it says there is no response. I checked for firewall rules, and even turned the firewall off, but it did not help.

Is there more needed for this to work than just creating the config file and running the Docker command?

Not using GPU.

**Additional context**
Host is a Linux Mint 20.3 machine.

wsxiaoys:
fix #313

### DIFF
--- a/crates/tabby-download/Cargo.toml
+++ b/crates/tabby-download/Cargo.toml
@@ -11,4 +11,5 @@ reqwest = { version = "0.11.18", features = ["stream", "json"] }
 anyhow = { workspace = true }
 serde = { workspace = true }
 serdeconv = { workspace = true }
+tracing = { worksapce = true }
 tokio-retry = "0.3.0"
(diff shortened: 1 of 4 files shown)


crates/tabby-download/Cargo.toml
```diff
@@ -11,4 +11,5 @@ reqwest = { version = "0.11.18", features = ["stream", "json"] }
 anyhow = { workspace = true }
 serde = { workspace = true }
 serdeconv = { workspace = true }
+tracing = { worksapce = true }
```
flavienbwk, line 14: Is there a typo here? (`worksapce` instead of `workspace`)

wsxiaoys: nice catch! fixing in pr https://github.com/TabbyML/tabby/pull/331

PR #357, opened 2023-08-16
feat: add model-converter [TAB-153]

vodkaslime:
# What this PR is about
We are building a one-click tool to convert models into ready-to-serve ctranslate2 models.

This PR is the phase 1 of the effort: building the backend tooling.

- check out source model from huggingface
- run command `ct2-transformers-converter --model ./ --output_dir ctranslate2 --quantization=float16 --trust_remote_code`
- run command `python -m transformers.convert_slow_tokenizers_checkpoints_to_fast --tokenizer_name $TokenizerName --checkpoint_name ./  --dump_path ./`
- push output to the hf model

Test running this backend tool with command `python ./main.py --model Salesforce/codet5p-220m --output_dir /tmp/model/cache --inference_mode seq2seq`, the converted model can be served by Tabby.


experimental/transformers-to-ctranslate/args.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    parser.add_argument(
+        "--model",
+        required=True,
+        help=(
+            "Name of the pretrained model to download, "
+            "or path to a directory containing the pretrained model."
+        ),
+    )
+    parser.add_argument(
```
wsxiaoys, line 16: do not make this an commandline interface, use its default value when converting.

experimental/transformers-to-ctranslate/args.py
```diff
@@ -0,0 +1,55 @@
...
+        "--activation_scales",
+        help=(
+            "Path to the pre-computed activation scales. Models may "
+            "use them to rescale some weights to smooth the intermediate activations "
+            "and improve the quantization accuracy. See "
+            "https://github.com/mit-han-lab/smoothquant."
+        ),
+    )
+    parser.add_argument(
```
wsxiaoys, line 25: simply copy all files from the origin model

experimental/transformers-to-ctranslate/args.py
```diff
@@ -0,0 +1,55 @@
...
+        help=(
+            "List of filenames to copy from the Hugging Face model to the converted "
+            "model directory."
+        ),
+    )
+    parser.add_argument(
+        "--revision",
+        help="Revision of the model to download from the Hugging Face Hub.",
+    )
+    parser.add_argument(
```
wsxiaoys, line 37: do not make this a commandline interface

experimental/transformers-to-ctranslate/args.py
```diff
@@ -0,0 +1,55 @@
...
+        action="store_true",
+        help="Enable the flag low_cpu_mem_usage when loading the model with from_pretrained.",
+    )
+    parser.add_argument(
+        "--trust_remote_code",
+        action="store_true",
+        help="Allow converting models using custom code.",
+    )
+
+    parser.add_argument(
```
wsxiaoys, line 48: can this be inferred？

experimental/transformers-to-ctranslate/main.py
```diff
@@ -0,0 +1,61 @@
...
+        force_download=True
+    )
+    convert_output_dir = os.path.join(model_path, "ctranslate2")
+
+    # Convert model into ctranslate
+    converter = TransformersConverter(
+        args.model,
+        activation_scales=args.activation_scales,
+        copy_files=args.copy_files,
+        load_as_float16=args.quantization in ("float16", "int8_float16"),
```
wsxiaoys, line 32: is `quantization` defined in argparser?

experimental/transformers-to-ctranslate/main.py
```diff
@@ -0,0 +1,61 @@
...
+    converter.convert(
+        output_dir=convert_output_dir,
+        vmap=None,
+        quantization=args.quantization,
+        force=True
+    )
+
+    # Convert model with fast tokenizer
+    os.chdir(model_path)
+    convert_slow_checkpoint_to_fast(
```
wsxiaoys, line 46: if a tokenizer filename exists, this step can be skipped.

experimental/transformers-to-ctranslate/main.py
```diff
@@ -0,0 +1,61 @@
...
+        tokenizer_name=args.tokenizer_name,
+        checkpoint_name='./',
+        dump_path='./',
+        force_download=False,
+    )
+
+    # Upload folder to huggingface
+    token = os.getenv("HF_TOKEN")
```
wsxiaoys, line 54: let's not uploading this to HF in this tool. Such functionality  should be provide separately.

wsxiaoys (changes requested):

experimental/transformers-to-ctranslate/args.py
vodkaslime: Can we make sure there is always a `tokenizer_config.json` file with a `tokenizer_class` field in it? If so, we could do it.

I checked out https://huggingface.co/stabilityai/stablecode-completion-alpha-3b-4k/blob/main/tokenizer_config.json and https://huggingface.co/Salesforce/codet5p-220m and found that they do have it. Want to make sure if this is some kind of routine. If yes, then we could infer from it.

experimental/transformers-to-ctranslate/main.py
vodkaslime: It is declared in https://github.com/TabbyML/tabby/blob/converter/experimental/transformers-to-ctranslate/args.py#L54, which is defined in https://github.com/OpenNMT/CTranslate2/blob/master/python/ctranslate2/converters/converter.py#L15

wsxiaoys: but you don't define it in arg parser right? so it will be `None`?

Neitherless, let's pass reasonble default to transformer converter, as long as the output file is stored as `float16`

experimental/transformers-to-ctranslate/main.py
```diff
@@ -42,20 +61,11 @@
...
 
     # Convert model with fast tokenizer
-    os.chdir(model_path)
-    convert_slow_checkpoint_to_fast(
-        tokenizer_name=args.tokenizer_name,
-        checkpoint_name='./',
-        dump_path='./',
-        force_download=False,
-    )
+    if not os.path.exists(os.path.join(model_path, "tokenizer2.json")):
```
wsxiaoys, line 60: should be `tokenizer.json`?

experimental/transformers-to-ctranslate/args.py
```diff
@@ -13,23 +13,6 @@ def make_parser():
...
-    )
-    parser.add_argument(
-        "--copy_files",
-        nargs="+",
-        help=(
-            "List of filenames to copy from the Hugging Face model to the converted "
-            "model directory."
-        ),
-    )
     parser.add_argument(
```
wsxiaoys, line 16: should this be given a reasonble default like `main`? or what if `None` is passed? will snapshot download api determine it automatically?

wsxiaoys (changes requested):

experimental/transformers-to-ctranslate/args.py
vodkaslime: yes snapshot download api automatically finds the newest.

https://huggingface.co/docs/huggingface_hub/guides/download#download-an-entire-repository

`downloads the latest revision by default. If you want a specific repository revision, use the revision parameter...`

wsxiaoys: in that case, let's remove this arg and pass `None` for now.

wsxiaoys:
Besides,

seems you're not generating `tabby.json`?

wsxiaoys (approved):

PR #359, opened 2023-08-17
feat(agent): monitor completion response time.

icycodes:
Complete TAB-165

---

Auto completion requests will expire in 5s by default. Tabby extension will show the notification if most completion requests timed out or average response time is too slow.

Screenshot from 2023-08-17 18-56-16
Screenshot from 2023-08-17 19-18-10

Screenshot from 2023-08-17 19-02-12
Screenshot from 2023-08-17 19-17-05

---
If running a large model ( >= 1B ) on cpu,  the message will be:
Screenshot from 2023-08-17 19-04-51

### DIFF
--- a/clients/tabby-agent/openapi/tabby.json
+++ b/clients/tabby-agent/openapi/tabby.json
@@ -37,17 +37,9 @@
         "responses": {
           "200": {
             "description": "Success",
-            "content": {
-              "application/json": {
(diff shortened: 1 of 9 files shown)


wsxiaoys:
Consider adjust wording based on `device` field in `/v1/health`?

PR #386, opened 2023-08-31
feat: map js,ts,jsx and tsx to js-ts as unified language [TAB-181]

vodkaslime:
# What this PR is about
`javascript`, `jsx`, `typescript` and `tsx` are interoperabe with each other. Therefore in indexing phase, we need to make inter-connection between these four languages. Therefore, after parsing phase (with tree sitter on each language), we unify the "language" tag as one: `js-ts`, so that prompt building functions would regard all the languages as `js-ts`. This language reduction works on both dataset (job1) and indexing (job2).


crates/tabby-scheduler/src/dataset.rs
```diff
@@ -72,6 +70,13 @@ fn get_language(ext: &OsStr) -> Option<&str> {
     };
 }
 
+fn reduce_language_if_needed(language: &str) -> &str {
+    return match LANGUAGE_REDUCE_MAP.get(language) {
```
wsxiaoys, line 74:
```rust
if language == "javascript" || language == "typescript" .... {
  "javascript-typescript"
} else {
  language
}
```

crates/tabby-scheduler/src/index.rs
```diff
@@ -62,6 +65,13 @@ pub fn index_repositories(_config: &Config) -> Result<()> {
     Ok(())
 }
 
+fn reduce_language_if_needed(language: &str) -> &str {
+    if ["javascript", "jsx", "typescript", "tsx"].contains(&language) {
```
wsxiaoys, line 69:
```suggestion
    if ["javascript", "jsx", "typescript", "tsx"].contains(&language) {
      "javascript-typescript"
    } else {
      language
    }
```

crates/tabby-scheduler/src/index.rs
```diff
@@ -46,10 +46,13 @@ pub fn index_repositories(_config: &Config) -> Result<()> {
                     continue;
                 }
             }
+
+            let language = reduce_language_if_needed(&doc.language);
+            print!("{}", language);
```
wsxiaoys, line 51:
```suggestion
```

wsxiaoys (changes requested):
Otherwise LGTM

wsxiaoys (approved):

PR #389, opened 2023-09-01
feat(agent): added adaptive completion debouncing.

icycodes:
Complete TAB-176

Changes Made:
- Removed the fixed request delay in ide plugin, and introduced adaptive completion debouncing in tabby-agent.
- Renamed the agent function `getComptions` to `provideCompletions` with debouncing functionality.

Completion debouncing can be configured in the `config.toml`.
```
[completion.debounce]
mode = "adaptive"

[completion.debounce]
mode = "fixed"
interval = 200
```

When adpative mode is activated, the debouncing delay is dynamically calculated based on the following factors:
- Avg keystroke interval
- Accept probability for current completion context, calculated from trigger character and suffix
- Avg completion response time of the tabby server


wsxiaoys:
Consider bump the `package.json` version file to next major version? maybe with suffix `-dev`

PR #390, opened 2023-09-01
test: add cargo test to github actions and run only unit tests in ci [TAB-185]

vodkaslime:
# What this PR is about
1. Add `cargo test` to github PR action so that tests would be automatically run when PRs are raised.
2. Remove integ test from `tabby-scheduler`, which would be handled separately from unit tests.


crates/tabby-scheduler/src/lib.rs
```diff
@@ -55,32 +55,3 @@ pub async fn scheduler(now: bool) -> Result<()> {
 
     Ok(())
 }
-
-#[cfg(test)]
```
wsxiaoys, line 59: Could you move this to integration test: https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

and only run unit test in ci?

crates/tabby-scheduler/tests/integration_test.rs
```diff
@@ -0,0 +1,30 @@
...
+        set_tabby_root(TempDir::default().to_path_buf());
+
+        let config = Config {
+            repositories: vec![Repository {
+                git_url: "https://github.com/TabbyML/interview-questions".to_owned(),
+            }],
+            experimental: Experimental::default(),
+        };
+
+        repository::sync_repositories(&config).unwrap();
```
wsxiaoys, line 26: might simply call `tabby_scheduler::scheduler(now = true)`?

by doing so we can keep dataset / index / repository package private

vodkaslime: Sure but how do we pass in the config then? Writing to a local config.toml file?

vodkaslime: Or we could add a `config: Option<Config>` param to `scheduler()` function. Then in `main()` we pass `None` to ask `scheduler()` to load actual config file in real use case.

vodkaslime: Adding param rather than writing local file, in order to avoid actual fs io in integ test.

PR #414, opened 2023-09-08
fix(vscode): do not show inline completion when autocomplete widget is visible.

icycodes:
TAB-175


wsxiaoys:
please ref corresponding TAB issues

Issue #409, opened 2023-09-07
Some other serve backend?


leiwen83:
Hi,

I wonder that whether tabby could add more inference backend like vllm?
Since vllm did a good job in serving batching inference, I think it shall also benefit tabby if deployed at scale.

If this idea is accepted, I'd like to contribute my PR for it.

wsxiaoys:
One idea I have in mind is to create an HTTP API binding for implementing Text Generation. Since vllm provides support for an OpenAPI-like interface, the support should be straightforward.

However, there are certain challenges in the implementation process, such as HTTP cancellation propagation and the need for flexible support for different HTTP backends (e.g., OpenAI, TGI, Palm). Because of these challenges, I haven't started implementing it yet.

leiwen83:
Is there some demo for your prototype http-api-bindings? So that other people could follow your desgin?

For "HTTP cancellation propagation", do you mean like vscode lose server connection, you want to tabby server flush out its previous request?

I think maybe current llama.cpp servering could also be made as http server to provide http interface. So that [TextGeneration](https://github.com/TabbyML/tabby/blob/main/crates/tabby-inference/src/lib.rs) could directly talk to those different cient at will smoothly?

wsxiaoys:
Hi @leiwen83 

#419 integrated an `experimental-http` interface which integrates `Codey` model into tabby as a special type of backend.

> I think maybe current llama.cpp servering could also be made as http server to provide http interface.
While it is technologically feasible, doing so would be incorrect :). Fundamentally, vllm, OpenAI, TGI, and Vertex are individual services or remote APIs that can only be accessed through HTTP API. The inclusion of this support is primarily for quality evaluation purposes at the moment.

leiwen83:
Hi @wsxiaoys ,

I am very excited for this change, and begin to do some experiment base on it.

As the first step, I try to follow your README.md in the new component, to run "cargo run --example simple", but got below error. I am run it within the docker built with latest Dockerfile, so is there anything I missed?

```
error: failed to run custom build command for `openssl v0.10.52`

Caused by:
  process didn't exit successfully: `/data/aigc/workset/code/tabby/target/debug/build/openssl-e51f637ee7c7173e/build-script-build` (exit status: 1)
  --- stderr
[...]

PR #422, opened 2023-09-10
feat: separte Dockefile into two

leiwen83:
So that with tabby_base, we could have setup a easy dev environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,6 @@
-FROM ghcr.io/opennmt/ctranslate2:3.17.1-ubuntu20.04-cuda11.2 as source
-FROM nvidia/cuda:11.2.2-cudnn8-devel-ubuntu20.04 as builder
+# syntax = docker/dockerfile:experimental
 
-ENV CTRANSLATE2_ROOT=/opt/ctranslate2
-COPY --from=source $CTRANSLATE2_ROOT $CTRANSLATE2_ROOT
-
-ENV DEBIAN_FRONTEND=noninteractive
-RUN apt-get update && \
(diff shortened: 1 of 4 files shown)


leiwen83:
@wsxiaoys Could you help review? Now with Dockefile_base, we could have a stable dev docker env.

wsxiaoys:
Hi @leiwen83 

Thank you for the PR. For now, I'd prefer not to recommend a Docker-based development environment for first-time contributors. Setting up a CPU-based environment (without CUDA) is much easier to get started with. I noticed your comments on #409, but it seems you've already resolved the issue (did you need to install `libssl-dev`?).

leiwen83:
> Hi @leiwen83
> 
> Thank you for the PR. For now, I'd prefer not to recommend a Docker-based development environment for first-time contributors. Setting up a CPU-based environment (without CUDA) is much easier to get started with. I noticed your comments on #409, but it seems you've already resolved the issue (after install `libssl-dev`?).

I closed #409 for build from the top is ok and didn't blocking the fastchat integration test, the example still has problem to test with...
I am relative new to rust dev, and notice rust environment is not easy to setup with. So I thought it maybe a good idea to a have dev Dockerfile to put everyone in the same page, even for first-time contributor?

wsxiaoys:
I agree that setting up the Rust environment can be a bit challenging, but I still recommend configuring it in the host environment. This way, the package registry and caching will work seamlessly right from the start.

The [Build](https://github.com/TabbyML/tabby#build) section in the README might be a bit tricky to follow, but instead of adding another layer of complexity, let's work on improving it.

leiwen83:
Got it

PR #447, opened 2023-09-15
fix(intellij): fix conflict with completion widget.

icycodes:
TAB-175


bitnom:
is this why I have to do shift+tab to get the completions?

icycodes:
Hi, @bitnom 

I apologize for the confusion. I'm not clear on the problem you are facing.  
To accept the ghost text shown by the Tabby IntelliJ plugin, you can simply use the Tab key. In case there is a popup widget, such as the default completion widget in the IntelliJ IDEs, you can press Esc to close the popup first. Then, you can proceed to use Tab to accept the completion from the Tabby plugin.  
If you have encountered a problem or if you have any suggestions, please feel free to open a new issue for further discussion.  

This PR is trying to fix conflicts between the Tabby inline completion and other popup widgets in IntelliJ IDEs. However, we have determined that this implementation is not the most suitable approach and will close it soon.

PR #458, opened 2023-09-18
fix(tabby): fix swagger's local server use local port

hufeng:
```
 servers(
        (url = "https://playground.app.tabbyml.com", description = "Playground server"),
        (url = "http://localhost:8080", description = "Local server"),
    ),
```
The default is to use port 8080, but serve can customize the port through --port, resulting in request errors

### DIFF
--- a/crates/tabby/src/serve/mod.rs
+++ b/crates/tabby/src/serve/mod.rs
@@ -161,10 +161,7 @@ pub async fn main(config: &Config, args: &ServeArgs) {
 
     info!("Starting server, this might takes a few minutes...");
 
-    // swagger add local server
-    let mut doc = ApiDoc::openapi();
-    add_localhost_server(&mut doc, args.port);
-
+    let doc = add_localhost_server(ApiDoc::openapi(), args.port);
     let app = Router::new()
         .merge(SwaggerUi::new("/swagger-ui").url("/api-docs/openapi.json", doc))
         .nest("/v1", api_router(args, config))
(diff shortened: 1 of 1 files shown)


crates/tabby/src/serve/mod.rs
```diff
@@ -161,8 +160,20 @@ pub async fn main(config: &Config, args: &ServeArgs) {
     }
 
     info!("Starting server, this might takes a few minutes...");
+
+    // swagger add local server
+    let mut doc = ApiDoc::openapi();
```
wsxiaoys, line 165: extract this into a function like `add_localhost_server(doc: Doc) -> Doc`

wsxiaoys (changes requested):
Thanks for the PR, one nit

crates/tabby/src/serve/mod.rs
hufeng: 👌，Let me refactor the code ～

crates/tabby/src/serve/mod.rs
```diff
@@ -161,8 +160,13 @@ pub async fn main(config: &Config, args: &ServeArgs) {
...
+    add_localhost_server(&mut doc, args.port);
```
wsxiaoys, line 166:
```suggestion
    let doc = add_localhost_server(APIDoc::openapi(), args.port);
```

Will something like this work?

hufeng: yes, it more clear

wsxiaoys (approved):

PR #461, opened 2023-09-19
feat: proxy server address mapping to the model server

hufeng:
image

Before the Tabby service, if there is nginx's reverse proxy service, users can access Tabby's services through domain names. Currently, swagger only has https://playground.app.tabbyml.com and local service options, and cannot perform service completion through domain names.

current server：
image

add proxy server
image

### DIFF
--- a/crates/tabby/src/serve/mod.rs
+++ b/crates/tabby/src/serve/mod.rs
@@ -103,6 +103,10 @@ pub struct ServeArgs {
     #[clap(long)]
     model: String,
 
+    /// proxy server address mapping to the model server.
+    #[clap(long, default_value_t = String::from(""))]
+    proxy_server: String,
+
     #[clap(long, default_value_t = 8080)]
     port: u16,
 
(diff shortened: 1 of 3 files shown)


wsxiaoys (changes requested):
Thanks for the PR, overall looks good. But for a configuration like this I would really prefer it being part of the `config.toml` rather than as a command line argument interface.

Could you add a new sub struct `swagger` in [Config](https://github.com/TabbyML/tabby/blob/main/crates/tabby-common/src/config.rs#L12C6-L12C6) here? the pesudo code would be something like below:

```rust
pub struct Config {
  pub repositories: Vec<Repository>,
  pub swagger: SwaggerConfig
}

pub struct SwaggerConfig {
  pub server_url: Option<String>
}
```

crates/tabby/src/serve/mod.rs
```diff
@@ -251,17 +240,21 @@ fn add_localhost_server(doc: utoipa::openapi::OpenApi, port: u16) -> utoipa::ope
...
-fn add_proxy_server(doc: utoipa::openapi::OpenApi, proxy_server: &str) -> utoipa::openapi::OpenApi {
-    if proxy_server.is_empty() {
+fn add_proxy_server(
+    doc: utoipa::openapi::OpenApi,
+    server_url: Option<String>,
+) -> utoipa::openapi::OpenApi {
+    let server_url: String = server_url.unwrap_or_default();
+    if server_url.is_empty() {
         return doc;
     }
```
wsxiaoys, line 250:
```suggestion
    if server_url.is_none() {
        return doc;
    }
    let server_url = server_url.unwrap();
```

wsxiaoys (commented):
Otherwise LGTM

crates/tabby/src/serve/mod.rs
hufeng: Yes

wsxiaoys (approved):

PR #470, opened 2023-09-23
fix(playground): indent position after completion

wwayne:
Thanks for building the playground page.
Not sure if it is a bug, I was expecting the cursor can keep the same indent with the last line after a multi-lines autocompletion

Group 151

### DIFF
--- a/website/src/components/Monaco/index.js
+++ b/website/src/components/Monaco/index.js
@@ -154,17 +154,34 @@ class CompletionProvider {
   }
 
   toInlineCompletions(value, range) {
-    return (
-      value.choices.map((choice) => ({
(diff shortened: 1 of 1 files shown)


icycodes:
Thank you for your PR.

Indeed, it is unexpected for the cursor to move to the start of the line after accepting a completion. This issue is caused by the completion text ending with a newline character. We have implemented postprocessing logic in the tabby-agent (which is used by IDE plugins) to prevent this behavior. However, this logic has not been applied to the online playground at the moment. We will consider addressing this issue in the future.

wwayne:
Thanks for the detailed explanation

PR #473, opened 2023-09-25
fix(vim): error parsing tabby-agent.js path

ma0dan:
keep getting error below, even after yarn upgrade-agent manully:

> Tabby node script should be download first. Try to run `yarn upgrade-agent`.


adamnovak:
It looks like this change would be needed for compatibility with Vim 8, but the plugin is meant to support Vim 9 or later. On Vim 9, this change is not needed.

PR #475, opened 2023-09-26
Updated completion function name

AkashKarnatak:
`getCompletions` was changed to `provideCompletions` in commit #1c1cf44639d436078db82758baea457326b8a77d

### DIFF
--- a/clients/vim/autoload/tabby.vim
+++ b/clients/vim/autoload/tabby.vim
@@ -282,7 +282,7 @@ function! s:GetCompletion(id)
   endif
 
   let s:pending_request_id = tabby#job#Send(s:tabby, #{
-    \ func: 'getCompletions',
+    \ func: 'provideCompletions',
     \ args: [s:CreateCompletionRequest()],
     \ }, #{
     \ callback: function('s:HandleCompletion', [a:id]),


icycodes:
Thank you for your PR.

In the Vim plugin, an older but compatible version of `tabby-agent` is being retained in `clients/vim/node_scripts/tabby-agent.js`. This version should work with the `getCompletions` function.
The Vim plugin has not received updates for a few weeks, we plan to update it either this week or the next.

If you encounter any issues with the Vim plugin after manually upgrading `tabby-agent`, you can try to revert the changes via git. If you wish to use a newer version of `tabby-agent`, it may require more modifications to ensure compatibility.

We appreciate your understanding and patience as we work towards improving the Vim plugin. Thank you for your contribution.

AkashKarnatak:
I've been encountering a persistent error with the current version of the Vim plugin, consistently receiving the message `Tabby is enabled but not running`. In an attempt to resolve this issue, I decided to update the `clients/vim/node_scripts/tabby-agent.js` file by executing the build commands. However, I noticed that the function `getCompletions` had been renamed to `provideCompletions` in `tabby-agent`. Consequently, I had to manually update the function name in `tabby.vim` as well.

Now, upon reflection, I realize that it might be more practical to revert to an older version of the plugin to avoid these compatibility issues.

icycodes:
The error message `Tabby is enabled but not running` might be caused by an old Node version. Running `tabby-agent.js` on a Node version <16 may result in an unexpected process exit, causing the Vim script to report that Tabby is not running.  
It is required to use Node version >=16 for current vim plugin version. Using Node version >=18 is recommend.

While I cannot make sure that this is the problem you are facing, I hope this information helpful in resolving the issue.

PR #526, opened 2023-10-09
chore: set max timeout for /v1/completions handler

wsxiaoys:


crates/tabby/src/serve/mod.rs
```diff
@@ -201,44 +201,55 @@ fn api_router(args: &ServeArgs) -> Router {
...
-    } else {
-        router
-    };
+    routers.push({
+        Router::new()
+            .route(
+                "/v1/completions",
+                routing::post(completions::completions).with_state(completion_state),
+            )
+            .layer(TimeoutLayer::new(Duration::from_secs(3)))
```
icycodes, line 226: All completion request have a timeout limit of 3s, so this means there is no option to manually trigger and wait for a completion request with a longer response time?

wsxiaoys: In another word, even it's a manual completion, we should still provide SLA <= 3s

icycodes (approved):

PR #544, opened 2023-10-12
feat: add debug request / response to visualize prompting with source code index

wsxiaoys:
image


wsxiaoys:
TBD: add flag for easier SxS between prompt enhanced completion VS vanilla completion

PR #563, opened 2023-10-15
docs: combine docker run command into a single line for ease of running in non-linux shell (e.g windows powershell)

sensinsane:


wsxiaoys:
Hi, thanks for the PR, could you make following adjustment?

1. Adding subtitle for `Linux (Bash)` / `Windows (Powershell)` for each code block.
2. Add command to non-cuda tab
3. Change semantic tag to `docs`

wsxiaoys:
I just realized that the command line is the exact same, except for the line break (`\` in Linux, ` in PowerShell). Is that correct?

wsxiaoys:
In that case, let's simply keep the command in a single line

PR #567, opened 2023-10-16
chore: Added additional Python toplevelkeywords.

Hypnootika:


crates/tabby-common/assets/languages.toml
```diff
@@ -1,7 +1,12 @@
 [[config]]
 languages = ["python"]
 line_comment = "#"
-top_level_keywords = ["def", "from", "class", "import"]
+top_level_keywords = [
+    "def", "class", "return", "raise", "import", "from", "if", "elif", "else",
+    "try", "except", "finally", "for", "while", "break", "continue", "pass",
+    "async", "await", "with", "yield"
+]
+
```
wsxiaoys, line 9: Many keywords mentioned here would not be used at the top level. For example, `yield` won't be used in a global context.

However, some keywords, such as `from` and `async`, do have a chance of being used at the top level.

PR #568, opened 2023-10-16
feat: python.scm

Hypnootika:


wsxiaoys:
Hi, could you share more on how you tweak these treesitter queries?

PR #583, opened 2023-10-18
feat(agent): add count_select in AgentStats event.

icycodes:
Complete TAB-229


wsxiaoys:
if view event is ready, also add count_view?

PR #585, opened 2023-10-18
refactor(agent): remove deprecated agent options.

icycodes:
Complete TAB-226
Related #525

### DIFF
--- a/clients/tabby-agent/src/AgentConfig.ts
+++ b/clients/tabby-agent/src/AgentConfig.ts
@@ -53,9 +53,10 @@ export const defaultAgentConfig: AgentConfig = {
       mode: "adaptive",
       interval: 250, // ms
     },
+    // Deprecated: There is a timeout of 3s on the server side since v0.3.0.
     timeout: {
-      auto: 5000, // 5s
-      manually: 30000, // 30s
+      auto: 4000, // 4s
+      manually: 4000, // 4s
     },
   },
   logs: {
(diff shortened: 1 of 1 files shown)


clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -135,6 +168,12 @@ export const userAgentConfig = isBrowser
         async load() {
           try {
             const fileContent = await fs.readFile(this.filepath, "utf8");
+            // If the config file is the old template, and user has not modified it,
+            // Overwrite it with the new template.
+            if (fileContent.trim() === oldConfigTomlTemplate.trim()) {
+              await this.createTemplate();
+              return await this.load();
+            }
```
icycodes, line 176: Changes in the template does not take effect when user update tabby from old version and the `config.toml` file already exists.  
If user has not modified old template `config.toml`, we can overwrite it. Otherwise, we can not update it directly.

wsxiaoys:
you can Cherry pick 525 into this pr

icycodes:
> you can Cherry pick 525 into this pr

This PR add a new config.toml template that equative to #525 applied.  
The old config.toml template is retained and used to compare if user has modified the config file.  
If user has not modified old template config.toml, we can overwrite it using new template.

clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -109,6 +110,38 @@ const configTomlTemplate = `## Tabby agent configuration file
...
+
+## You can add custom request headers, e.g. for authentication.
+# [server.requestHeaders]
+# Authorization = "Bearer eyJhbGciOiJ..........."
+
+## Completion
+## You can set the debounce mode for auto completion requests when typing.
+# [completion.debounce]
+# mode = "adaptive" # or "fixed"
+# interval = 250 # ms, only used when mode is "fixed"
```
wsxiaoys, line 131:
```suggestion
```

I feel we should also remove the debounce part

PR #588, opened 2023-10-18
fix: update notification for response time issues not to show some su…

icycodes:
…ggestions when using local port.

Fix TAB-220


wsxiaoys (changes requested):
Otherwise LGTM

PR #605, opened 2023-10-21
feat(debug): add debug_options.raw_prompt to tabby api for evaluation purpose

wsxiaoys:


crates/tabby/src/serve/completions.rs
```diff
@@ -42,9 +38,18 @@ pub struct CompletionRequest {
 
 #[derive(Serialize, Deserialize, ToSchema, Clone, Debug)]
 pub struct DebugOptions {
-    /// When true, returns debug_data in completion response.
+    /// When raw_prompt is specified, it will be passed directly to the inference engine for completion. `segments` field in `CompletionRequest` will be ignored.
```
wsxiaoys, line 41:
```suggestion
    /// When `raw_prompt` is specified, it will be passed directly to the inference engine for completion. `segments` field in `CompletionRequest` will be ignored.
```

PR #620, opened 2023-10-23
chore: add support for building on jetson device

yusiwen:
The current building flow will fail on Jetson, because the CTranslate2 binding turns the "sse4.1" flag on while it's not available on Jetson.

I've added a function to check whether it's building on the NVIDIA Jetson device or not. If true, it will turn on CUDA and CUDNN flags of CTranslate2.

I've tested building on Intel CPU, AMD CPU, Jetson Orin Nano, Jetson Orin AGX.

### DIFF
--- a/crates/ctranslate2-bindings/build.rs
+++ b/crates/ctranslate2-bindings/build.rs
@@ -42,15 +42,13 @@ fn link_static() -> PathBuf {
         if cfg!(target_feature = "sse4.1") {
             config.cxxflag("-msse4.1");
         }
-        
+
         if cfg!(feature = "link_static_cuda") {
-            config
-                .define("WITH_CUDA", "ON")
-                .define("WITH_CUDNN", "ON");
-                
+            config.define("WITH_CUDA", "ON").define("WITH_CUDNN", "ON");
+
             if cfg!(target_arch = "aarch64") {
                 config.cxxflag("-mcpu=native");
-            } 
+            }
         } else {
             config.define("WITH_OPENBLAS", "ON");
         }
--- a/crates/ctranslate2-bindings/build.rs
+++ b/crates/ctranslate2-bindings/build.rs
@@ -28,7 +28,7 @@ fn main() {
 }
 
 fn check_jetson() -> bool {
-    if Path::new("/etc/hosts").exists() {
+    if Path::new("/proc/device-tree/model").exists() {
         let contents = fs::read_to_string("/proc/device-tree/model").expect("read /proc/device-tree/model failed");
         let matched = Regex::new(r"^NVIDIA.*Developer Kit").unwrap().is_match(&contents);
         if matched {
--- a/crates/ctranslate2-bindings/build.rs
+++ b/crates/ctranslate2-bindings/build.rs
@@ -54,18 +54,20 @@ fn link_static() -> PathBuf {
             .define("WITH_MKL", "OFF")
             .define("OPENMP_RUNTIME", "NONE");
 
-        if cfg!(target_arch = "aarch64") && check_jetson() {
-            config
-               .define("WITH_CUDA", "ON")
-               .define("WITH_CUDNN", "ON")
-               .cxxflag("-mcpu=native")
-        } else if cfg!(target_feature = "sse4.1") {
-            config
-                .define("WITH_OPENBLAS", "ON")
-                .cxxflag("-msse4.1")
+        if cfg!(target_feature = "sse4.1") {
+            config.cxxflag("-msse4.1");
+        }
+        
+        if cfg!(feature = "link_static_cuda") {
+            config.define("WITH_CUDA", "ON");
+            
+            if cfg!(target_arch = "aarch64") && check_jetson() {
+                config
+                   .define("WITH_CUDNN", "ON")
+                   .cxxflag("-mcpu=native");
+            } 
         } else {
-            config
-                .define("WITH_CUDA", "ON")
+            config.define("WITH_OPENBLAS", "ON");
         }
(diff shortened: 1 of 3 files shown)


crates/ctranslate2-bindings/build.rs
```diff
@@ -37,9 +52,21 @@ fn link_static() -> PathBuf {
...
+               .define("WITH_CUDNN", "ON")
+               .cxxflag("-mcpu=native")
+        } else if cfg!(target_feature = "sse4.1") {
+            config
+                .define("WITH_OPENBLAS", "ON")
+                .cxxflag("-msse4.1")
+        } else {
+            config
+                .define("WITH_CUDA", "ON")
+        }
```
wsxiaoys, line 69: Could you refactor above block with following branching structures?

```rust
if cfg!(target_os = "linux") {
  ...
  if cfg!(target_feature = "sse4.1") {
    ...
  }

  if cfg!(feature = "link_static_cuda") {
    // turn on cuda
    ...
    // if needed, branch here on aarch64 / x86
  } else {
    // Turn on openblas
    ...
  }
} else if cfg!(target_os = "macos") {
  ...
}
```

wsxiaoys (changes requested):

crates/ctranslate2-bindings/build.rs
yusiwen: Copy that! I will try to refactor it ASAP

yusiwen: @wsxiaoys I've refactored as you requested. Please take a look at this revision

crates/ctranslate2-bindings/build.rs
```diff
@@ -54,26 +54,28 @@ fn link_static() -> PathBuf {
...
-                .define("WITH_OPENBLAS", "ON")
-                .cxxflag("-msse4.1")
+        if cfg!(target_feature = "sse4.1") {
+            config.cxxflag("-msse4.1");
+        }
+        
+        if cfg!(feature = "link_static_cuda") {
+            config.define("WITH_CUDA", "ON");
+            
+            if cfg!(target_arch = "aarch64") && check_jetson() {
```
wsxiaoys, line 64:
```suggestion
            if cfg!(target_arch = "aarch64") {
```

check_jetson is no longer needed as you'll need to turn on link_static_cuda manually```

crates/ctranslate2-bindings/build.rs
```diff
@@ -54,26 +54,28 @@ fn link_static() -> PathBuf {
...
+                config
+                   .define("WITH_CUDNN", "ON")
```
wsxiaoys, line 66: this should be enabled no matter if aarch64 / x86?

wsxiaoys (changes requested):

crates/ctranslate2-bindings/build.rs
yusiwen: OK, got it.

crates/ctranslate2-bindings/build.rs
yusiwen: I found on Jetson cuDNN comes with JetPack by default. But on Linux (e.g., Ubuntu), cuDNN needs to be installed separately with CUDA. Does Tabby use cuDNN? If not, I'll remove it

wsxiaoys: Should be good to align with ctranslate2's CUDA compilation configuration

https://github.com/OpenNMT/CTranslate2/blob/master/python/tools/prepare_build_environment_linux.sh#L56C1-L57C1

yusiwen: OK, I got it.

yusiwen: @wsxiaoys `CUDA_DYNAMIC_LOADING=ON` in the script given by you probably conflicts with the literal meaning of `link_static_cuda`

wsxiaoys: I was specifically mentioning the `WITH_CUDNN=ON` in ctranslate2's cuda config

yusiwen: @wsxiaoys How about this version. I've deleted unused code.

wsxiaoys (approved):
LGTM and running CIs

yusiwen:
> LGTM and running CIs

Oops, seems that there is a lint problem. I'm pretty new to Rust, where should I have to look at?

wsxiaoys:
You could run `cargo +nightly fmt` to fix the lint

yusiwen:
Ok, I'm digging into it.

yusiwen:
@wsxiaoys Wierd, My `cargo +nightly fmt --check` shows everything is fine. Do I miss something?

yusiwen:
I've pushed the result of `cargo fmt`. The CI log doesn't show much of information about why it complains my code. How about this version?

wsxiaoys:
> @wsxiaoys I've pushed the result of `cargo fmt`. The CI log doesn't show much of information about why it complains my code. How about this version?

image

This is screenshot from link https://github.com/TabbyML/tabby/actions/runs/6623245595/job/17990064469?pr=620

yusiwen:
@wsxiaoys Yes, I've studied that log, cannot find any information why the code failed to lint. The error just shows "Process completed with exit code 1." at somewhere around line 42.

yusiwen:
@wsxiaoys By the way, while still on the subject of NVIDIA Jetson. Should I add an additional Dockerfile for building images runs on Jeton devices? It will be named as `Dockerfile.jetson`, sounds good?

PR #634, opened 2023-10-25
feat: supports PHP

XpycT:
- [x] Stop words support
- [x] Repository context support
- [x] Update documentation / changelog


crates/tabby-scheduler/Cargo.toml
```diff
@@ -22,6 +22,7 @@ tree-sitter-rust = "0.20.3"
 tree-sitter-typescript = "0.20.3"
 tree-sitter-go = "0.20.0"
 tree-sitter-ruby= "0.20.0"
+tree-sitter-php = { git = "https://github.com/tree-sitter/tree-sitter-php.git" }
```
wsxiaoys, line 25: please add a commit hash to lock the version

crates/tabby-common/assets/languages.toml
```diff
@@ -67,3 +67,31 @@ top_level_keywords = [
...
+    "interface",
+    "namespace",
+    "private",
+    "protected",
+    "public",
+    "require_once",
+    "require",
+    "static",
+    "trait",
+    "use",
```
wsxiaoys, line 96: does it make sense to add `<?php` and `?>` here?

wsxiaoys (changes requested):

crates/tabby-common/assets/languages.toml
wsxiaoys: Hey @XpycT - could you confirm this?

wsxiaoys (approved):

wsxiaoys:
Hi,

I've revert this PR due to following error:

```
SET TABBY ROOT: '/var/folders/gd/3gkhr4g50zbf_zcpf00vmvfr0000gn/T/rstest'
thread 'tests::end_to_end' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidCapture("module")', crates/tabby-scheduler/src/dataset.rs:282:22
```

It seems the query provided in tree-sitter-php is invalid.

PR #635, opened 2023-10-25
Don't get stuck in 'initializing' when script not found

gjedeer:
I've found this when wondering why I'm not getting any completion and status was saying "Tabby is initializing"


icycodes (approved):
@gjedeer 
Thank you for your contribution!

PR #645, opened 2023-10-26
feat: upgrade llama.cpp

wsxiaoys:
this change will merge once we've updated gguf files for all models in https://tabby.tabbyml.com/docs/models/

fix TAB-281

### DIFF
--- a/crates/llama-cpp-bindings/src/engine.cc
+++ b/crates/llama-cpp-bindings/src/engine.cc
@@ -21,7 +21,6 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
   TextInferenceEngineImpl(owned<llama_model> model, owned<llama_context> ctx) :
     model_(std::move(model)),
     ctx_(std::move(ctx)) {
-      batch_ = llama_batch_init(N_BATCH, 0);
   }
 
   void start(rust::Slice<const uint32_t> input_token_ids) override {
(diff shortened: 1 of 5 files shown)


crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -21,7 +21,6 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
   TextInferenceEngineImpl(owned<llama_model> model, owned<llama_context> ctx) :
     model_(std::move(model)),
     ctx_(std::move(ctx)) {
-      batch_ = llama_batch_init(N_BATCH, 0);
```
wsxiaoys, line 24: Previous usage of batch api in tabby generates a segmentation fault with updated llama.cpp version - roll back to llama_batch_get_one as workaround, will revisit this when integrating the continuous batching support.

crates/tabby-common/src/path.rs
```diff
@@ -89,4 +89,8 @@ impl ModelDir {
     pub fn ggml_q8_0_file(&self) -> String {
         self.path_string("ggml/q8_0.gguf")
     }
+
+    pub fn ggml_q8_0_v2_file(&self) -> String {
```
wsxiaoys, line 93: updated llama.cpp requires re-converting all starcoder models, updating filepath to keep forward compatibility.

PR #652, opened 2023-10-27
feat(agent): add experimental option: scope of indentation filter.

icycodes:
Complete TAB-268

Add this config to enable experimental feature:
```
[postprocess.limitScopeByIndentation]
keepBlockScopeWhenCompletingLine = true
```
When completion is continuing the current line, limit the scope to:
- false(default): the line scope, meaning use the next indent level as the limit.
- true: the block scope, meaning use the current indent level as the limit.

Example:
off:
Screenshot from 2023-10-27 19-30-49

on:
Screenshot from 2023-10-27 19-30-17

### DIFF
--- a/clients/tabby-agent/src/AgentConfig.ts
+++ b/clients/tabby-agent/src/AgentConfig.ts
@@ -20,6 +20,14 @@ export type AgentConfig = {
       manually: number;
     };
   };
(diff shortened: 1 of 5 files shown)


wsxiaoys:
Please add `experimental` prefix

PR #679, opened 2023-10-30
fix: minor ts and eslint errors

shixin-guo:
- replace "pino.stdSerializers.err" by import it directly since it is been imported by pino, so sometime Typescript can not resolve the type reference (https://github.com/microsoft/TypeScript/issues/42873)
- fix some obvious TS errors

By the way, should we add ESLint to the package.json script commands? I noticed that the ESLint and Prettier configration files exist, but it seems they've never been run. After triggering the Prettier command, many files changed, so I haven't included these changes in the PR until I receive confirmation from you.

### DIFF
--- a/clients/tabby-agent/package.json
+++ b/clients/tabby-agent/package.json
@@ -43,6 +43,7 @@
     "object-hash": "^3.0.0",
     "openapi-fetch": "^0.7.6",
     "pino": "^8.14.1",
+    "pino-std-serializers": "^6.2.2",
     "rotating-file-stream": "^3.1.0",
     "stats-logscale": "^1.0.7",
     "toml": "^3.0.0",
--- a/clients/tabby-agent/src/logger.ts
+++ b/clients/tabby-agent/src/logger.ts
@@ -1,4 +1,5 @@
 import pino from "pino";
+import { err as pinoStdSerializersError } from "pino-std-serializers";
 import { isBrowser, isTest, testLogDebug } from "./env";
 
 /**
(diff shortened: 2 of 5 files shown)


icycodes:
Hi, @shixin-guo 
Thank you for your PR!
* Could you also add the `yarn.lock` changes for `pino-std-serializers`.
* The fixes for the TypeScript errors look good.
* ESLInt is not currently being used, although the config has not been cleaned up yet. (And I don't have the ESlint Plugin enabled in my VSCode, so I think the errors you fixed were reported by the ESLint Plugin.)
* Prettier is currently in used, `yarn lint` should works properly, and it is also checked in CI. However, I am not sure if you have any additional local config that could cause more changes.

shixin-guo:
> Prettier is currently in used, yarn lint should works properly, and it is also checked in CI. However, I am not sure if you have any additional local config that could cause more changes.
image

You might want to run "**yarn format:check**" in the "/ee/**tabby-ui**" folder. I've come across several warnings locally. We had plans to implement the "importOrder" rule, but we haven't installed [prettier-plugin-sort-imports](https://github.com/IanVS/prettier-plugin-sort-imports) yet. I'm unsure if these rules are necessary for our project because they could result in significant code changes. So, I've decided to hold off on making any changes and am waiting for the team's decision on whether to keep or remove these rules.

icycodes:
Hi, @shixin-guo  
Thank you for bringing up the code style issue.   
After #950, the typescript strict checks and eslint is enabled. It appears that PR is outdated, and I will consider to close it.  
If you have any suggestions on formatting rules or any other improvements, please feel free to open a new PR. We always welcome contributions.

PR #684, opened 2023-11-01
test(agent): add golden test.

icycodes:
Complete TAB-265


clients/tabby-agent/tests/server.docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: "3.5"
```
wsxiaoys, line 1: 1. do not checkin this docker compose.
2. test against StarCoder-1B instead of SantaCoder-1B

wsxiaoys (changes requested):

PR #699, opened 2023-11-05
feat(agent): add experimental limit scope by syntax.

icycodes:
Complete TAB-263

Added experimental feature: use syntax parser to trim completions in post-processing.
Syntax parser:
- tsx (for javascript, typescript, jsx, tsx)
- python
- rust
- go
- ruby

Add this config to enable:
```
[postprocess.limitScope]
experimentalSyntax = false
```


icycodes:
Split to #757 and #758

PR #724, opened 2023-11-08
feat: add ability to only serve chat model

erfanium:
I don't have enough VRAM to deploy both models in the same machine. so i think it's a good idea to be able to deploy normal model and chat model separately.


wsxiaoys:
I'll implement this once https://github.com/TabbyML/tabby/issues/729 is done

Issue #689, opened 2023-11-02
Improve Timeout on Client Agent with CPU configuration


jhwgh1968:
I feel like this is a bug, technically speaking? But as noted in other issues, there is not much support for CPU configuration anymore. This is especially since the removal of the < 1B parameter models. So it is probably an enhancement.

While I have gotten everything working, my issue is that the agent has a hard-coded 4 seconds expectation for the average response time. After 3 requests, considers the server unhealthy and then locks out further completions:

https://github.com/TabbyML/tabby/blob/90e446bfbaef76aedd43d46e74c0ba8a92109acb/clients/tabby-agent/src/CompletionProviderStats.ts#L187

I think this makes sense for most cases, but I personally would not mind having a longer timeout. The code it generates for me is worth it.

Currently I have to hack the source code to achieve this. Would it be possible to add this to the agent configuration? Such as:

```toml
## You can set the timeout for completion requests.
[completion.timeout]
auto = 5000 # ms, for auto completion when typing
manually = 30000 # ms, for manually triggered completion
healthy_limit = 4000 # ms <-- new
```

---
Please reply with a 👍 if you want this feature.

wsxiaoys:
Since we've reintroduced the [server-side timeout option](https://github.com/TabbyML/tabby/commit/21ec60eddfc4d2948a771acf486691e04cafc5c1) (going to be released on v0.5), I agree that it should be fine to add back these options to `agent/config.toml`, especially since it's intended for guru users anyway. 

Given that you've already made it work locally, would you mind submitting a PR?

icycodes:
> After 3 requests, considers the server unhealthy and then locks out further completions:
It only shows a warning for an unhealthy status, you should still be able to receive completions.
However, I agree that the threshold of 4 seconds for the warning should be configurable.

f1zy2:
I was expecting the timeout to be the reason why I couldn't get any completion. But you say that we should still receive them besides that, right?

f1zy2:
> > After 3 requests, considers the server unhealthy and then locks out further completions
> 
> It only shows a warning for an unhealthy status, you should still be able to receive completions. However, I agree that the threshold of 4 seconds for the warning should be configurable.

I was expecting the timeout to be the reason why I couldn't get any completion. But you say that we should still receive them besides that, right?

icycodes:
> > > After 3 requests, considers the server unhealthy and then locks out further completions
> > 
> > 
[...]

PR #739, opened 2023-11-09
feat(agent): add back agent timeout config.

icycodes:
Fix TAB-299 
Related to #689

### DIFF
--- a/clients/tabby-agent/src/AgentConfig.ts
+++ b/clients/tabby-agent/src/AgentConfig.ts
@@ -1,4 +1,5 @@
 import { isBrowser } from "./env";
+import { getProperty, deleteProperty } from "dot-prop";
 
 export type AgentConfig = {
   server: {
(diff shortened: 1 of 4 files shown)


clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -21,6 +21,23 @@ export type AgentConfig = {
       auto: number;
```
wsxiaoys, line 21: use single timeout for both auto and manual in config

clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -21,6 +21,23 @@ export type AgentConfig = {
...
       manually: number;
     };
+    statistics: {
```
wsxiaoys, line 24: remove this from config. All these numbers should be inferred from timeout.

wsxiaoys (changes requested):

PR #742, opened 2023-11-09
feat: add LLAMA_CPP_N_THREADS env

### LINKED ISSUE #738: Add environment variable to control the intra-thread number of thread
Create an environment variable `LLAMA_CPP_N_THREADS` and pass it to llama_set_n_threads.

Code location:  https://github.com/TabbyML/tabby/blob/03ff80efdb792e7e504d28ccedf622d8a851423b/crates/llama-cpp-bindings/src/engine.cc#L291C66-L291C66

Reference: https://github.com/ggerganov/llama.cpp/blob/57ad015dc3011b046ed5a23186c86ea55f987c54/llama.h#L488C20-L488C39

erfanium:
i think we don't need to use `llama_set_n_threads` method. it's likely used for runtime modifications. we only need to set ctx_params

fixes #738

### DIFF
--- a/crates/llama-cpp-bindings/src/engine.cc
+++ b/crates/llama-cpp-bindings/src/engine.cc
@@ -316,11 +316,10 @@ std::unique_ptr<TextInferenceEngine> create_engine(bool use_gpu, rust::Str model
   llama_context_params ctx_params = llama_context_default_params();
   ctx_params.n_ctx = N_CTX * parallelism;
   ctx_params.n_batch = N_BATCH;
-  if (const char* n_threads = std::getenv("LLAMA_CPP_N_THREADS")) {
-    ctx_params.n_threads = std::stoi(n_threads);
-  }
-  if (const char* n_threads_batch = std::getenv("LLAMA_CPP_N_THREADS_BATCH")) {
-    ctx_params.n_threads_batch = std::stoi(n_threads_batch);
+  if (const char* _n_threads = std::getenv("LLAMA_CPP_N_THREADS")) {
+    int n_threads = std::stoi(_n_threads);
+    ctx_params.n_threads = n_threads;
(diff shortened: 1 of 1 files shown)


crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -316,8 +316,13 @@ std::unique_ptr<TextInferenceEngine> create_engine(bool use_gpu, rust::Str model
   llama_context_params ctx_params = llama_context_default_params();
   ctx_params.n_ctx = N_CTX * parallelism;
   ctx_params.n_batch = N_BATCH;
+  if (const char* n_threads = std::getenv("LLAMA_CPP_N_THREADS")) {
+    ctx_params.n_threads = std::stoi(n_threads);
```
wsxiaoys, line 320: let's reuse `LLAMA_CPP_NTHREADS` for both n_threads / n_threas_batch

wsxiaoys (changes requested):

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -316,8 +316,12 @@ std::unique_ptr<TextInferenceEngine> create_engine(bool use_gpu, rust::Str model
   llama_context_params ctx_params = llama_context_default_params();
   ctx_params.n_ctx = N_CTX * parallelism;
   ctx_params.n_batch = N_BATCH;
+  if (const char* _n_threads = std::getenv("LLAMA_CPP_N_THREADS")) {
+    int n_threads = std::stoi(_n_threads);
```
wsxiaoys, line 320:
```suggestion
  if (const char* n_thread_str = std::getenv("LLAMA_CPP_N_THREADS")) {
    int n_threads = std::stoi(n_thread_str);
```

wsxiaoys (approved):

PR #745, opened 2023-11-10
Use upstream TAGGING_QUERY

Squadrick:
The diff between the in-repo query and upstream query:

```
< (type_declaration (type_spec name: (type_identifier) @name)) @definition.type
---
> (type_spec
>   name: (type_identifier) @name) @definition.type
> 
> (type_identifier) @name @reference.type
```


wsxiaoys:
Removing `@reference.type` is by purpose - they're not helping for repository context

PR #750, opened 2023-11-10
refactor: add experimental-http feature

darknight:
This PR is for closing https://github.com/TabbyML/tabby/issues/735

Changes include:
-  Add `experimental-http` feature in `Cargo.toml`
- Make dependency `http-api-bindings` as `optional`
- Update `create_engine` method, only when `experimental-http is enabled` **AND** `tabby serve --device experimental-http`, call `http_api_bindings::create`, other cases go to normal flow
- Update `main` method, only when `experimental-http is enabled` **AND** `tabby serve --device experimental-http`, just log message without loading model, other cases load model normally.

### DIFF
--- a/crates/tabby/Cargo.toml
+++ b/crates/tabby/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2021"
 
 [features]
 cuda = ["llama-cpp-bindings/cuda"]
-experimental-http = []
+experimental-http = ["dep:http-api-bindings"]
 
 [dependencies]
 tabby-common = { path = "../tabby-common" }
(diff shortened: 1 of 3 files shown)


wsxiaoys (approved):
LGTM

Please fix the semantic commit lint

crates/tabby/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 [features]
 cuda = ["llama-cpp-bindings/cuda"]
+experimental-http = []
```
wsxiaoys, line 8: Should making the dependency `http-api-bindings` be optional?

wsxiaoys (changes requested):
adding a comments on dependency of `http-api-bindings`

Issue #752, opened 2023-11-10
[deployment architecture] general questions about how to run the `scheduler`


ghthor:
1. does the scheduler utilize GPU when performing the index?
1. should I be building the dataset & index out of band from `tabby serve` and only loading them at `tabby serve` startup
1. My companies monorepo produced a dataset size of 122M. Scheduler is consuming 9G of ram and is still indexing after 3 hours of runtime, does this seem correct?

wsxiaoys:
Replied inline

> 1. does the scheduler utilize GPU when performing the index?

No. Only CPU is utilized.

> 2. should I be building the dataset & index out of band from `tabby serve` and only loading them at `tabby serve` startup

`tabby serve` support automatically loading of indexing snapshots, so `tabby scheduler` can be run as an individual process to automatically build the index at a interval.


> 3. My companies monorepo produced a dataset size of 122M. Scheduler is consuming 9G of ram and is still indexing after 3 hours of runtime, does this seem correct?

The RAM usage appears to be correct. However, the indexing strategy is not currently heavily optimized. Nonetheless, achieving the task in 3 hours seems like a reasonable outcome given the current implementation.

ghthor:
@wsxiaoys Thanks for the quick response

> 3\. My companies monorepo produced a dataset size of 122M. Scheduler is consuming 9G of ram and is still indexing after 3 hours of runtime, does this seem correct?

It is actually still creating the index and it's been running for 5hrs now and the memory usage is up to 12G[1] according to my orchestration system(nomad); BUT I'm not sure that is actually being consumed by the rust binary. Maybe that memory usage is coming from mem mapped files? I'm not exactly sure how to check that.

```shell
root@2485851e65d4:/# ps auxf
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root      184844  0.1  0.0   4248  2688 pts/0    Ss   18:33   0:00 bash
root      184933  0.0  0.0   5900  2304 pts/0    R+   18:33   0:00  \_ ps auxf
root           1 17.9  0.4 4668200 315632 ?      Ssl  13:45  51:44 /opt/tabby/bin/tabby scheduler --now
root@2485851e65d4:/# echo $((315632 / 1024))
308
root@2485851e65d4:/# du -hs /tmp/
0	/tmp/
root@2485851e65d4:/# du -hs /data/
43G	/data/
root@2485851e65d4:/# du -hs /data/
[...]

PR #753, opened 2023-11-10
fix: tantivy requires more memory for indexing in new version.

wsxiaoys:
https://github.com/quickwit-oss/tantivy/issues/2156

related #752


ghthor (commented):
:rocket:
:fire: 
:fire:
:earth_americas:

PR #766, opened 2023-11-11
feat(intellij): add loading status in intellij plugin when fetching completions for automatic trigger mode.

costanzo:
When users are using IntelliJ plugin, users have no idea whether the plugin is sending completion request to the server. 

This new feature is like the github copilot. When the plugin is waiting for the response from server, the icon in the right bottom will be like a loading icon.


icycodes:
Hi @costanzo,

Thank you for your pull request!

Currently, there is a loading status implemented for the manual trigger mode in this section of the code:
https://github.com/TabbyML/tabby/blob/e3518b46a55f9d61d092543e51e297a1a27dab47/clients/intellij/src/main/kotlin/com/tabbyml/intellijtabby/status/StatusBarWidgetFactory.kt#L126-L134

For the automatic mode, you can try adding similar code. Please avoid modifying the Agent status code section.

By the way, initially I did not consider showing the loading status for the automatic mode because the status changes quite frequently and I felt it may not be necessary. However, I now agree that adding the loading status for the automatic mode would be beneficial.

costanzo:
@icycodes I have updated the logic and it looks perfect now. 
Do you have any plan to use any other completion strategies? like caching? 
The requests to server are pretty frequent. I have read some articles about github copilot plugin like [this](https://zhuanlan.zhihu.com/p/639993637). 
There are quite a lot strategies there deciding when to send request to server.

icycodes:
> @icycodes I have updated the logic and it looks perfect now. Do you have any plan to use any other completion strategies? like caching? The requests to server are pretty frequent. I have read some articles about github copilot plugin like [this](https://zhuanlan.zhihu.com/p/639993637). There are quite a lot strategies there deciding when to send request to server.

Currently, we have implemented several client-side strategies, including caching, debouncing, and post-processing in the Tabby Agent, which is shared among IDE plugins. For the IntelliJ plugin, the Tabby Agent runs on a separate Node.js process and acts as the intermediary between the IDE and Tabby server.
However, we understand that we still have room for improvement. We will continually optimizing Tabby moving forward.

PR #776, opened 2023-11-13
fix: Source code index is not ready

darknight:
When start `tabby serve` with debug log level, I saw debug log like below:

```
2023-11-13T12:29:04.341206Z DEBUG tabby::services::code: crates/tabby/src/services/code.rs:61: Source code index is not ready `Directory does not exist: '/home/<user>/.tabby/index'.`
```

After creating `index` dir manually, still got unexpected log like:

```
2023-11-13T08:06:39.550986Z DEBUG tabby::services::code: crates/tabby/src/services/code.rs:58: Source code index is not ready `Failed to open file for read: 'FileDoesNotExist("meta.json")'`
```

Turns out, `tabby serve` starts with the call of `CodeSearchService::new()`, inside which it spawns a task to call `CodeSearchImpl::load()` method. This `load` method tries to open directory `~/.tabby/index/` and read `meta.json` file.

Both file and directory do not exist at all **UNLESS** we run `tabby scheduler` first, which will create them for writing.

`tabby serve` is independent of `tabby scheduler`, so running `tabby scheduler` before `tabby serve` makes no sense.

This PR is to fix this issue by reusing `create or open` logic for `tabby serve`

### DIFF
--- a/crates/tabby-scheduler/src/index.rs
+++ b/crates/tabby-scheduler/src/index.rs
@@ -8,7 +8,7 @@ use tabby_common::{
     path::index_dir,
     SourceFile,
 };
-use tantivy::{directory::MmapDirectory, doc, Index};
+use tantivy::{directory::MmapDirectory, doc, schema::Schema, Index};
 
 use crate::utils::tqdm;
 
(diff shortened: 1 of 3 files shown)


wsxiaoys:
Hey, this behavior is functioning as intended, but I'm open to discussing whether there's a better approach.

The requirements here are as follows:

1. We'd like to allow the user to start `tabby serve` in a blank setup.
2. When `tabby serve` is running, it should be able to pick up a snapshot created in another process.
3. (current behavior) When the index is not ready (or broken), leave it as is.

This PR essentially changes 3 to creating the index. While this is acceptable, I feel the current behavior is more favorable as it isolates the functionality between commands (`tabby serve` will never write the index). This makes it easier for ops in certain scenarios.

Let me know your thoughts on this.

ghthor:
> 3\. (current behavior) When the index is not ready (or broken), leave it as is.

If this happens and the error message is printed about there being no index is it required to restart `tabby serve` process after `tabby scheduler` has been run and created an index?

If that's the case, I'd advocate for no change in behavior and instead just enhancing the log message with how to resolve having no index.

```
To create an index you must configure repositories and run `tabby scheduler` and restart `tabby serve`
```

If restarted `tabby serve` is not required, then that of course would be left out.

darknight:
Let's leave it as it is for now, close PR.

PR #812, opened 2023-11-17
feat: enable tabby to print debug logs

costanzo:
Add --enable-debug flag so tabby will print debug logs.


wsxiaoys:
You could enable debug log by setting
```
RUST_LOG=debug tabby ...
```

costanzo:
I have tried the `RUST_LOG=debug` eviroment flag. Indeed some modules' debug logs are enabled but the tabby debug log itself is not enabled. Because it forced to INFO here `crates/tabby/src/main.rs`:

```
...
.add_directive("tabby=info".parse().unwrap())
...
```

I am not sure why it is set INFO here but maybe we can remove that and accept the `RUST_LOG` environment flag to control?

wsxiaoys:
> I am not sure why it is set INFO here but maybe we can remove that and accept the `RUST_LOG` environment flag to control?

The goal is to enable default info-level logging for the Tabby package. Perhaps we can implement a check to prevent the acceptance of a more verbose environment variable override, such as RUST_LOG=tabby=debug.

costanzo:
Oh, I just realized the default tracing log level is ERROR not INFO if `RUST_LOG` is not set.

> Perhaps we can implement a check to prevent the acceptance of a more verbose environment variable override, such as `RUST_LOG=tabby=debug`

Why do we need to avoid a verbose environment variable override. I mean we definitely need to print debug logs in Tabby package, right?

wsxiaoys:
> > Perhaps we can implement a check to prevent the acceptance of a more verbose environment variable override, such as `RUST_LOG=tabby=debug`
> 
> Why do we need to avoid a verbose environment variable override. I mean we definitely need to print debug logs in Tabby package, right?

To *avoid* pretenting the acceptance :)

Based on https://docs.rs/tracing-subscriber/latest/src/tracing_subscriber/filter/env/builder.rs.html#69

A proper implementation would be
```rust
let filter = EnvFilter::builder()
     .with_default_directive(LevelFilter::ERROR.into())
     .parse_lossy("tabby=info")
     ...
     .from_env_lossy()
```

costanzo:
Hi @wsxiaoys if I am understanding correctly, we want to set the default log level to be **error**, and set default tabby log level to be **info**. At the same time we want the envrionment log settings to be able to override the default tabby log level?

I check the tracing-subscriber doc and find there is no way of doing that.

`parse_lossy` method will return an EnvFilter, which will no longer parse env. 

So I just set default directive to be `LevelFilter::INFO` which can enable Tabby pacakge logging. And info log in other packages is helpful for various purposes.

wsxiaoys:
I think it's still doable - though you might need dig into the implementation of `EnvFilter` a bit.

wsxiaoys:
This is done recently: https://github.com/TabbyML/tabby/blob/24c544cfa144c1ed7298f89f1ccccc27cb697f7d/crates/tabby/src/main.rs#L223

PR #813, opened 2023-11-17
feat: add support for Kotlin language

sonique6784:
Added support for Kotlin language


website/docs/programming-languages.md
```diff
@@ -42,3 +42,4 @@ For an actual example of an issue or pull request adding the above support, plea
 |   PHP    |                    🚫                    |                       🚫                        |
 |   Perl   |                    🚫                    |                       🚫                        |
 |  Scala   |                    🚫                    |                       🚫                        |
+|  Kotlin  |                    🚫                    |                       🚫                        |
```
wsxiaoys, line 45: Could you add a row in line 31 to indicate version supporting kotlin?

crates/tabby-common/assets/languages.toml
```diff
@@ -41,6 +41,31 @@ top_level_keywords = [
...
+    "sealed",
+    "suspend",
+    "fun",
+    "import",
+    "interface",
+    "package",
+    "var",
+    "val",
+    "enum",
+    "return",
```
wsxiaoys, line 66: is return a top level keywords in kotlin?

wsxiaoys (changes requested):
Please also fix the ci

crates/tabby-common/assets/languages.toml
sonique6784: @wsxiaoys , I removed "return" , thanks for catching that.
Would it make sense to add:
"data class" instead of just "data"
"sealed class" instead of just "sealed"

website/docs/programming-languages.md
sonique6784: done

crates/tabby-common/assets/languages.toml
wsxiaoys: data / sealed should be good in this case

sonique6784:
I rebased and move the kotlin lang to tags.rs,
I ran `make fix` and `cargo test --bin tabby --lib` both are successful

ready for review

crates/tabby-scheduler/src/dataset/tags.rs
```diff
@@ -73,6 +73,12 @@ lazy_static! {
                     .unwrap(),
                 ),
             ),
+            (
+                "kotlin",
+                TagsConfigurationSync(
+                    TagsConfiguration::new(tree_sitter_kotlin::language(), "", "").unwrap(),
```
wsxiaoys, line 79: Without the tags query, indexing will not function properly (as no tags will be extracted), resulting in the `Repository Context` functionality not working.

For guidance, you can refer to examples such as https://github.com/TabbyML/tabby/blob/main/crates/tabby-scheduler/queries/go.scm and implement it using https://tree-sitter.github.io/tree-sitter/playground.

Alternatively, you can choose to remove this marker and retain only the `Stop Words` section checked in the markdown document.

sonique6784: that make sense. I have added kotlin.scm to add the tags query.

crates/tabby-scheduler/queries/kotlin.scm
```diff
@@ -0,0 +1,37 @@
+(
+  (comment)* @doc
+  .
+  (function_declaration
+    name: (identifier) @name) @definition.function
+  (#strip! @doc "^//\\s*")
+  (#set-adjacent! @doc @definition.function)
+)
```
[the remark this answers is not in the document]
sonique6784: thanks, I played around with that tool and fixed the kotlin.scm file

crates/tabby-scheduler/queries/kotlin.scm
```diff
@@ -0,0 +1,19 @@
...
+  (class_declaration (type_identifier) @name) @definition.class
+)
+
+(
+  (object_literal  (delegation_specifier) @name) @definition.object
+)
+
+(
+  (type_alias (type_identifier) @name) @definition.type
+)
```
wsxiaoys, line 15:
```suggestion
```

Type alias doesn’t really introduce meaningful info for code completion?

crates/tabby-scheduler/queries/kotlin.scm
```diff
@@ -0,0 +1,19 @@
...
+
+(
+  (import_header (identifier) @name) @definition.import
+)
```
wsxiaoys, line 19:
```suggestion
```

ditto

crates/tabby-scheduler/queries/kotlin.scm
```diff
@@ -0,0 +1,19 @@
...
+  (function_declaration (simple_identifier) @name) @definition.function
+)
+
+(
+  (class_declaration (type_identifier) @name) @definition.class
+)
+
+(
+  (object_literal  (delegation_specifier) @name) @definition.object
+)
```
wsxiaoys, line 11:
```suggestion
```

is object literal something actually referencable, from a cross file context perspective?

wsxiaoys (commented):
Otherwise LGTM

crates/tabby-scheduler/queries/kotlin.scm
sonique6784: yes it is, I woud keep it

PR #824, opened 2023-11-18
feat(vim): update tabby-agent.

icycodes:
Update tabby-agent for vim plugin.


paralin:
@icycodes Hi - is there any info on where these files come from? Thanks!

PR #829, opened 2023-11-18
feat: update graphql query to read token

darknight:
- Update GraphQL schema
- Add impl to read token


ee/tabby-webserver/Cargo.toml
```diff
@@ -25,7 +25,7 @@ serde.workspace = true
 tabby-common = { path = "../../crates/tabby-common" }
 tarpc = { version = "0.33.0", features = ["serde-transport"] }
 thiserror.workspace = true
-tokio.workspace = true
+tokio = { workspace = true, features = ["fs"] }
```
darknight, line 28: `cargo run --package tabby-webserver --example update-schema` will fail if `fs` feature is not specified.

PR #838, opened 2023-11-19
feat: add Prometheus support to Tabby.

jpoisso:
1) Added `axum-prometheus` to Cargo.toml

2) Added `metrics.rs` as a simple route to produce a metrics endpoint with prometheus-formatted events

3) Added `/v1/metrics` endpoint with API doc entry

4) Added the `PrometheusLayer` to the root layers.

This change effectively allows for external Prometheus to scrape metrics from Tabby during execution in order to monitor operations (resources, timings and overall usage).

Metrics can be viewed from the API docs and by calling `/v1/metrics` on a running instance.

### DIFF
--- a/crates/tabby/src/serve.rs
+++ b/crates/tabby/src/serve.rs
@@ -5,6 +5,7 @@ use std::{
 };
 
 use axum::{routing, Router, Server};
+use axum_prometheus::PrometheusMetricLayer;
 use axum_tracing_opentelemetry::opentelemetry_tracing_layer;
 use clap::Args;
 use tabby_common::{
@@ -49,7 +50,7 @@ Install following IDE / Editor extensions to get started with [Tabby](https://gi
     servers(
         (url = "/", description = "Server"),
     ),
-    paths(routes::log_event, routes::completions, routes::completions, routes::health, routes::search),
+    paths(routes::log_event, routes::completions, routes::completions, routes::health, routes::search, routes::metrics),
     components(schemas(
         api::event::LogEventRequest,
         completion::CompletionRequest,
@@ -175,6 +176,10 @@ async fn api_router(
         args.chat_model.as_deref(),
         &args.device,
     ));
+
+    let (prometheus_layer, metric_handle) = PrometheusMetricLayer::pair();
+    let metrics_state = Arc::new(metric_handle);
+
     routers.push({
         Router::new()
             .route(
@@ -189,6 +194,10 @@ async fn api_router(
                 "/v1/health",
                 routing::get(routes::health).with_state(health_state),
             )
+            .route(
+            "/v1/metrics",
+            routing::get(routes::metrics).with_state(metrics_state),
+        )
     });
 
     if let Some(completion_state) = completion_state {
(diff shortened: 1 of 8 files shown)


crates/tabby/src/serve.rs
```diff
@@ -189,6 +194,10 @@ async fn api_router(
                 "/v1/health",
                 routing::get(routes::health).with_state(health_state),
             )
+            .route(
```
wsxiaoys, line 197: move this to `/metrics`

wsxiaoys (changes requested):
Thanks for the PR, just one nit

crates/tabby/src/serve.rs
```diff
@@ -189,6 +194,10 @@ async fn api_router(
...
+            "/v1/metrics",
```
wsxiaoys, line 198: 1. Make the route listen to `/metrics`
2. Add the route to `worker.rs` as well

crates/tabby/src/worker.rs
```diff
@@ -63,19 +64,25 @@ async fn make_chat_route(context: WorkerContext, args: &WorkerArgs) -> Router {
...
-        routing::post(routes::completions).with_state(completion_state),
-    )
+    let metrics_handle = Arc::new(metrics_handle);
+
+    Router::new()
+        .route(
+            "/v1/completions",
+            routing::post(routes::completions).with_state(completion_state),
+        )
+        .route(
```
wsxiaoys, line 82: this can be put in `main` line 102? (since both completion worker / chat worker should expose metrics)

crates/tabby/src/serve.rs
jpoisso: Thanks for the review!

There were some changes merged in the meantime which introduced some additional work, let me know if that's what you had in mind

crates/tabby/src/worker.rs
jpoisso: I'll look into it shortly, I'm getting up to speed on the codebase as a whole

jpoisso: It seems much better now.  

I'm having some issues running Tabby from a fresh built binary if I specify a model, any advice?

For testing metrics on `serve`, I can test it by not specifying a model (endpoints fail but metrics works), but for `worker` commands, it expects a model and it fails to run.

For example:

`RUST_BACKTRACE=1 target/debug/tabby worker::completion --url test.com --model 'TabbyML/StarCoder-1B' --token test`


`thread 'main' panicked at crates/tabby-common/src/registry.rs:72:51:
called `Result::unwrap()` on an `Err` value: Error(TrackableError { kind: Other, cause: Some(Cause(Os { code: 2, kind: NotFound, message: "No such file or directory" })), history: History([Location { module_path: "serdeconv::convert_json", file: "/home/redacted/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serdeconv-0.4.1/src/convert_json.rs", line: 53, message: "" }]) })`

wsxiaoys: the PR looks good - will test the worker for you.

PR #842, opened 2023-11-20
fix: make chat sidebar color consistent with main page sidebar

wsxiaoys:
cc @liangfung


wsxiaoys:
cc @liangfung 

I moved the chat sidebar related to components - since they won't be reused in other page anyway (and if they do - we can refactor at the time)

PR #844, opened 2023-11-20
fix(agent): Hotfix 1.1.1. Fix agent completion stats initialization error.

icycodes:


clients/intellij/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.1.1
```
wsxiaoys, line 1: should be 1.1.2？

icycodes: I changed only vscode extension to 1.1.2.   
IntelliJ and vim plugin does not have a 1.1.1 version yet.

PR #850, opened 2023-11-20
refactor: move magic numbers into tabby_common::constants

### LINKED ISSUE #816: Refactoring the magic numbers related to LLM context length
So far we have following places defines magic number related to context length

https://github.com/TabbyML/tabby/blob/main/crates/tabby/src/services/completion.rs#L200
https://github.com/TabbyML/tabby/blob/main/crates/tabby/src/services/chat.rs#L55
https://github.com/TabbyML/tabby/blob/main/crates/llama-cpp-bindings/src/engine.cc#L18
https://github.com/TabbyML/tabby/blob/main/crates/tabby/src/services/completion/completion_prompt.rs#L14

We shall refactor these to a centralized place (e.g a rust enum).

erfanium:
an attempt to partially fix #816

I'll make another PR for magic numbers inside engine.cc

### DIFF
--- a/crates/tabby-common/src/constants.rs
+++ b/crates/tabby-common/src/constants.rs
@@ -1,0 +1,46 @@
+use crate::languages::get_language;
+use crate::languages::Language;
+use crate::languages::EMPTY_LANGUAGE;
+
+pub struct TextGenerationOptions {
+    pub max_input_length: usize,
+    pub max_decoding_length: usize,
+    pub sampling_temperature: f32,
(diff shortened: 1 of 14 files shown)


crates/tabby-common/src/constants.rs
wsxiaoys, line 1: The content of this file should goes to `TextGenerationOptions` inside `tabby-inference.

wsxiaoys (changes requested):

Issue #815, opened 2023-11-17
API endpoint to serve static files under repositories


wsxiaoys:
* GET `/repositories/{name}/resolve/{...path}` (returns `~/.tabby/repositories/{name}/{..path}` content, if it's dir, returns json file with file / dir names under the directory, with a special mime type)
  - [Example](https://github.com/tokio-rs/axum/blob/main/examples/static-file-server/src/main.rs)

* GET `/repositories/{name}/meta/{...path}` (returns everything except the content field)

To efficiently serve json files, I recommend simplify read every rows into a memory (except the content field). In future this can be easily replaced  (e.g use sqlite)

Example record
```
{
  "git_url": "https://github.com/TabbyML/tabby",
  "filepath": "tests/default.loadtest.js",
  "content": "import http from \"k6/http\";\nimport { check, group, sleep, abortTest } from \"k6\";\n\nexport const options = {\n  stages: [\n    { duration: \"5s\", target: 8 },\n    { duration: \"20s\", target: 8 },\n    { duration: \"5s\", target: 0 },\n  ],\n  hosts: {\n    \"api.tabbyml.com\": __ENV.TABBY_API_HOST,\n  },\n  // Below thresholds are tested against TabbyML/StarCoder-1B served by NVIDIA T4 GPU.\n  thresholds: {\n    http_req_failed: ['rate<0.01'], // http errors should be less than 1%\n    http_req_duration: [\"med<2500\", \"avg<2200\", \"p(95)<2500\"],\n  },\n};\n\nexport default () => {\n  const payload = JSON.stringify({\n    language: \"python\",\n    segments: {\n      prefix: \"def binarySearch(arr, left, right, x):\\n    mid = (left +\"\n    },\n  });\n  const headers = { \"Content-Type\": \"application/json\" };\n  const res = http.post(\"http://api.tabbyml.com/v1/completions\", payload, {\n    headers,\n  });\n  check(res, { success: (r) => r.status === 200 });\n  sleep(0.5);\n};\n",
  "language": "javascript-typescript",
  "max_line_length": 86,
  "avg_line_length": 27.69697,
  "alphanum_fraction": 0.5902851,
  "tags": [
    {
      "range": {
        "start": 581,
        "end": 723
      },
      "name_range": {
        "start": 586,
        "end": 595
      },
      "line_range": {
        "start": 565,
        "end": 597
      },
      "is_definition": false,
      "syntax_type_name": "call"
    },
    {
      "range": {
[...]

PR #851, opened 2023-11-21
feat: add api to serve files under repositories

darknight:
For closing https://github.com/TabbyML/tabby/issues/815

Changes include:

- Add 2 public APIs
  - `resolve` for fetching file or dir info from local repository located at `~/.tabby/repositories/`
  - `meta` for fetching dataset info from file `~/.tabby/dataset/files.jsonl`
- Add api spec doc

Please check `ee/tabby-webserver/docs/api_spec.md` for response examples.

### DIFF
--- a/crates/tabby-common/src/config.rs
+++ b/crates/tabby-common/src/config.rs
@@ -50,12 +50,16 @@ impl RepositoryConfig {
             let path = self.git_url.strip_prefix("file://").unwrap();
             path.into()
         } else {
-            repositories_dir().join(to_filename(&self.git_url))
+            repositories_dir().join(self.name())
         }
     }
 
     pub fn is_local_dir(&self) -> bool {
         self.git_url.starts_with("file://")
+    }
+
+    pub fn name(&self) -> String {
+        filenamify(&self.git_url)
     }
 }
 
@@ -73,13 +77,9 @@ impl Default for ServerConfig {
     }
 }
 
-pub fn to_filename<S: AsRef<str>>(s: S) -> String {
-    filenamify(s)
-}
-
 #[cfg(test)]
 mod tests {
-    use super::{to_filename, Config, RepositoryConfig};
+    use super::{Config, RepositoryConfig};
 
     #[test]
     fn it_parses_empty_config() {
@@ -102,8 +102,10 @@ mod tests {
     }
 
     #[test]
-    fn test_to_filename() {
-        let url = "https://github.com/TabbyML/tabby.git".to_string();
-        assert_eq!(to_filename(url), "https_github.com_TabbyML_tabby.git");
+    fn test_repository_config_name() {
+        let repo = RepositoryConfig {
+            git_url: "https://github.com/TabbyML/tabby.git".to_owned(),
+        };
+        assert_eq!(repo.name(), "https_github.com_TabbyML_tabby.git");
     }
 }
--- a/crates/tabby-common/src/config.rs
+++ b/crates/tabby-common/src/config.rs
@@ -50,7 +50,7 @@ impl RepositoryConfig {
             let path = self.git_url.strip_prefix("file://").unwrap();
             path.into()
         } else {
-            repositories_dir().join(filenamify(&self.git_url))
+            repositories_dir().join(to_filename(&self.git_url))
         }
     }
 
@@ -73,9 +73,13 @@ impl Default for ServerConfig {
     }
 }
 
+pub fn to_filename<S: AsRef<str>>(s: S) -> String {
+    filenamify(s)
+}
+
 #[cfg(test)]
 mod tests {
-    use super::{Config, RepositoryConfig};
+    use super::{to_filename, Config, RepositoryConfig};
 
     #[test]
     fn it_parses_empty_config() {
@@ -95,5 +99,11 @@ mod tests {
             git_url: "https://github.com/TabbyML/tabby".to_owned(),
         };
         assert!(!repo.is_local_dir());
+    }
+
+    #[test]
+    fn test_to_filename() {
+        let url = "https://github.com/TabbyML/tabby.git".to_string();
+        assert_eq!(to_filename(url), "https_github.com_TabbyML_tabby.git");
     }
 }
--- a/ee/tabby-webserver/Cargo.toml
+++ b/ee/tabby-webserver/Cargo.toml
@@ -29,6 +29,8 @@ thiserror.workspace = true
 tokio = { workspace = true, features = ["fs"] }
 tokio-rusqlite = "0.4.0"
 tokio-tungstenite = "0.20.1"
+tower = { version = "0.4", features = ["util"] }
+tower-http = { version = "0.4.0", features = ["fs", "trace"] }
 tracing.workspace = true
 unicase = "2.7.0"
 
--- a/ee/tabby-webserver/docs/api_spec.md
+++ b/ee/tabby-webserver/docs/api_spec.md
@@ -1,0 +1,143 @@
+# API Specs
+
+## Repository api: `/repositories`
+
+### Resolve
+
+Get file or directory content from local repositories
+
+**URL:** `/repositories/{name}/resolve/{path}`
+
+**Method:** `GET`
+
+**Request examples:**
+
+- Get directory content
+
+```shell
+curl --request GET \
+  --url http://localhost:8080/repositories/https_github.com_TabbyML_tabby.git/resolve/
+
+curl --request GET \
+  --url http://localhost:9090/repositories/https_github.com_TabbyML_tabby.git/resolve/ee/tabby-webserver/
+```
+
+- Get file content
+
+```shell
+curl --request GET \
+  --url http://localhost:8080/repositories/https_github.com_TabbyML_tabby.git/resolve/package.json
+
+curl --request GET \
+  --url http://localhost:9090/repositories/https_github.com_TabbyML_tabby.git/resolve/ee/tabby-webserver/src/api.rs
+```
+
+**Response examples:**
+
+- All directory query will return a list of string, with each string represents an entry under that directory. The `Content-Type` for directory query is `application/vnd.directory+json`.
+
+For `/repositories/https_github.com_TabbyML_tabby.git/resolve/ee/tabby-webserver/`, the response is:
+
+```json
+{
+  "entries": [
+    "ee/tabby-webserver/src",
+    "ee/tabby-webserver/ui",
+    "ee/tabby-webserver/examples",
+    "ee/tabby-webserver/Cargo.toml",
+    "ee/tabby-webserver/graphql"
+  ]
+}
+```
+
+- The file query will return file content, the `Content-Type` will be guessed from the file extension.
+
+For request `/repositories/https_github.com_TabbyML_tabby.git/resolve/package.json`, the content type is `application/json`, and the response is:
+
+```json
+{
+  "private": true,
+  "workspaces": [
+    "clients/tabby-agent",
+    "clients/vscode",
+    "clients/vim",
+    "clients/intellij"
+  ],
+  "engines": {
+    "node": ">=18"
+  }
+}
+```
+
+For request `/repositories/https_github.com_TabbyML_tabby.git/resolve/ee/tabby-webserver/src/api.rs`, the content type is `text/x-rust`, and the response is:
+
+```text
+use async_trait::async_trait;
+use juniper::{GraphQLEnum, GraphQLObject};
+use serde::{Deserialize, Serialize};
+use tabby_common::api::{
+    code::{CodeSearch, CodeSearchError, SearchResponse},
+    event::RawEventLogger,
+};
+use thiserror::Error;
+use tokio_tungstenite::connect_async;
+
+use crate::websocket::WebSocketTransport;
+
+#[derive(GraphQLEnum, Serialize, Deserialize, Clone, Debug)]
+pub enum WorkerKind {
+    Completion,
+    Chat,
+}
+
+......omit......
+```
+
+### Meta
+
+Get dataset entry for each indexed file in the repository
+
+**URL:** `/repositories/{name}/meta/{path}`
+
+**Method:** `GET`
+
+**Request example:**
+
+```shell
+curl --request GET \
+  --url http://localhost:9090/repositories/https_github.com_TabbyML_tabby.git/meta/ee/tabby-webserver/src/lib.rs
+```
+
+**Response example:**
+
+The `Content-Type` for successful response is always `application/json`.
+
+```json
+{
+  "git_url": "https://github.com/TabbyML/tabby.git",
+  "filepath": "ee/tabby-webserver/src/lib.rs",
+  "language": "rust",
+  "max_line_length": 88,
+  "avg_line_length": 26.340782,
+  "alphanum_fraction": 0.56416017,
+  "tags": [
+    {
+      "range": {
+        "start": 0,
+        "end": 12
+      },
+      "name_range": {
+        "start": 8,
+        "end": 11
+      },
+      "line_range": {
+        "start": 0,
+        "end": 12
+      },
+      "is_definition": true,
+      "syntax_type_name": "module"
+    },
+    ......omit......
+  ]
+}
+```
--- a/ee/tabby-webserver/src/lib.rs
+++ b/ee/tabby-webserver/src/lib.rs
@@ -51,7 +51,7 @@ pub async fn attach_webserver(
         .route("/graphql", routing::get(playground("/graphql", None)))
         .layer(Extension(schema))
         .route("/hub", routing::get(ws_handler).with_state(ctx))
-        .nest("/repositories", repositories::routers());
+        .nest("/repositories", repositories::routes());
 
     let ui = ui
         .route("/graphiql", routing::get(graphiql("/graphql", None)))
--- a/ee/tabby-webserver/src/lib.rs
+++ b/ee/tabby-webserver/src/lib.rs
@@ -32,8 +32,6 @@ use schema::Schema;
 use server::ServerContext;
 use tarpc::server::{BaseChannel, Channel};
 
-use crate::repositories::repo::load_dataset;
-
 pub async fn attach_webserver(
     api: Router,
     ui: Router,
@@ -53,23 +51,13 @@ pub async fn attach_webserver(
         .route("/graphql", routing::get(playground("/graphql", None)))
         .layer(Extension(schema))
         .route("/hub", routing::get(ws_handler).with_state(ctx))
-        .nest("/repositories", repositories_routers().await);
+        .nest("/repositories", repositories::routers());
 
     let ui = ui
         .route("/graphiql", routing::get(graphiql("/graphql", None)))
         .fallback(ui::handler);
 
     (api, ui)
-}
-
-async fn repositories_routers() -> Router {
-    load_dataset().await.unwrap();
-
-    Router::new()
-        .route("/:name/resolve/", routing::get(repositories::resolve))
-        .route("/:name/resolve/*path", routing::get(repositories::resolve))
-        .route("/:name/meta/", routing::get(repositories::meta))
-        .route("/:name/meta/*path", routing::get(repositories::meta))
 }
 
 async fn distributed_tabby_layer(
--- a/ee/tabby-webserver/src/lib.rs
+++ b/ee/tabby-webserver/src/lib.rs
@@ -11,6 +11,7 @@ use tracing::{error, warn};
 use websocket::WebSocketTransport;
 
 mod db;
+mod repositories;
 mod server;
 mod ui;
 mod websocket;
@@ -31,6 +32,8 @@ use schema::Schema;
 use server::ServerContext;
 use tarpc::server::{BaseChannel, Channel};
 
+use crate::repositories::repo::load_dataset;
+
 pub async fn attach_webserver(
     api: Router,
     ui: Router,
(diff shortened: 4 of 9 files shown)


ee/tabby-webserver/src/lib.rs
```diff
@@ -40,14 +42,32 @@ pub async fn attach_webserver(
     let ctx = Arc::new(ServerContext::new(conn, logger, code));
     let schema = Arc::new(create_schema());
 
+    repositories::load_dataset().await.unwrap();
+
+    let repo_routes = Router::new()
+        .route(
+            "/:name/resolve/",
```
darknight, line 49: The axum [doc](https://docs.rs/axum/latest/axum/struct.Router.html#wildcards) says `wildcard path doesn’t match empty segments`, so this path will fallback to `ui::handler` if not specify explicitly here.

ee/tabby-webserver/src/repositories.rs
```diff
@@ -0,0 +1,113 @@
...
+        let path = path.strip_prefix(&root)?;
+        let path = path.to_str().unwrap();
+        entries.push(path.to_string());
+    }
+
+    let body = Json(ListDir { entries }).into_response();
+    let resp = Response::builder()
+        .header(
+            header::CONTENT_TYPE,
+            mime::APPLICATION_OCTET_STREAM.as_ref(),
```
darknight, line 93: For dir resolve, we set special content type `application/octet-stream` even though response is json

ee/tabby-webserver/docs/api_spec.md
```diff
@@ -0,0 +1,169 @@
+# Repositories resolver API
```
wsxiaoys, line 1:
```suggestion
# API Specs

## Repository api: `/repositories`
```

ee/tabby-webserver/docs/api_spec.md
```diff
@@ -0,0 +1,169 @@
...
+
+```shell
+curl --request GET \
+  --url http://localhost:8080/repositories/https_github.com_TabbyML_tabby.git/resolve/package.json
+
+curl --request GET \
+  --url http://localhost:9090/repositories/https_github.com_TabbyML_tabby.git/resolve/ee/tabby-webserver/src/api.rs
+```
+
+## Success Responses
```
wsxiaoys, line 31: There's no need to split the documentation in such a verbose way (otherwise we should just use openapi doc), provide minimal request / response would be sufficent.

ee/tabby-webserver/src/repositories.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+#[derive(Serialize)]
+struct ListDir {
+    entries: Vec<String>,
+}
+
+/// Load dataset
+pub async fn load_dataset() -> Result<()> {
+    // `SourceFile::all()` depends on `std::io`, so it's blocking.
```
wsxiaoys, line 58: Maybe refactor SourceFile::all into non-blocking (await style)?

ee/tabby-webserver/src/lib.rs
```diff
@@ -40,14 +42,32 @@ pub async fn attach_webserver(
     let ctx = Arc::new(ServerContext::new(conn, logger, code));
     let schema = Arc::new(create_schema());
 
+    repositories::load_dataset().await.unwrap();
```
wsxiaoys, line 45: Extract into a function

ee/tabby-webserver/src/routes/repositories.rs
wsxiaoys, line 1: I suggest put this simply being `src/repositories.rs`, and routes in webserver are not expected to be reused at the moment.

Refactoring can always be done in future, I usually prefer not creating abstraction in first implementation.

the existing `repositories.rs` can be a submodule under `lib/repositories/`

ee/tabby-webserver/src/routes/repositories.rs
```diff
@@ -0,0 +1,66 @@
...
+    match resolve_file(root, &repo).await {
+        Ok(resp) => Ok(resp),
+        Err(err) => {
+            warn!("{}", err);
+            Err(StatusCode::INTERNAL_SERVER_ERROR)
+        }
+    }
+}
+
+#[utoipa::path(
```
wsxiaoys, line 47: ditto

ee/tabby-webserver/src/routes/repositories.rs
```diff
@@ -0,0 +1,66 @@
...
+    get,
+    path = "/repositories/{name}/resolve/{path}",
+    operation_id = "resolve",
+    responses(
+        (status = 200, description = "Success"),
+        (status = 404, description = "Not Found"),
+        (status = 500, description = "Internal Server Error"),
+    )
+)]
```
wsxiaoys, line 17: There's no need to provide utoipa documentation since we're not exposing the api doc

ee/tabby-webserver/src/lib.rs
wsxiaoys: Let's create vendor mime type: `application/vnd.directory+json`

ee/tabby-webserver/src/repositories.rs
wsxiaoys: Let's create vendor mime type: `application/vnd.directory+json`

ee/tabby-webserver/Cargo.toml
```diff
@@ -29,8 +30,11 @@ thiserror.workspace = true
 tokio = { workspace = true, features = ["fs"] }
 tokio-rusqlite = "0.4.0"
 tokio-tungstenite = "0.20.1"
+tower = { version = "0.4", features = ["util"] }
+tower-http = { version = "0.4.0", features = ["fs", "trace"] }
 tracing.workspace = true
 unicase = "2.7.0"
+utoipa = { workspace = true, features = ["axum_extras", "preserve_order"] }
```
wsxiaoys, line 37: remove

ee/tabby-webserver/Cargo.toml
```diff
@@ -29,8 +30,11 @@ thiserror.workspace = true
 tokio = { workspace = true, features = ["fs"] }
 tokio-rusqlite = "0.4.0"
 tokio-tungstenite = "0.20.1"
+tower = { version = "0.4", features = ["util"] }
```
wsxiaoys, line 33: is this necessary?

ee/tabby-webserver/Cargo.toml
```diff
@@ -11,6 +11,7 @@ async-trait.workspace = true
 axum = { workspace = true, features = ["ws"] }
 bincode = "1.3.3"
 chrono = "0.4"
+filenamify = "0.1.0"
```
wsxiaoys, line 14: I suggest have a unified place to convert a git url to name - current implementation let `tabby-scheduler` and `tabby-webserver` each rolls their own implementation

ee/tabby-webserver/src/repositories.rs
wsxiaoys: (can be individual PR)

ee/tabby-webserver/Cargo.toml
darknight: The `oneshot` method used in `resolve_file` method comes from `tower::ServiceExt` trait, so it's necessary.

ee/tabby-webserver/src/lib.rs
darknight: I didn't get your point, `load_dataset` is already one single function call, do you mean to put it in another function then log the error?

wsxiaoys: Hey - i mean that changes between line 45 to 61 can be extracted into a function?

ee/tabby-webserver/src/repositories.rs
darknight: As I checked, there's no async counterpart of `glob::glob()`.
Even though we open file and create reader asynchronously, there still exists blocking call.

crates/tabby-common/src/config.rs
```diff
@@ -73,9 +73,13 @@ impl Default for ServerConfig {
     }
 }
 
+pub fn to_filename<S: AsRef<str>>(s: S) -> String {
```
wsxiaoys, line 76: would it easier to read if we put `to_filename` a method of `RepositoryConfig`? e.g `name`.

ee/tabby-webserver/src/repositories.rs
```diff
@@ -0,0 +1,49 @@
+pub(crate) mod repo;
```
wsxiaoys, line 1:
```suggestion
mod repo;
```

ee/tabby-webserver/src/lib.rs
```diff
@@ -57,6 +61,16 @@ pub async fn attach_webserver(
     (api, ui)
 }
 
+async fn repositories_routers() -> Router {
```
wsxiaoys, line 64: move this function to `repositories.rs`

ee/tabby-webserver/src/repositories/repo.rs
wsxiaoys, line 1: rename to `resolve.rs`

ee/tabby-webserver/src/repositories/repo.rs
```diff
@@ -0,0 +1,122 @@
...
+};
+use hyper::Body;
+use serde::{Deserialize, Serialize};
+use tabby_common::{to_filename, SourceFile};
+use tokio::sync::OnceCell;
+use tower::ServiceExt;
+use tower_http::services::ServeDir;
+use tracing::error;
+
+pub(crate) static DATASET: OnceCell<HashMap<DatasetKey, SourceFile>> = OnceCell::const_new();
```
wsxiaoys, line 18: Instead of exporting a static variable, expose a function with following signature:
```
pub async fn resolve_meta(...) -> Meta
```

ee/tabby-webserver/src/repositories/repo.rs
```diff
@@ -0,0 +1,122 @@
...
+                return dataset;
+            }
+        };
+        for mut file in iter {
+            let key = DatasetKey {
+                local_name: to_filename(file.git_url.as_str()),
+                rel_path: file.filepath.clone(),
+            };
+            // exclude content from response data
+            file.content = "".to_string();
```
wsxiaoys, line 75: Do not reuse tabby_common::SouceFile, define response type as new type

wsxiaoys (changes requested):
Mostly coding style / api style suggestions

crates/tabby-common/src/config.rs
```diff
@@ -50,13 +50,17 @@ impl RepositoryConfig {
...
     }
 
     pub fn is_local_dir(&self) -> bool {
         self.git_url.starts_with("file://")
     }
+
+    pub fn name<S: AsRef<str>>(s: S) -> String {
+        filenamify(s)
+    }
 }
```
wsxiaoys, line 64: can this be
```
pub fn name(&self) -> String {
  filenamify(self.git_url)
}

```

wsxiaoys, line 64: can this be
```
pub fn name(&self) -> String {
  filenamify(self.git_url)
}

```

darknight: this method is also used in `load_dataset` to generate `DatasetKey`, current implementation is:
```
let key = DatasetKey {
    local_name: RepositoryConfig::name(file.git_url.as_str()),
    rel_path: file.filepath.clone(),
};
```

If we define `name` like what you described, above implementation must be changed to:
```
let conf = RepositoryConfig { git_url: file.git_url.clone() };
let key = DatasetKey {
    local_name: conf.name(),
    rel_path: file.filepath.clone(),
};
```

I'm ok with it, what's your opinion?

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -0,0 +1,154 @@
...
+            max_line_length: file.max_line_length,
+            avg_line_length: file.avg_line_length,
+            alphanum_fraction: file.alphanum_fraction,
+            tags: file.tags,
+        }
+    }
+}
+
+/// Load dataset
+pub async fn load_dataset() -> Result<()> {
```
wsxiaoys, line 82:
```suggestion
async fn load_meta() -> Result<()> {
```

This function is no longer needed to be public, and the static variable can be initialized with `lazy_static`

crates/tabby-common/src/config.rs
wsxiaoys: I think there's a missing mapping between git_url to RepositoryConfig - I suggest in load_dataset we simply fetch `Config::load()` and build a map between git url to repository config, then fetch the name.

In future we might introduce a configurable `name` field inside of `RepositoryConfig`, this refactor will also make it easier to happen.

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -0,0 +1,154 @@
...
+};
+use hyper::Body;
+use serde::{Deserialize, Serialize};
+use tabby_common::{config::RepositoryConfig, SourceFile, Tag};
+use tokio::sync::OnceCell;
+use tower::ServiceExt;
+use tower_http::services::ServeDir;
+use tracing::error;
+
+static DATASET: OnceCell<HashMap<DatasetKey, Meta>> = OnceCell::const_new();
```
wsxiaoys, line 18: use `lazy_static`

wsxiaoys (changes requested):
Otherwise LGTM

ee/tabby-webserver/src/repositories/resolve.rs
darknight: `load_meta` is used in `repositories.rs` file which requires it to be `public`, or compiler is unhappy.

I narrow down it to `pub(crate)` anyway.

darknight:
Btw, the ci always failed on `Install Rust` step.

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -0,0 +1,170 @@
...
+use lazy_static::lazy_static;
+use serde::{Deserialize, Serialize};
+use tabby_common::{config::Config, SourceFile, Tag};
+use tokio::sync::OnceCell;
+use tower::ServiceExt;
+use tower_http::services::ServeDir;
+use tracing::error;
+
+lazy_static! {
+    static ref DATASET: OnceCell<HashMap<DatasetKey, Meta>> = OnceCell::const_new();
```
wsxiaoys, line 20:
```suggestion
    static ref DATASET: HashMap<DatasetKey, Meta> = load_meta();
```

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -0,0 +1,170 @@
...
+                dataset.insert(key, file.into());
+            }
+        }
+        dataset
+    })
+    .await?;
+
+    DATASET.set(dataset)?;
+    Ok(())
+}
```
wsxiaoys, line 123:
```suggestion
/// TODO: implement autoreloading logic in future (so tabby process could capture the file changes produced by  tabby-scheduler command)
async fn load_meta() -> HashMap<DatasetKey, Meta> {
    let mut dataset = HashMap::new();
    let repo_conf = match Config::load() {
        Ok(config) => config
            .repositories
            .into_iter()
            .map(|repo| (repo.git_url.clone(), repo))
            .collect::<HashMap<_, _>>(),
        Err(err) => {
            error!("load config: {}", err);
            return dataset;
        }
    };
    let iter = match SourceFile::all() {
        Ok(all) => all,
        Err(err) => {
            error!("load dataset: {}", err);
            return dataset;
        }
    };
    for file in iter {
        if let Some(name) = repo_conf.get(&file.git_url).map(|repo| repo.name()) {
            let key = DatasetKey {
                local_name: name,
                rel_path: file.filepath.clone(),
            };
            dataset.insert(key, file.into());
        }
    }
    dataset
}
```

wsxiaoys (changes requested):

ee/tabby-webserver/src/repositories.rs
```diff
@@ -0,0 +1,58 @@
...
+use axum::{extract::Path, http::StatusCode, response::Response, routing, Json, Router};
+use tabby_common::path::repositories_dir;
+use tracing::{debug, instrument, warn};
+
+use crate::{
+    repositories,
+    repositories::resolve::{resolve_dir, resolve_file, resolve_meta, Meta, Repository},
+};
+
+pub async fn routers() -> Router {
```
wsxiaoys, line 13:
```suggestion
pub fn routers() -> Router {
```

ee/tabby-webserver/src/repositories.rs
```diff
@@ -0,0 +1,58 @@
...
+        Err(err) => {
+            warn!("{}", err);
+            Err(StatusCode::INTERNAL_SERVER_ERROR)
+        }
+    }
+}
+
+#[instrument(skip(repo))]
+async fn meta(Path(repo): Path<Repository>) -> Result<Json<Meta>, StatusCode> {
+    debug!("repo: {:?}", repo);
```
wsxiaoys, line 52:
```suggestion
```

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -0,0 +1,159 @@
...
+use tabby_common::{config::Config, SourceFile, Tag};
+use tower::ServiceExt;
+use tower_http::services::ServeDir;
+use tracing::error;
+
+lazy_static! {
+    static ref DATASET: HashMap<DatasetKey, Meta> = load_meta();
+}
+
+const MIME_VENDOR: &str = "application/vnd.directory+json";
```
wsxiaoys, line 22:
```suggestion
const DIRECTORY_MIME_TYPE: &str = "application/vnd.directory+json";
```

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -0,0 +1,159 @@
...
+            language: file.language,
+            max_line_length: file.max_line_length,
+            avg_line_length: file.avg_line_length,
+            alphanum_fraction: file.alphanum_fraction,
+            tags: file.tags,
+        }
+    }
+}
+
+/// TODO: implement autoreloading logic in future (so tabby process could capture the file changes produced by  tabby-scheduler command)
```
wsxiaoys, line 83:
```suggestion
/// TODO: implement auto reloading logic in future (so changes produced by tabby-scheduler command will be loaded)
```

wsxiaoys (changes requested):
Last round of nits!

wsxiaoys:
Please also fix the ci

ee/tabby-webserver/src/repositories.rs
```diff
@@ -0,0 +1,56 @@
...
+    let is_dir = tokio::fs::metadata(full_path.clone())
+        .await
+        .map(|m| m.is_dir())
+        .unwrap_or(false);
+
+    if is_dir {
+        return match resolve_dir(root, full_path).await {
+            Ok(resp) => Ok(resp),
+            Err(err) => {
+                warn!("{}", err);
```
wsxiaoys, line 34: Seems major case of error are 404 - not found? Then it's not make sense to write a warning log and returns 500?

ee/tabby-webserver/src/repositories.rs
```diff
@@ -0,0 +1,56 @@
...
+                Err(StatusCode::INTERNAL_SERVER_ERROR)
+            }
+        };
+    }
+
+    match resolve_file(root, &repo).await {
+        Ok(resp) => Ok(resp),
+        Err(err) => {
+            warn!("{}", err);
```
wsxiaoys, line 43: ditto

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -0,0 +1,159 @@
...
+fn load_meta() -> HashMap<DatasetKey, Meta> {
+    let mut dataset = HashMap::new();
+    let repo_conf = match Config::load() {
+        Ok(config) => config
+            .repositories
+            .into_iter()
+            .map(|repo| (repo.git_url.clone(), repo))
+            .collect::<HashMap<_, _>>(),
+        Err(err) => {
+            error!("load config: {}", err);
```
wsxiaoys, line 93: This will write an error log on for all tabby instance without a config file, which is not desired.

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -0,0 +1,159 @@
...
+            return dataset;
+        }
+    };
+    let iter = match SourceFile::all() {
+        Ok(all) => all,
+        Err(err) => {
+            error!("load dataset: {}", err);
```
wsxiaoys, line 100: ditto

wsxiaoys (changes requested):

ee/tabby-webserver/src/repositories.rs
darknight: This Err branch is to handle **io** error which may occur on `read_dir`, `read_dir.next_entry()`.
We call `resolve_dir` only if the path **is** dir and exists, so there's no 404 happen here.

If path is invalid dir, `is_dir` will be `false`, so it's fallback to `resolve_file`.

ee/tabby-webserver/src/repositories.rs
darknight: This Err branch is to take care of `oneshot` call, which return a `Result`.

If file doesn't exist, `onshot` return an empty response with `404` status code, so it is on `Ok` branch. Err branch is for other unexpected cases, so I set `500` status code.

ee/tabby-webserver/src/repositories/resolve.rs
darknight: Previously, `load_meta` is run in a tokio task, so logging error, imo, is necessary.

Since we load data directly now, what's the expected behavior here?

ee/tabby-webserver/src/repositories.rs
wsxiaoys:
```suggestion
                warn!("failed to resolve_dir <{}>: {}", full_path, err);
```

I see, polishing the error message a bit

ee/tabby-webserver/src/repositories/resolve.rs
wsxiaoys: I'll say simply ignoring it

wsxiaoys (approved):

PR #864, opened 2023-11-22
fix: Skip creating `usage_anonymous_id` when TABBY_DISABLE_USAGE_COLLECTION is set

praktiskt:
I noticed the file `usage_anonymous_id` (containing the anonymous identifier) was written to disk even when tracking has been disabled. This seems unnecessary since it's not used as far as I can tell.

### DIFF
--- a/crates/tabby-common/src/usage.rs
+++ b/crates/tabby-common/src/usage.rs
@@ -16,6 +16,26 @@ struct UsageTracker {
 
 impl UsageTracker {
     fn new() -> Self {
+        if std::env::var("TABBY_DISABLE_USAGE_COLLECTION").is_ok() {
+            eprintln!(
+                "
+  \x1b[34;1mTELEMETRY\x1b[0m
+
+  TABBY_DISABLE_USAGE_COLLECTION is set and telemetry has been disabled.
+
+  \x1b[1mWelcome to Tabby!\x1b[0m
+
+  If you have any questions or would like to engage with the Tabby team, please join us on Slack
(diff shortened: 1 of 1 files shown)


crates/tabby-common/src/usage.rs
```diff
@@ -16,6 +16,26 @@ struct UsageTracker {
...
+
+  TABBY_DISABLE_USAGE_COLLECTION is set and telemetry has been disabled.
+
+  \x1b[1mWelcome to Tabby!\x1b[0m
+
+  If you have any questions or would like to engage with the Tabby team, please join us on Slack
+  (https://tinyurl.com/35sv9kz2).
+
+"
+            );
```
wsxiaoys, line 32:
```suggestion
```

Outputing this for every launch seems a bit verbose to me.

crates/tabby-common/src/usage.rs
```diff
@@ -16,6 +16,26 @@ struct UsageTracker {
...
+            return Self {
+                id: String::from("00000000-0000-0000-000000000000"),
```
wsxiaoys, line 34: Make the field optional, instead of setting a dummy value.

wsxiaoys (changes requested):
Thanks for the PR. Agree it's better the anonymous id file is not created if the env var being set.

Adding a few comments.

crates/tabby-common/src/usage.rs
praktiskt: :+1:

PR #885, opened 2023-11-24
feat(agent): support read clipboard as prompt

icycodes:
Complete TAB-351


clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -66,6 +70,10 @@ export const defaultAgentConfig: AgentConfig = {
       experimentalStripAutoClosingCharacters: false,
       maxPrefixLines: 20,
       maxSuffixLines: 20,
+      clipboard: {
+        minChars: 3,
+        maxChars: 1000,
```
wsxiaoys, line 75: default to 2000

PR #895, opened 2023-11-25
feat: Add support Intel

### LINKED ISSUE #631: Intel Arc / XPU support
I would be great to be able to run Tabby locally on my Intel Arc GPU. 

**Additional context**

This is currently possible in tools like llama.cpp by compiling with OpenCL support. I have no idea how that would (or could) translate to Rust.

---

cromefire:
Realized via Intel MKL (oneAPI)

Fixes #631 (Intel)

Depends on #902


cromefire:
Current state is that the intel image build, just returns 501 on the API call without any logged error at all. The AMD image doesn't quite build yet, it has some sort of error in the linking stage:
```
25.06   = note: /usr/bin/ld: /root/workspace/target/release/build/llama-cpp-bindings-f80cf7b588122741/out/build/libllama.a(ggml-cuda.cu.o): undefined reference to symbol 'rocblas_initialize'
25.06           /usr/bin/ld: /opt/rocm-5.7.0/lib/librocblas.so.3: error adding symbols: DSO missing from command line
25.06           collect2: error: ld returned 1 exit status
```

I added the arg, but it doesn't seem to want to apply it:
```rust
println!("cargo:rustc-link-arg=-Wl,--copy-dt-needed-entries");
```

cromefire:
Okay, gut the 501 under control by adding options for OneAPI and ROCm to it, now I just gotta test it (it runs at least, but not sure whether it's using the GPU) and I somehow need to get the ROCm build under control.

cromefire:
So in theory the intel container should work now... but for some reason it just doesn't want to use the GPU...

wsxiaoys:
> So in theory the intel container should work now... but for some reason it just doesn't want to use the GPU...

is there sth like NVIDIA container toolkit need to be installed for oneAPI?

cromefire:
Nope, just pass through `/dev/dri` (with that `sycl-ls` correctly reports the GPU, but llama.cpp doesn't seem to actually use it)

cromefire:
So the ROCm image is now definitely working. `TabbyML/DeepseekCoder-6.7B` was just locking up the GPU entierly for some reason, but `TabbyML/DeepseekCoder-1.3B` runs inference in like 5ms on my RX 7900 XTX.

wsxiaoys:
> So the ROCm image is now definitely working. `TabbyML/DeepseekCoder-6.7B` was just locking up the GPU entierly for some reason, but `TabbyML/DeepseekCoder-1.3B` runs inference in like 5ms on my RX 7900 XTX.

Great! You might consider extracting ROCm as individual PR for review to get it checked in.

cromefire:
Yeah I'll see tomorrow whether I can get a handle on OneAPI or whether I'll postpone that and extract ROCm, but it's 4:30 AM for me, so I really need to do that tomorrow (technically later today).

Also I really hate C and it's library linking nonsense... that cost me so much time with this....

cromefire:
Also note to my future self: I need to figure out whats happening with the cuda_devices list and the Frontend and match that for ROCm and oneAPI if possible.

cromefire:
@wsxiaoys also would be great I the intellj extension would be available for Rust Rover, then I could write Tabby code using Tabby. Probably just a setting or so.

cromefire:
AMD stuff is "moved" to #902, because it already works pretty okay

wsxiaoys:
> Could we get llama.cpp as a shared library or so? That sounds way easier

To confirm, you've been able to make llama.cpp itself work on Intel Arc, but not for tabby, correct?

icycodes:
> @wsxiaoys also would be great I the intellj extension would be available for Rust Rover, then I could write Tabby code using Tabby. Probably just a setting or so.

Hi, @cromefire 
Thank you for this suggestion. I noticed that the latest Rust Rover preview version is v233, but the Tabby plugin's metadata currently states that it supports versions v222-232. We should update this range in the next release.
If you want to try it out before the next release, you can build the plugin locally and install it from a file.  Related: #903

itlackey:
It would brle great! I am hoping one or the other get merged soon.

BTW here are some things I put together to test llama.cpp on Arc. The logs so the current speeds I am getting.
https://github.com/itlackey/llama.cpp-opencl

cromefire:
> The logs so the current speeds I am getting.

Have you tried SYCL vs. Vulkan vs. OpenCL by any chance? (If they actually already run...)

wsxiaoys:
Closing as vulkan support will be released in 0.10

PR #902, opened 2023-11-26
feat: Add Support for Devices other than CUDA in Telemetry and Web UI

### LINKED ISSUE #636: AMD GPUs (ROCm) support
I'd be a great addition if we could also run the project GPU-accelerated in AMD GPUs too. Thanks!

---

cromefire:
Realized via AMD hipBLAS (ROCm/HIP)

Fixes #636


cromefire:
Okay, the GPU name is also in the UI now, it's more like the ID currently, but I've opened an issue for the library to get the marketing (user recognizable) name: PTFOPlayer/rocm_lib#2

cromefire:
Documentation is also there now, it should be pretty okay-ish for a start.

.github/workflows/docker.yml
wsxiaoys, line 1: Creat a new workflow file for rocm build

wsxiaoys (changes requested):
Thanks for the PR.

First round of comments focusing on source code

crates/tabby/src/main.rs
```diff
@@ -1,22 +1,23 @@
...
 use opentelemetry::{
     global,
     sdk::{propagation::TraceContextPropagator, trace, trace::Sampler, Resource},
     KeyValue,
 };
 use opentelemetry_otlp::WithExportConfig;
-use tabby_common::config::Config;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter, Layer};
 
+use tabby_common::config::Config;
```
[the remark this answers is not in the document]
cromefire: This seems to be from your rustfmt config. rustfmt always seems to change it back if I try to commit it:
image
image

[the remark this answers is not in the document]
wsxiaoys: I believe the issue arises from the difference in configuration for rustfmt. Nonetheless, I would prefer to keep the formatting change as a separate and isolated modification.

crates/tabby/src/services/health.rs
```diff
@@ -18,6 +23,8 @@ pub struct HealthState {
     arch: String,
     cpu_info: String,
     cpu_count: usize,
+    gpu_devices: Vec<String>,
+    #[deprecated(note = "Use the more generic gpu_devices instead")]
```
wsxiaoys, line 27:
```suggestion
```

Let's don't deprecate the field. There're downstream usage of this field relying on it to detect nvidia card.

crates/tabby/src/services/health.rs
```diff
@@ -26,19 +33,21 @@ impl HealthState {
...
         Self {
             model: model.map(|x| x.to_owned()),
             chat_model: chat_model.map(|x| x.to_owned()),
             device: device.to_string(),
             arch: ARCH.to_string(),
             cpu_info,
             cpu_count,
-            cuda_devices,
+            gpu_devices: gpu_devices.clone(),
+            cuda_devices: gpu_devices,
```
wsxiaoys, line 50: Since you added a new field `gpu_device`, let's keep cuda_devices' original semantic (only include nvidia cuda devices).

ee/tabby-ui/app/(dashboard)/components/worker-card.tsx
```diff
@@ -78,8 +78,37 @@ export default function RunnerCard({
...
+                <path d="M8 11V9" />
+                <path d="M16 11V9" />
+                <path d="M12 11V9" />
+                <path d="M2 15h20" />
+                <path d="M2 7a2 2 0 0 1 2-2h16a2 2 0 0 1 2 2v1.1a2 2 0 0 0 0 3.837V17a2 2 0 0 1-2 2H4a2 2 0 0 1-2-2v-5.1a2 2 0 0 0 0-3.837Z" />
+              </svg>
+              <p title={x} className={textClass}>{x}</p>
+            </Info>
+          ))}
+        {device == 'rocm' &&
```
wsxiaoys, line 109: there's no need to separate rocm and cuda display in worker-card.

```suggestion
        {device == 'rocm' || device == 'cuda'
```

Remove line 81 ~ 108

ee/tabby-ui/app/(dashboard)/page.tsx
```diff
@@ -1,24 +1,17 @@
...
-import { useWorkers } from '@/lib/hooks/use-workers'
-import { WorkerKind } from '@/lib/gql/generates/graphql'
-import { useGraphQL } from '@/lib/hooks/use-graphql'
-import { getRegistrationTokenDocument } from '@/lib/gql/request-documents'
-import { CopyButton } from '@/components/copy-button'
+import {useWorkers} from '@/lib/hooks/use-workers'
+import {WorkerKind} from '@/lib/gql/generates/graphql'
+import {useGraphQL} from '@/lib/hooks/use-graphql'
+import {getRegistrationTokenDocument} from '@/lib/gql/request-documents'
+import {CopyButton} from '@/components/copy-button'
```
wsxiaoys, line 14: Don't mix formatting changes. (similar comments on other ts files with import style changed)

Probably we should maintain a .editorconfig to avoid situations like this in future.

wsxiaoys (changes requested):

crates/tabby/src/services/health.rs
cromefire: That's what the device type is for though, having 2 ways of getting the information is pretty unclean and will inevitably lead to issues. The original field wasn't even correct. And just because it's deprecated doesn't mean it's going away anytime soon. And do you want to maintain fields for every single HW vendor? It's 2 now, maybe 3 soon and the way things go it'll be 10s in the future probably, so having to maintain vendor specific fields for each won't make sense.

Keep the cuda devices for compatibility (hence it being deprecated, not removed) and let's move to a more generic approach. If that implementation doesn't suit you, you could also think about changing to `gpu_devices` list to containing a device struct, that has like a name, serial number, id and type or so, that would probably be the cleanest one.

crates/tabby/src/services/health.rs
cromefire: Sure can do that if you really want... originally it was also displaying cuda devices though that weren't being used. So my thought is that applications not updated to properly check for non-CUDA devices yet, will just display them as cuda devices for now, which is technically wrong, but at least they are displaying them and don't just silently drop them. I think that would be better for users of non-NVIDIA GPUs than the software displaying nothing at all, but if you see that differently, I can change that.

ee/tabby-ui/app/(dashboard)/page.tsx
cromefire: Yeah otherwise IntelliJ IDEs will just format everything according to their code style as you can see. They do prefer keeping everything as they want if not told otherwise. I'd recommend adding prettier or so.

crates/tabby/src/services/health.rs
wsxiaoys: > The original field wasn't even correct. 

Could you provide further details on why the original field was incorrect?

The rationale for not deprecating it is straightforward: we currently have downstream compoent that relies on this specific field and is not likely to be modified in the immediate future. Therefore, I propose the following options:

> I'm fine to either:
> adding rocm device to the list directly (Since it uses a cuda ish api)
> or adding a new field gpu_devices, while keeping cuda_devices as is

Maintaining support for the cuda_devices field is crucial for these downstream components, and it is imperative that we preserve it.

ee/tabby-ui/app/(dashboard)/page.tsx
cromefire: To be clear this is not prettier, this is just IntelliJ formatting every import that it touches, which it does if there aren't prettier or a properly configured editorconfig, which is exactly why it was good to have one. I didn't change it on purpose, but there are no defaults here, so the IDE used its own defaults.

crates/tabby/src/main.rs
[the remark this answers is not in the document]
cromefire: Well the configuration is set from the `.rustfmt.yaml` file in the repo, so it would be the same. I'll turn it off for now, but that of course means I can't guarantee that the code is properly formatted.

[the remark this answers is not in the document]
cromefire: Okay figured out that your config requires that you have rustfmt nightly, so now it should be better

crates/tabby/src/services/health.rs
cromefire: Fixed

ee/tabby-ui/app/(dashboard)/page.tsx
cromefire: Should be better now after I manually adjusted some Rust Rover configuration

ee/tabby-ui/app/(dashboard)/components/worker-card.tsx
cromefire: Way better now with the refactoring

crates/tabby-common/src/api/accelerator.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(Serialize, Deserialize, GraphQLEnum, ToSchema, PartialEq, Clone, Debug)]
+pub enum DeviceType {
+    Cuda,
+    Rocm,
+}
+
+#[derive(Serialize, Deserialize, GraphQLObject, ToSchema, Clone, Debug)]
+pub struct Accelerator {
+    /// Universally unique ID of the accelerator, if available
+    pub uuid: Option<String>,
```
cromefire, line 14: The biggest issue now is whether we can prevent this field from being part of the telemetry. If not, it definitely has to be removed, as it's basically almost equivalent to a serial number. It's nice for tracking a GPU in the accelerators, but it's a huge issue if that data leaves your control.

cromefire:
As for the why I use `rocminfo` instead of a lib that there's no rust wrapper for `libhsa-runtime64` and I treid wring my own quickly, but due to C being C it seems to create just more problems than it solves. And then you have to link it again, which is rather fragile or load in on runtime, which is more of a guessing game, while `rocminfo` should be available whenever rocm is installed and is just on the path and just works.

wsxiaoys:
Hey, considering the increasing size of this pull request, I have extracted #913 to focus on the basic ROCm support as a preliminary step

Amzd:
The other PR didn't add any doc but I tried the command in documentation of this PR but couldn't get it to work, is there a trick to it I'm missing?
> docker run -it \
>   --device /dev/dri --device /dev/kfd \
>   -p 8080:8080 -v $HOME/.tabby:/data \
>   tabbyml/tabby-rocm \
>   serve --model TabbyML/StarCoder-1B --device rocm

cromefire:
> The other PR didn't add any doc but I tried the command in documentation of this PR but couldn't get it to work, is there a trick to it I'm missing?
> > docker run -it \
> >   --device /dev/dri --device /dev/kfd \
> >   -p 8080:8080 -v $HOME/.tabby:/data \
> >   tabbyml/tabby-rocm \
> >   serve --model TabbyML/StarCoder-1B --device rocm

The other PR didn't include the docker image, you'll have to build it yourself

Frozen-byte:
When I try to build via docker-compose build I get the following error:
```
8.092    Compiling llama-cpp-bindings v0.6.0-dev (/root/workspace/crates/llama-cpp-bindings)
8.096 error: failed to run custom build command for `llama-cpp-bindings v0.6.0-dev (/root/workspace/crates/llama-cpp-bindings)`
8.096 
8.096 Caused by:
8.096   process didn't exit successfully: `/root/workspace/target/release/build/llama-cpp-bindings-65e03e49d2a060b2/build-script-build` (exit status: 101)
8.096   --- stderr
8.096   thread 'main' panicked at crates/llama-cpp-bindings/build.rs:8:5:
8.096   Please init submodules with `git submodule update --init --recursive` and try again
8.096   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
8.096 warning: build failed, waiting for other jobs to finish...

```
Still investigating why this happens, maybe you provide a faster solution as my research.

My docker-compose.yml file:
```
version: '3.5'

services:
  tabby:
    build:
      context: .
      dockerfile: rocm.Dockerfile
    command: serve  --model TabbyML/SantaCoder-1B --device rocm
    volumes:
      - "$HOME/Docker/tabby/data:/data"
    ports:
      - 8229:8080
    environment:
      - TABBY_DISABLE_USAGE_COLLECTION=1
    devices:
      - /dev/dri
      - /dev/kfd
```

Frozen-byte:
I am sadly unable to run this docker:
```
$ docker-compose up 
Recreating tabby_tabby_1 ... done
Attaching to tabby_tabby_1
tabby_1  | 2023-12-09T17:02:11.296325Z  INFO tabby::serve: crates/tabby/src/serve.rs:105: Starting server, this might takes a few minutes...
tabby_1  | ggml_init_cublas: GGML_CUDA_FORCE_MMQ:   no
tabby_1  | ggml_init_cublas: CUDA_USE_TENSOR_CORES: yes
tabby_1  | ggml_init_cublas: found 2 ROCm devices:
tabby_1  |   Device 0: AMD Radeon RX 6900 XT, compute capability 10.3
tabby_1  |   Device 1: AMD Radeon Graphics, compute capability 10.3
tabby_1  | ggml_cuda_set_main_device: using device 0 (AMD Radeon RX 6900 XT) as main device
tabby_1  | 
tabby_1  | CUDA error 98 at /root/workspace/crates/llama-cpp-bindings/llama.cpp/ggml-cuda.cu:6779: invalid device function
tabby_1  | current device: 0
tabby_tabby_1 exited with code 1
```

my docker-compose.yml:
```
version: '3.5'

services:
  tabby:
    build:
      context: .
      dockerfile: rocm.Dockerfile
    command: serve --model TabbyML/StarCoder-1B --device rocm
    volumes:
      - "$HOME/Docker/tabby/data:/data"
    ports:
      - 8229:8080
    environment:
      - TABBY_DISABLE_USAGE_COLLECTION=1
    devices:
      - /dev/dri
      - /dev/kfd
```

cromefire:
You'll probably need to remove your iGPU from the container, ROCm sometimes has some problems with that (alternatively I think in the llama.cpp docs it also called out an environment variable to make it only use a specific GPU).

cromefire:
So updating this from main starts to get complicated now, because there's conflicts in a lot of places, so it would be great to get this in like some time soon... I'm trying to fix it up now, but I'm not sure how long I can do this, it's quite time consuming.

Frozen-byte:
> You'll probably need to remove your iGPU from the container, ROCm sometimes has some problems with that (alternatively I think in the llama.cpp docs it also called out an environment variable to make it only use a specific GPU).

Thank you! I can verify that pinning devices works perfectly now:
```
    devices:
      - /dev/dri/card0
      - /dev/dri/renderD128
      - /dev/kfd
```

cromefire:
Also the llama.cpp fork should probably be updated sooner than later, as there's some fixes for AMD GPU performance in there.

wsxiaoys:
Hey @cromefire - I would suggest first have a PR of building docker image / binary checkin first, then we can review how to integrate to rocm device information into dashboard.

cromefire:
Sure. I'll extract that.

PS: Also keep in mind that it won't only be ROCm device info soon, but probably oneAPI and/or Vulkan as well.

cromefire:
Just removed the conflicts and maybe another friendly reminder that these changes suck up a lot of time the longer they lie around and I constantly have to resolve the conflicts...

JumpLink:
super nice, when will the PR be merged?

PR #905, opened 2023-11-27
fix: run container as non-root

praktiskt:
First stab at running the Tabby container as non-root. 

I've not explored thoroughly if there are other directories that needs to be owned by the `tabby` user. I am able to run a simple serve (`tabby serve --model ...`) with this setup.

Hoping CI will tell me more, if there is more.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,5 +61,10 @@ RUN ln -s /usr/lib/x86_64-linux-gnu/libnvidia-ml.so.1 \
 COPY --from=build /opt/tabby /opt/tabby
 
 ENV TABBY_ROOT=/data
+ARG WRITE_DIRS="${TABBY_ROOT} ${TABBY_ROOT}/events"
+RUN useradd -u 1001 tabby && \
+    mkdir -p ${WRITE_DIRS} && \
+    chown tabby ${WRITE_DIRS}
+USER tabby
 
 ENTRYPOINT ["/opt/tabby/bin/tabby"]


Dockerfile
```diff
@@ -61,5 +61,10 @@ RUN ln -s /usr/lib/x86_64-linux-gnu/libnvidia-ml.so.1 \
 COPY --from=build /opt/tabby /opt/tabby
 
 ENV TABBY_ROOT=/data
+ARG WRITE_DIRS="${TABBY_ROOT} ${TABBY_ROOT}/events"
```
wsxiaoys, line 64: This won't work - as $TABBY_ROOT could be mounted to external directory, which might not writable to 1001 uid.

praktiskt: Yes, this was the intended effect. To me it makes sense if the container don't write to the user file system as root, even if its mounted. 

Any alternate takes on how to solve this? Do you agree?

wsxiaoys: The tradeoff being considered here is that we want to avoid inconveniencing the first-time user with the manual execution of `chmod` (as they are likely running `docker -it` directly).

One solution I have implemented is to create a new Docker image based on Tabby's distribution, similar to the one provided for Hugging Face deployment: https://huggingface.co/spaces/TabbyML/tabby-template-space/blob/main/Dockerfile

praktiskt: The idea isn't that the user should run chmod, it's that the tabby container shouldn't modify user space, especially not as root. 

I agree that it should be as smooth as possible for new users. 

Where does tabby need to write today?

wsxiaoys:
> @wsxiaoys Do you think you'll ever want to approach something like this with the project? If not, I'll close the PR so we don't leave it lingering forever.

Hey - sorry for missing the last response. As a reference, we do run tabby as non-root for huggingface space tutorials, https://huggingface.co/spaces/TabbyML/tabby-template-space/blob/main/Dockerfile

I'm still not sure whether it can be easily integrated into the repository, as we want to avoid the hassle of using 'chmod' when invoking 'docker run'.

PR #925, opened 2023-12-01
fix: generate unicode characters error

xcnick:
This PR resolves the issue of Unicode character decoding errors in the Deepseek Coder model, referencing the content in the following two links:

- https://github.com/ggerganov/llama.cpp/pull/4118
- https://github.com/shibe2/llama.cpp/blob/724dd5c2300ed6df926ddd8a1ed677337280646b/examples/server/server.cpp#L960-L986

### DIFF
--- a/crates/llama-cpp-bindings/src/engine.cc
+++ b/crates/llama-cpp-bindings/src/engine.cc
@@ -17,7 +17,6 @@ TextInferenceEngine::~TextInferenceEngine() {}
 namespace {
 constexpr size_t N_BATCH = 512;  // # per batch inference.
 constexpr size_t N_CTX = 4096;   // # max kv history.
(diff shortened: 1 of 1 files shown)


wsxiaoys:
Thanks for the pull request. I've also noticed the issue but haven't had a chance to debug and ensure it fixes the CJK decoding.

To confirm, have you built Tabby with this PR and verified that it fixes some cases for you?

xcnick:
Yes, the current version of Tabby encounters an error stating 'ignoring non utf-8/utf-16 output' when using the deepseek-coder-instruct model on the playground interface to generate code wrapped in Markdown format. However, after implementing this PR, the generation of Markdown-formatted code works correctly.

PR #935, opened 2023-12-01
refactor(webserver): cleanup user crud

wsxiaoys:


wsxiaoys:
cc @darknight

ee/tabby-webserver/src/service/db.rs
```diff
@@ -167,25 +167,38 @@ impl DbConn {
...
+                let mut stmt = c.prepare(
                     r#"INSERT INTO users (email, password_encrypted, is_admin) VALUES (?, ?, ?)"#,
-                    params![email, password_encrypted, is_admin],
-                )
+                )?;
+                let id = stmt.insert((email, password_encrypted, is_admin))?;
+                Ok(id)
             })
             .await?;
         if res != 1 {
```
darknight, line 181: This `if` check should be removed since the meaning of `res` has changed from `number of rows` to `user id`

wsxiaoys: nice catch, fixed

PR #912, opened 2023-11-28
feat(ee): implement user authentication api

darknight:
For closing https://github.com/TabbyML/tabby/issues/805

Changes include:

- Update GraphQL schema, add 4 mutations, generated schema looks like
```
type Mutation {
  register(
    email: String!
    username: String!
    password1: String!
    password2: String!
  ): RegisterResponse!
  tokenAuth(
    email: String
    username: String
    password: String!
  ): TokenAuthResponse!
  verifyToken(token: String!): VerifyTokenResponse!
}
```
The `refreshToken` implementation has not been finalized, so I do plan to raise separate PR for it.

- Update mutation `resetRegistrationToken`, only `admin` is allowed to execute it

There are some implementations need to be discussed, **comment inline**.

PR #938, opened 2023-12-02
feat: implement refresh_token API

darknight:
follow up of https://github.com/TabbyML/tabby/pull/912

Changes include:

- use uuid without hyphen as refresh token
- refresh token expires in 7 days by default
- create `refresh_tokens` table, add r/w operations
- implement refresh_token logic
- update `register` API, since it's new user, we always generate new refresh token and return it
- update `token_auth` API, first we check if there exists valid (unexpired) refresh token for current user, if not, generate new one

Pending task:
To clean up expired refresh tokens, there are two options:
1. do clean up during login (i.e. token_auth), if no valid refresh token, we generate new one, meanwhile, delete all expired ones.
2. spawn a background task when server starts, run clean up periodically

@wsxiaoys what's your suggestion?


ee/tabby-webserver/src/service/auth.rs
```diff
@@ -173,19 +177,57 @@ impl AuthenticationService for DbConn {
...
+        let Ok(access_token) = generate_jwt(Claims::new(UserInfo::new(
+            user.email.clone(),
+            user.is_admin,
+        ))) else {
+            return Err(RefreshTokenError::Unknown);
+        };
+
+        let resp = RefreshTokenResponse::new(
+            access_token,
+            refresh_token.token,
```
wsxiaoys, line 226: is it a better practice to return a updated refresh token?

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -173,19 +177,57 @@ impl AuthenticationService for DbConn {
             return Err(TokenAuthError::InvalidPassword);
         }
 
+        let utc_ts = chrono::Utc::now().timestamp();
+        let refresh_token = match self.get_user_unexpired_token(user.id, utc_ts).await? {
```
wsxiaoys, line 181: does it makes more sense to always return a new token? (So different login can be distinguished and revoked individually)

wsxiaoys:
> spawn a background task when server starts, run clean up periodically

I'll suggest go with this approach, example:
https://github.com/TabbyML/tabby/blob/main/crates/tabby-scheduler/src/lib.rs#L11

The tokio job can be spawned when creating db conn.

ee/tabby-webserver/src/service/auth.rs
darknight: That's a good question, I had the same thought.

With some google, I read a point of view that this could cause security issue since there's no way to prevent user from abusing this API, to generate huge amounts of refresh tokens.

By always returning new token, I see a few possible issues:
- user will never be logged out  (unless we revoke all of them)
- db rows grow unexpectedly
- when login, not sure which refresh token to return.
- when revoke, might cause revoke partially done
  - say user login A device with refresh token A, B device with refresh token B. when revoke A, B device still have access

Technically, it's definitely ok.

ee/tabby-webserver/src/service/auth.rs
darknight: hmm...so this is expected behavior?

wsxiaoys: Otherwise a client can just refresh access token without update its local refresh token?

ee/tabby-webserver/src/service/auth.rs
wsxiaoys: > user will never be logged out (unless we revoke all of them)
Why? A user will be logged out as long as his current refresh token expires (since he need to re-login by using token auth)

> db rows grow unexpectedly
agree

> when login, not sure which refresh token to return.
should be a new token.

> when revoke, might cause revoke partially done
> say user login A device with refresh token A, B device with refresh token B. when revoke A, B device still have access
This behavior could easily adjust given the requests:
1. If we wanna device B being logged out, we should just revoke all refresh token belong to this user.
2. If we wanna device B stayed logged in, we'll just leave it as is.

Isn't this part of the reason we make refresh token a separate table to make each of it can be revoke individually?

ee/tabby-webserver/src/service/auth.rs
wsxiaoys: note this somewhat coupled with the behavior above (always update the token with refreshing).

Here we should always create a new refresh token for each login.

ee/tabby-webserver/src/service/auth.rs
darknight: > > user will never be logged out (unless we revoke all of them)
> 
> Why? A user will be logged out as long as his current refresh token expires (since he need to re-login by using token auth)

I mean a scenario like this: say user login, got refresh token A, user do refresh with A, got new access token and new refresh token B, B expires later than A, so user can continue to use B to get access token and another refresh token C, something like that, is it possible?

> > db rows grow unexpectedly
> 
> agree, this is problematic, and seems the only way is to apply certain rate limiting.
> 
> > when login, not sure which refresh token to return.
> 
> should be a new token.
> 
> > when revoke, might cause revoke partially done
> > say user login A device with refresh token A, B device with refresh token B. when revoke A, B device still have access
> 
> This behavior could easily adjust given the requests:
> 
> 1. If we wanna device B being logged out, we should just revoke all refresh token belong to this user.
> 2. If we wanna device B stayed logged in, we'll just leave it as is.
> 
> Isn't this part of the reason we make refresh token a separate table to make each of it can be revoke individually?

hmm...ya, I remember it is. How about return new refresh token for each login (i.e. `token_auth`), but not for `refresh_token`?

We put rate limit on login request, so user won't abuse it, make sense?

wsxiaoys: > I mean a scenario like this: say user login, got refresh token A, user do refresh with A, got new access token and new refresh token B, B expires later than A, so user can continue to use B to get access token and another refresh token C, something like that, is it possible?

I think this is the average practice - and it's easy to adjust the behavior (because we're updating the same row, we control the new expire time at refresh time anyway

> How about return new refresh token for each login (i.e. token_auth), but not for refresh_token?

for refresh token - we're not creating new row in refresh token, we only update the value of `token.

> We put rate limit on login request, so user won't abuse it, make sense?

No need to implement this in this PR

ee/tabby-webserver/Cargo.toml
```diff
@@ -29,6 +29,7 @@ tabby-common = { path = "../../crates/tabby-common" }
 tarpc = { version = "0.33.0", features = ["serde-transport"] }
 thiserror.workspace = true
 tokio = { workspace = true, features = ["fs"] }
+tokio-cron-scheduler = "0.9.4"
```
darknight, line 32: Unfortunately, `job_scheduler` has no async support, but db operation is async.

So import a new lib which inspired by `job_scheduler` but with async support

wsxiaoys (commented):
Otherwise LGTM

ee/tabby-webserver/src/service/db_job.rs
wsxiaoys, line 1: Just call this `cron`

wsxiaoys (approved):
Please also:
1. Resolve conflicts.
2. Update schema file (if needed)

PR #941, opened 2023-12-03
feat: Add OpenAI API compatibility to chat completions API

itlackey:
The intent of this pull request is to update the chat completions API endpoint to be compatible with the OpenAI API specifications.


itlackey:
The initial pass has the updated response format but is not fully compatible yet. Mostly needs the first and last message in the stream to be formatted differently than the rest of the stream currently.

wsxiaoys:
fixed in https://github.com/TabbyML/tabby/pull/1109

Issue #909, opened 2023-11-28
Windows build for tabby


wsxiaoys:
llama.cpp itself is fully compatible for windows, tabby itself doesn't relies on any system specific libs, so it's possible to refactor and generates a windows native build for tabby.

For now windows users have to relies on WSL2 for native cuda support.

darknight:
I do find a platform specific dependence which is `aim = "1.8.5"` from `tabby-download`, it has no support on windows os.

Need to find an alternative I suppose.

PR #948, opened 2023-12-06
feat: support build tabby on windows

darknight:
For https://github.com/TabbyML/tabby/issues/909

The trial is partially successful, raise PR for further discussion.

See comments inline.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,19 +26,27 @@ jobs:
     container: ${{ matrix.container }}
     strategy:
       matrix:
-        binary: [aarch64-apple-darwin, x86_64-manylinux2014, x86_64-manylinux2014-cuda117]
+        binary: [aarch64-apple-darwin, x86_64-manylinux2014, x86_64-manylinux2014-cuda117, x86_64-pc-windows-msvc]
         include:
           - os: macos-latest
             target: aarch64-apple-darwin
+            raw: tabby
             binary: aarch64-apple-darwin
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
+            raw: tabby
             binary: x86_64-manylinux2014
             container: quay.io/pypa/manylinux2014_x86_64
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
+            raw: tabby
             binary: x86_64-manylinux2014-cuda117
             container: sameli/manylinux2014_x86_64_cuda_11.7
+            build_args: --features cuda
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+            raw: tabby.exe
+            binary: x86_64-pc-windows-msvc.exe
             build_args: --features cuda
 
     env:
@@ -77,13 +85,27 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
 
-      - run: bash ./ci/prepare_build_environment.sh
+      - name: Prepare build environment for macOS & Linux
+        run: bash ./ci/prepare_build_environment.sh
+        if: runner.os != 'Windows'
+
+      - name: Prepare build environment for Windows
+        run: ./ci/prepare_build_environment.ps1
+        if: runner.os == 'Windows'
+
+      - name: Install CUDA toolkit for Windows
+        uses: Jimver/cuda-toolkit@v0.2.11
+        with:
+          cuda: '11.7.1'
+          method: 'network'
+          sub-packages: '["nvcc", "cudart", "cublas", "cublas_dev", "thrust", "visual_studio_integration"]'
+        if: runner.os == 'Windows'
 
       - name: Bulid release binary
         run: cargo build ${{ matrix.build_args }} --release --target ${{ matrix.target }} --package tabby
 
       - name: Rename release binary
-        run: mv target/${{ matrix.target }}/release/tabby tabby_${{ matrix.binary }}
+        run: mv target/${{ matrix.target }}/release/${{ matrix.raw }} tabby_${{ matrix.binary }}
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,23 +30,20 @@ jobs:
         include:
           - os: macos-latest
             target: aarch64-apple-darwin
-            raw: tabby
             binary: aarch64-apple-darwin
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-            raw: tabby
             binary: x86_64-manylinux2014
             container: quay.io/pypa/manylinux2014_x86_64
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-            raw: tabby
             binary: x86_64-manylinux2014-cuda117
             container: sameli/manylinux2014_x86_64_cuda_11.7
             build_args: --features cuda
           - os: windows-latest
             target: x86_64-pc-windows-msvc
-            raw: tabby.exe
-            binary: x86_64-pc-windows-msvc.exe
+            binary: x86_64-pc-windows-msvc
+            ext: .exe
             build_args: --features cuda
 
     env:
@@ -105,14 +102,14 @@ jobs:
         run: cargo build ${{ matrix.build_args }} --release --target ${{ matrix.target }} --package tabby
 
       - name: Rename release binary
-        run: mv target/${{ matrix.target }}/release/${{ matrix.raw }} tabby_${{ matrix.binary }}
+        run: mv target/${{ matrix.target }}/release/tabby${{ matrix.ext }} tabby_${{ matrix.binary }}${{ matrix.ext }}
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v3
         with:
           retention-days: 3
-          name: tabby_${{ matrix.binary }}
-          path: tabby_${{ matrix.binary }}
+          name: tabby_${{ matrix.binary }}${{ matrix.ext }}
+          path: tabby_${{ matrix.binary }}${{ matrix.ext }}
 
   pre-release:
     if: github.event_name == 'push'
--- a/ci/prepare_build_environment.ps1
+++ b/ci/prepare_build_environment.ps1
@@ -1,0 +1,1 @@
+choco install --yes protoc
--- a/crates/llama-cpp-bindings/build.rs
+++ b/crates/llama-cpp-bindings/build.rs
@@ -12,6 +12,8 @@ fn main() {
 
     println!("cargo:rerun-if-changed=include/engine.h");
     println!("cargo:rerun-if-changed=src/engine.cc");
+    println!("cargo:rustc-link-lib=llama");
+    println!("cargo:rustc-link-lib=ggml_static");
 
     build_llama_cpp();
     build_cxx_binding();
@@ -29,9 +31,16 @@ fn build_llama_cpp() {
     if cfg!(feature = "cuda") {
         config.define("LLAMA_CUBLAS", "ON");
         config.define("CMAKE_POSITION_INDEPENDENT_CODE", "ON");
-        println!("cargo:rustc-link-search=native=/usr/local/cuda/lib64");
+        if cfg!(target_os = "windows") {
+            let Ok(cuda_path) = env::var("CUDA_PATH") else {
+                panic!("CUDA_PATH is not set");
+            };
+            println!(r"cargo:rustc-link-search=native={}\lib\x64", cuda_path);
+        } else {
+            println!("cargo:rustc-link-search=native=/usr/local/cuda/lib64");
+            println!("cargo:rustc-link-lib=culibos");
+        }
         println!("cargo:rustc-link-lib=cudart");
-        println!("cargo:rustc-link-lib=culibos");
         println!("cargo:rustc-link-lib=cublas");
         println!("cargo:rustc-link-lib=cublasLt");
     }
@@ -82,7 +91,7 @@ fn build_llama_cpp() {
     let dst = config.build();
     if cfg!(target_os = "windows") {
         println!(
-            "cargo:rustc-link-search=native={}\\build\\{}",
+            r"cargo:rustc-link-search=native={}\build\{}",
             dst.display(),
             config.get_profile()
         );
@@ -92,9 +101,6 @@ fn build_llama_cpp() {
 }
 
 fn build_cxx_binding() {
-    println!("cargo:rustc-link-lib=llama");
-    println!("cargo:rustc-link-lib=ggml_static");
-
     cxx_build::bridge("src/lib.rs")
         .file("src/engine.cc")
         .flag_if_supported("-Iinclude")
--- a/crates/llama-cpp-bindings/build.rs
+++ b/crates/llama-cpp-bindings/build.rs
@@ -75,8 +75,9 @@ fn build_llama_cpp() {
     // By default, this value is automatically inferred from Rust’s compilation profile.
     // For Windows platform, we always build llama.cpp in release mode.
     // See https://github.com/TabbyML/tabby/pull/948 for more details.
-    #[cfg(target_os = "windows")]
-    config.profile("Release");
+    if cfg!(target_os = "windows") {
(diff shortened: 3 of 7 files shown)


crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -9,6 +7,7 @@
 #include <ggml.h>
 #include <llama.h>
 
+#include "llama-cpp-bindings/include/engine.h"
```
darknight, line 10: Based on cxx.rs doc:
`In CXX's integration with Cargo, all #include paths begin with a crate name by default`

Without this change, `cargo build` failed on Windows (but not sure why it could pass on MacOS & Linux)

crates/tabby-common/src/path.rs
```diff
@@ -6,7 +6,7 @@ lazy_static! {
     static ref TABBY_ROOT: Mutex<Cell<PathBuf>> = {
         Mutex::new(Cell::new(match env::var("TABBY_ROOT") {
             Ok(x) => PathBuf::from(x),
-            Err(_) => PathBuf::from(env::var("HOME").unwrap()).join(".tabby"),
+            Err(_) => home::home_dir().unwrap().join(".tabby"),
```
darknight, line 9: On Windows platform, there is no default `HOME` env, by right it should be `USERPROFILE` or `HOMEDRIVE+HOMEPATH`.

But let's borrow help from a lib called `home` which is a small used by `Cargo` & `rustup`, so should be working for most of platforms.

crates/llama-cpp-bindings/build.rs
```diff
@@ -10,8 +10,8 @@ fn main() {
         "Please init submodules with `git submodule update --init --recursive` and try again"
     );
 
-    println!("cargo:rerun-if-changed=cc/*.h");
-    println!("cargo:rerun-if-changed=cc/*.cc");
+    println!("cargo:rerun-if-changed=include/engine.h");
+    println!("cargo:rerun-if-changed=src/engine.cc");
```
darknight, line 14: There's no `cc` directory, so the config might not take effect.

For this change, I just followed [the demo](https://github.com/dtolnay/cxx/blob/master/demo/build.rs) from cxx:
```
fn main() {
    cxx_build::bridge("src/main.rs")
        .file("src/blobstore.cc")
        .flag_if_supported("-std=c++14")
        .compile("cxxbridge-demo");

    println!("cargo:rerun-if-changed=src/main.rs");
    println!("cargo:rerun-if-changed=src/blobstore.cc");
    println!("cargo:rerun-if-changed=include/blobstore.h");
}
```

crates/llama-cpp-bindings/build.rs
```diff
@@ -69,7 +69,14 @@ fn main() {
     }
 
     let dst = config.build();
-    println!("cargo:rustc-link-search=native={}/build", dst.display());
+    if cfg!(target_os = "windows") {
+        #[cfg(debug_assertions)]
```
darknight, line 73: `debug_assertions` is turned on when `debug-assertions` settiongs from cargo is true.
The default profile for `cargo build` is:
```
[profile.dev]
......
debug-assertions = true
```

So this help differentiate between `debug` mode or `release` mode.

crates/llama-cpp-bindings/build.rs
```diff
@@ -69,7 +69,14 @@ fn main() {
...
+        println!("cargo:rustc-link-search=native={}\\build\\Debug", dst.display());
+        #[cfg(not(debug_assertions))]
+        println!("cargo:rustc-link-search=native={}\\build\\Release", dst.display());
```
darknight: Turns out, the folder for `llama.lib` & `ggml_static.lib` are different depending which build mode is using on windows platform, compared to linux platform, the libs are always generated under `tabby/target/[debug or release]/build/llama-cpp-bindings-7425e10c6b84e425/out/build`.

crates/tabby-download/Cargo.toml
```diff
@@ -9,4 +9,4 @@ anyhow = { workspace = true }
 tracing = { workspace = true }
 tokio-retry = "0.3.0"
 sha256 = "1.4.0"
-aim = "1.8.5"
+#aim = "1.8.5"
```
darknight, line 7: Failed to compile on windows platform.

After tracing, I found dependence chain:

```
aim -> skim-navi -> skim -> tuikit -> nix
```

This `nix` crate is specifically for *nix platform

darknight:
My test env:
```
OS:
Windows 11

rustup default:
1.73.0-x86_64-pc-windows-msvc (default)

C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\bin\Hostx64\x64> .\cl.exe
Microsoft (R) C/C++ Optimizing Compiler Version 19.35.32217.1 for x64
```

**Local test:**
Both `cargo build` and `cargo build --release` were done successfully, but the generated `tabby.exe` by `cargo build` seemed corrupted, it can not be run at all. The release build is working as expected.

I was suspecting probably because of unclean local development environment, so I setup a window runner on GitHub, see `test-rust-windows.yml` for more info.

**CI test:**
Can take a look at latest build: https://github.com/darknight/tabby/actions/runs/7109544886/job/19354636544

It's same result as my local testing, `tabby.exe` in release mode is able to execute and print `********enter main then exit********`, while debug more produce an error:
```
error: process didn't exit successfully: `target\debug\tabby.exe` (exit code: 0xc000001d, STATUS_ILLEGAL_INSTRUCTION)
```

With some expriments and googling, I didn't have much luck, so raise a PR for public discussion. But I do suspect this corruption is due to incorrect build on `llama-cpp-bindings` (either `llama.cpp` or the binding itself)

*WSL test:**

For the change, I also tested in `Ubuntu 22.04` via wsl, it's working fine. This is expected and proved those changes have no impact on already supported platforms (I don't have Mac but I think should be working as well).

cc: @wsxiaoys

wsxiaoys:
> For the change, I also tested in `Ubuntu 22.04` via wsl, it's working fine. This is expected and proved those changes have no impact on already supported platforms (I don't have Mac but I think should be working as well).

I haven’t go through all details - but one thing might worth to try is you could build chat example on windows with llama.cpp instructions. As long as it works, we should have a way to make it behave properly in tabby as well

darknight:
> First of all: I'm really looking forward to the possibility of using tabby under Windows. This would be fantastic and a great help to me. And i if have to admit that i have no insight to the code here - and do not know Rust at all. So please forgive me if my suggestion is nonsense. But from my understanding your main problem is to get llama.cpp build on different OS ?

Your understanding is correct. The main problem I faced currently is if I build `llama.cpp` in `Release` mode, everything is fine, but if build in `Debug` mode, the generated llama libraries seems not working. Currently I'm investigating the root cause.

> If my assumption is correct: you may want to consider this repo from mozilla: https://github.com/Mozilla-Ocho/llamafile - for future builds?

I took a quick look at this repo, seems its target is to generate `one single-file executable` to run everywhere. While, what we need here are the libraries built from `llama.cpp` (more specificially, `llama.lib` & `ggml_static.lib`). Those libs will be used in tabby.

> If not: sorry for not keeping my mouth shut. ;-)

Don't say that man, any idea/discussion/suggestion is welcome. @ichDaheim

darknight:
> > Your understanding is correct. The main problem I faced currently is if I build `llama.cpp` in `Release` mode, everything is fine, but if build in `Debug` mode, the generated llama libraries seems not working. Currently I'm investigating the root cause.
> 
> One workaround I would like to take is to build `llama.cpp` in release mode regardless of rust build, I think this is also the current behavior in linux / mac.

If this is acceptable, I think the PR is ready for review.

Essentially, I just added conditional compilation settings for windows, so the changes are supposed to be completely back-compatible on *nx platforms.

Both build:

```
cargo build --package llama-cpp-bindings
-OR-
cargo build --package llama-cpp-bindings --release
```

succeed on my local laptop.

**Be aware** that this is the first step to enable tabby to build on windows.

`cargo build` still failed due to `tabby` crate depends on `tabby-download` crate, which has platform specific dependence.

I plan to raise **2** more PRs separately:
1. find `aim` alternative to replace it in `tabby-download`
2. update `build.rs` again to enable feature `cuda` & `rocm`, currently both features assume lib path in *nix style, need to change accordingly.

wsxiaoys:
aim itself is opensource and it looks like its implementation is straight forward to port out from it

crates/llama-cpp-bindings/build.rs
```diff
@@ -69,8 +72,27 @@ fn main() {
         println!("cargo:rustc-link-lib=hipblas");
     }
 
+    // By default, this value is automatically inferred from Rust’s compilation profile.
+    // For Windows platform, we always build llama.cpp in release mode.
+    // See https://github.com/TabbyML/tabby/pull/948 for more details.
+    #[cfg(target_os = "windows")]
+    config.profile("Release");
```
wsxiaoys, line 79: this can be move inside of block 82 directly? without `#[cfg(...)]`

wsxiaoys:
Please add windows build to release.yml: https://github.com/TabbyML/tabby/blob/main/.github/workflows/release.yml

crates/llama-cpp-bindings/build.rs
darknight: Not exactly, `config.build()` in the next line will start the process of compiling. So we need to make the config ready before the build starts.

Alternatively, we could have following code:
```
    if cfg!(target_os = "windows") {
        config.profile("Release");
        let dst = config.build();
        println!(
            "cargo:rustc-link-search=native={}\\build\\{}",
            dst.display(),
            config.get_profile()
        );
    } else {
        let dst = config.build();
        println!("cargo:rustc-link-search=native={}/build", dst.display());
    }
```
Is this ok?

darknight:
> Please add windows build to release.yml: https://github.com/TabbyML/tabby/blob/main/.github/workflows/release.yml (with cuda on)

`cuda` settings need to be updated to reflect cuda lib path on windows, and I'm working on that.

Is it ok that I raise a separate PR to include cuda change + workflow update?

crates/llama-cpp-bindings/build.rs
wsxiaoys, line 79:
```suggestion
    if cfg!(target_os = "windows") {
        config.profile("Release");
    }
```

wsxiaoys:
please rebase against main

crates/llama-cpp-bindings/build.rs
```diff
@@ -26,7 +29,14 @@ fn main() {
...
+        if cfg!(target_os = "windows") {
+            let Ok(cuda_path) = env::var("CUDA_PATH") else {
+                panic!("CUDA_PATH is not set");
+            };
+            println!(r"cargo:rustc-link-search=native={}\lib\x64", cuda_path);
+        } else {
+            println!("cargo:rustc-link-search=native=/usr/local/cuda/lib64");
+        }
         println!("cargo:rustc-link-lib=cudart");
         println!("cargo:rustc-link-lib=culibos");
```
darknight, line 41: I did some search work, and compared the cuda doc carefully and found:

Until `12.0.0`, there indeed exists `culibos` regardless of os type, it is a thread abstraction layer, as mentioned here: https://docs.nvidia.com/cuda/archive/12.0.0/cusparse/index.html#static-library-support

Since `12.0.1`, in the similar page, there's no such keyword found: https://docs.nvidia.com/cuda/archive/12.0.1/cusparse/index.html#static-library-support

I do suspect this lib has been removed from `12.0.1` and afterwards, that's why I failed to compile on my laptop cause I installed latest cuda tookit (`12.3.1`).

So this line should continue to work on release build until we update settings.

As for local development, we can raise a separate PR to fix it for both windows & *nix platforms.

wsxiaoys: as long as the ci build pass - we're good

Whether to fix the build for cuda12 can be discussed separately

ci/prepare_build_environment.ps1
```diff
@@ -0,0 +1,3 @@
+Echo "install protocol buffer compiler..."
+choco install protoc
+choco install cuda --version=11.7.1.51694
```
darknight, line 3: One more note, I installed cuda by downloading complete `.exe` file (3.05GB), and it installed to `C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.3` by default.

Here I just use `choco` and assume it has similar behaviour, but be aware it may cause issue here.

wsxiaoys:
debugging in https://github.com/TabbyML/tabby/pull/1009

wsxiaoys:
fixed, please rebase

crates/llama-cpp-bindings/build.rs
```diff
@@ -26,9 +31,16 @@ fn main() {
...
         config.define("CMAKE_POSITION_INDEPENDENT_CODE", "ON");
-        println!("cargo:rustc-link-search=native=/usr/local/cuda/lib64");
+        if cfg!(target_os = "windows") {
+            let Ok(cuda_path) = env::var("CUDA_PATH") else {
+                panic!("CUDA_PATH is not set");
+            };
+            println!(r"cargo:rustc-link-search=native={}\lib\x64", cuda_path);
+        } else {
+            println!("cargo:rustc-link-search=native=/usr/local/cuda/lib64");
+            println!("cargo:rustc-link-lib=culibos");
```
darknight, line 41: Turns out, even cuda `11.7` doesn't contain `culibos` for windows (one build failure: https://github.com/TabbyML/tabby/actions/runs/7158762484/job/19491219353)
So move it to `else` branch.

.github/workflows/release.yml
```diff
@@ -26,20 +26,28 @@ jobs:
...
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
+            raw: tabby
             binary: x86_64-manylinux2014-cuda117
             container: sameli/manylinux2014_x86_64_cuda_11.7
             build_args: --features cuda
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+            raw: tabby.exe
+            binary: x86_64-pc-windows-msvc.exe
```
darknight, line 49: In the last commit, changed `x86_64-pc-windows-msvc` to `x86_64-pc-windows-msvc.exe`.
without this change, last step `Upload artifacts` could not find binary to upload but report no error.

wsxiaoys: can you add a field named like `ext`, with linux / mac unset but on windows set to `.exe`, then you could construct string in later steps without doing if checks on target_os

.github/workflows/release.yml
```diff
@@ -77,13 +85,27 @@ jobs:
...
+          method: 'network'
+          sub-packages: '["nvcc", "cudart", "cublas", "cublas_dev", "thrust", "visual_studio_integration"]'
+        if: runner.os == 'Windows'
 
       - name: Bulid release binary
         run: cargo build ${{ matrix.build_args }} --release --target ${{ matrix.target }} --package tabby
 
       - name: Rename release binary
-        run: mv target/${{ matrix.target }}/release/tabby tabby_${{ matrix.binary }}
+        run: mv target/${{ matrix.target }}/release/${{ matrix.raw }} tabby_${{ matrix.binary }}
```
wsxiaoys, line 108: `.../tabby${ext}`

wsxiaoys:
Otherwise LGTM!

.github/workflows/release.yml
darknight: nice idea 👍

wsxiaoys:
Thanks for the effort!

ichDaheim:
Hi @darknight @wsxiaoys 
the windows version does not start on my machine. under powershell cmd the command ".\tabby_x86_64-pc-windows-msvc.exe serve --model TabbyML/CodeLlama-7B --chat-model TabbyML/Mistral-7B" does not do anything at all. 
Double clicking the .exe gives me the error message that "cublas64_11.dll" and "cudart64.dll" are not found. (do not have a nvidia gpu). Guess this means the current version is only for users with a nvidia graphics card ?

wsxiaoys:
Yes - the distributed binary requires cuda runtime (thus nvidia gpu card). If you're interested in CPU only binary distribution, feel free to file an issue, thanks!

PR #961, opened 2023-12-06
Add ssh-client to Dockerfile

chpeer:
This PR adds the ssh-client as well as the ca-certificates to the Dockerfile and updates the documentation accordingly to reflect the change.

With this change, the scheduler can now clone repositories via SSH rather than just HTTP.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,9 +20,7 @@ RUN apt-get update && \
         protobuf-compiler \
         git \
         cmake \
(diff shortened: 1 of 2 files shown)


chpeer:
I've not managed to build the Dockerfile due to some environment setup issues. I've tested it on a custom Dockerfile which was derived from the tabby Dockerfile on DockerHub.

PR #989, opened 2023-12-08
feat(tabby-ui): properly set accessToken for api access if user is login

wsxiaoys:


ee/tabby-ui/components/prompt-form.tsx
```diff
@@ -45,7 +46,6 @@ function PromptFormRenderer(
   const [queryCompletionUrl, setQueryCompletionUrl] = React.useState<
     string | null
   >(null)
-  const latestFetchKey = React.useRef('')
```
wsxiaoys, line 48: this fetchKey check no longer work once we switched to array key, is it ok to remove it?

@liangfung

liangfung: yes, it is ok to remove it and we should call `setOptions` according to the return `data` from `useSwr` hook with `React.useEffect`. `SWR` handles the case of request sequencing.

PR #996, opened 2023-12-09
feat: add aim-downloader

darknight:
Follow up of https://github.com/TabbyML/tabby/pull/948

As @wsxiaoys suggested in the [comment](https://github.com/TabbyML/tabby/pull/948#issuecomment-1845588932), we can maintain a cuszomized verson of `aim` to keep using it.

Luckily, `aim` is MIT license and codebase is not big, so I did following changes:

- created `aim-downloader`, copy `aim` source code to this crate
- removed `main.rs` & `driver.rs`, removed `skim-navi` dependence
  - `main.rs` provides command-line interface, which is not needed as a lib
  - `driver.rs` is only used by `main.rs`, so removed
    - `http_serve_folder.rs` is only used in `driver.rs`, which enables `aim` serve a folder over http. It's useless when aim as a lib, so removed. Also removed its dependence: ~~warpy = "0.3.35"~~
  - `skim-navi` is only used by `driver.rs` and incompatible with windows, so removed: ~~skim-navi = "0.1.10"~~
- based on aim's [feature matrix](https://github.com/mihaigalos/aim?tab=readme-ov-file#feature-matrix), `s3` & `ssh` download does not support **resume**, which is why we use this lib. So they're useless, removed them and related test data
- re-implement `untildify`, another incompatible lib with windows. `untildify` is a one-function crate which convert a tilde path to its full path(e.g. `~/.ssh` to `/home/user/.ssh`). It's also MIT license, so I copied unit tests from its codebase, instead of copying it's implementation, I just re-wrote the function which defined in `untildify.rs`.

With these changes, together with the changes from https://github.com/TabbyML/tabby/pull/948 applied in tabby, I'm able to build `tabby` successfully regardless of which build mode.

And I'm able to run `download` without issue:
**Release** mode:
Screenshot 2023-12-09 200759
**Debug** mode:
Screenshot 2023-12-09 202907

Pending work:
- Shall we keep narrowing down `aim` to support http download only? (so we don't check-in unused code)


wsxiaoys:
Since we only utilize `HttpsHandler`, shall we remove all other handles?

crates/aim-downloader/src/untildify.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn test_returns_empty_string() {
+        env::remove_var("HOME");
+
+        let home = Path::new("/User/Untildify");
+        env::set_var("HOME", home.as_os_str());
+
+        assert_eq!(untildify("Desktop"), "Desktop");
+        assert_eq!(untildify(""), "");
+        assert_eq!(untildify("/"), "/");
+        // assert_eq!(untildify("~/Desktop/~/Code"), "/User/Untildify/Desktop/");
```
darknight, line 51: Original implementation used regex to capture the first tilde, then truncate the rest string.
Current implementation does not check if tilde exists in the middle of string, so this case will fail.

wsxiaoys:
since we don't run any integration tests for aim - we can remove the tests dir?

darknight:
> since we don't run any integration tests for aim - we can remove the tests dir?

One more thing, some of the tests require network access to test http handler, for example: `get_https_works`, shall we disable it as well?

crates/aim-downloader/src/https.rs
```diff
@@ -0,0 +1,268 @@
...
+            .basic_auth(parsed_address.username, Some(parsed_address.password))
+            .send()
+            .await
+            .map_err(|_| format!("Failed to GET already uploaded size from {}", &output))
+            .unwrap();
+        res.content_length().unwrap_or(0)
+    }
+}
+
+#[ignore]
```
darknight, line 171: test requires network access

crates/aim-downloader/src/https.rs
```diff
@@ -0,0 +1,268 @@
...
+    let expected_hash = "0e0f0d7139c8c7e3ff20cb243e94bc5993517d88e8be8d59129730607d5c631b";
+    let out_file = "tokei-x86_64-unknown-linux-gnu.tar.gz";
+
+    let result = HTTPSHandler::get("https://github.com/XAMPPRocky/tokei/releases/download/v12.0.4/tokei-x86_64-unknown-linux-gnu.tar.gz", out_file, &mut WrappedBar::new_empty(), expected_hash).await;
+
+    assert!(result.is_ok());
+    std::fs::remove_file(out_file).unwrap();
+}
+
+#[ignore]
```
darknight, line 183: test requires network access and external test file

crates/aim-downloader/src/https.rs
```diff
@@ -0,0 +1,268 @@
...
+    .unwrap();
+
+    let _ = HTTPSHandler::get("https://github.com/Byron/dua-cli/releases/download/v2.10.2/dua-v2.10.2-x86_64-unknown-linux-musl.tar.gz", out_file, &mut WrappedBar::new_empty_verbose(), "").await;
+
+    let actual_size = std::fs::metadata(out_file).unwrap().len();
+    assert_eq!(actual_size, expected_size);
+    std::fs::remove_file(out_file).unwrap();
+}
+
+#[ignore]
```
darknight, line 201: same as above

crates/aim-downloader/src/https.rs
```diff
@@ -0,0 +1,268 @@
...
+"#;
+
+    let result = HTTPSHandler::list("https://example.com").await.unwrap();
+    let result = str::replace(&result, "    \n    ", "");
+    let result = str::replace(&result, "</style>    \n</head>", "</style>\n</head>");
+
+    assert_eq!(result, expected);
+}
+
+#[ignore]
```
darknight, line 258: same as above, add `ignore` to skip.

crates/aim-downloader/src/netrc.rs
```diff
@@ -0,0 +1,64 @@
...
+    file.write_all(b"machine mydomain.com login myuser password mypass port 1234")
+        .unwrap();
+
+    assert!(netrc(true).is_some());
+    assert!(netrc(false).is_some());
+
+    std::fs::remove_file(".netrc.test_unit").unwrap();
+}
+
+#[ignore]
```
darknight, line 52: test failed then running on windows in the last step `remove_file` call - `cannot find the file specified`.

run individually is fine. I suspect race condition with above test considering `cargo test` run in parallel.

So move test logic from this test case to previous one, and disable this.

Issue #877, opened 2023-11-23
Extract dependencies for javascript `package-lock.json` / `package.json`


wsxiaoys:

wsxiaoys:
it's more like what we've done in https://github.com/TabbyML/tabby/blob/main/crates/tabby-scheduler/src/dataset/deps/rust.rs

Basically you'll need to create a function like `process_package_lock_json` and returns `Vec<Package>`

PR #1004, opened 2023-12-10
feat: Add a javascript.rs file to solve issue #877

Someet-Git:
- In this PR I have created a javascript.rs file which is to extract dependencies.
- some changes in the mod.rs
Basically to solve #877 .

Hey @wsxiaoys can please check this and suggest any changes.


crates/tabby-scheduler/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+tokio = { workspace = true, features = ["full"] }
```
wsxiaoys, line 9: is this necessary?

crates/tabby-scheduler/src/dataset/deps/javascript.rs
```diff
@@ -0,0 +1,39 @@
...
+use tabby_common::Package;
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+struct NpmPackage {
+    name: String,
+    version: String,
+}
+
+fn extract_npm_deps<'a, I>(packages: I) -> Vec<Package>
```
wsxiaoys, line 12: this generates all transitive dependencies, not just `direct` dependencies.

That's why I suggest utilizing https://crates.io/crates/package-lock-json-parser/0.1.1 (which provides a util to extract direct dependencies)

wsxiaoys (changes requested):

crates/tabby-scheduler/Cargo.toml
Someet-Git: was giving error while running test

crates/tabby-scheduler/src/dataset/deps/javascript.rs
Someet-Git: ok let me try this

Someet-Git:
@wsxiaoys can u review the latest commit?

wsxiaoys:
Hi! Thanks for the updates! Have you verified this will only generates direct dependencies, while skip the transitive ones?

One example is let’s say the repository depends on lib A while lib A depends on lib B, we would really prefer only lib A is extracted for the purpose.

Someet-Git:
Hey @wsxiaoys i have done some changes in the javascript.rs file to import the direct deps, and also verified that.

Someet-Git:
Hey @wsxiaoys 
i'm a bit confused here, for example in the  package-lock.json file the for `client/tabby-agent`
there are `dependencies` and `devdependencies`, so we need to extract the name and version of the same?

I checked that it is extracting dependencies of node_module in the package-lock.json file. Is it require?

wsxiaoys:
Hey - thanks for the ping, will give it a try once I got some bandwidth, mostly I wanna make sure this extract only immediate dependencies (which is not obvious by looking from PR)

PR #1012, opened 2023-12-10
feat: Add rocm builds and documentation

### LINKED ISSUE #636: AMD GPUs (ROCm) support

cromefire:
Fixes #636

Proper display in Telemetry and UI not included, see #902 for that.


.github/workflows/release.yml
```diff
@@ -44,7 +44,7 @@ jobs:
             target: x86_64-unknown-linux-gnu
             binary: x86_64-manylinux2014-rocm5.7
             container: rocm/dev-ubuntu-22.04:5.7-complete
-            apt_dependencies: git pkg-config libssl-dev cmake protobuf-compiler
```
wsxiaoys: this is usually handled in bash ./ci/prepare_build_environment.sh

.github/workflows/release.yml
```diff
@@ -44,7 +44,7 @@ jobs:
             target: x86_64-unknown-linux-gnu
             binary: x86_64-manylinux2014-rocm5.7
             container: rocm/dev-ubuntu-22.04:5.7-complete
```
wsxiaoys, line 46: this doesn't seems to be an actual manylinux build - could you confirm?

wsxiaoys (commented):
Please ensure that the pull request (PR) focuses on a single part. I would suggest that you only modify the Docker-related CI configuration files (.yml/.Dockerfile) in this PR to facilitate a smoother merge.

.github/workflows/release.yml
cromefire: Probably not I guess? Pretty sure there's no container for ROCm for that, I can probably just rename the binary. The resulting binary is fine, it doesn't link against anything special:
```
~$ ldd /tmp/tabby_x86_64-manylinux2014-rocm5.7
	linux-vdso.so.1 (0x00007ffcdaac1000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f5757e00000)
	libamdhip64.so.5 => /opt/rocm-5.7.0/lib/libamdhip64.so.5 (0x00007f5756200000)
	librocblas.so.3 => /opt/rocm-5.7.0/lib/librocblas.so.3 (0x00007f5736800000)
	libhipblas.so.1 => /opt/rocm-5.7.0/lib/libhipblas.so.1 (0x00007f575bee1000)
	libssl.so.3 => /lib/x86_64-linux-gnu/libssl.so.3 (0x00007f575be3d000)
	libcrypto.so.3 => /lib/x86_64-linux-gnu/libcrypto.so.3 (0x00007f5736200000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f575be1b000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f5758119000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5735e00000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f575bf78000)
	libamd_comgr.so.2 => /opt/rocm-5.7.0/lib/libamd_comgr.so.2 (0x00007f572d400000)
	libhsa-runtime64.so.1 => /opt/rocm-5.7.0/lib/libhsa-runtime64.so.1 (0x00007f572d000000)
	libnuma.so.1 => /lib/x86_64-linux-gnu/libnuma.so.1 (0x00007f575be0c000)
	librocsolver.so.0 => /opt/rocm-5.7.0/lib/librocsolver.so.0 (0x00007f56cf200000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f575bdf0000)
	libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007f575bdbe000)
	libelf.so.1 => /lib/x86_64-linux-gnu/libelf.so.1 (0x00007f575bda0000)
	libdrm.so.2 => /opt/amdgpu/lib/x86_64-linux-gnu/libdrm.so.2 (0x00007f57580ff000)
	libdrm_amdgpu.so.1 => /opt/amdgpu/lib/x86_64-linux-gnu/libdrm_amdgpu.so.1 (0x00007f57580f1000)
	librocsparse.so.0 => /opt/rocm-5.7.0/lib/librocsparse.so.0 (0x00007f568d400000)
```

.github/workflows/release.yml
cromefire: Well I don't have access to the type of the image from there, so not sure how I could do it from there.

cromefire: Ah and the script also runs too late, because it runs after the checkout, but then it'll fail because the git is not installed

cromefire:
> I would suggest that you only modify the Docker-related CI configuration files (.yml/.Dockerfile) in this PR to facilitate a smoother merge.

Well I mean documentation seems to be a pretty important part of it and I wouldn't think there are many issues there. Sure I can split it again, but I'm sinking many hours into testing and splitting this, there's a limit somewhere to what I can manage as well. This already consumed most of my Sunday now.

.github/workflows/release.yml
wsxiaoys: In that case - I would suggest just rely on docker build for rocm atm, non manylinux build won't be very useful (as it's specific to certain distributions)

wsxiaoys:
> > I would suggest that you only modify the Docker-related CI configuration files (.yml/.Dockerfile) in this PR to facilitate a smoother merge.
> 
> Well I mean documentation seems to be a pretty important part of it and I wouldn't think there are many issues there. Sure I can split it again, but I'm sinking many hours into testing and splitting this, there's a limit somewhere to what I can manage as well. This already consumed most of my Sunday now.

I fully understand and would like to express my gratitude once again for your contribution :)

I mainly refer to certain changes (e.g., Rust 1.73 -> 1.8, manylinux build). These are all valuable, but they require careful testing to ensure that everything still works.

The documentation part is fine to retain; I will provide comments once we reach a good point in the code section.

cromefire:
> I mainly refer to certain changes (e.g., Rust 1.73 -> 1.8, manylinux build). These are all valuable, but they require careful testing to ensure that everything still works.

Yeah sure I can just downgrade that again. Thought as docker uses latest stable anyway, might as well update that.

.github/workflows/release.yml
cromefire: It doesn't rely on anything distro specific, so it should be fine. Almost no one uses manylinux outside of python and there are still rarely problems. Seems like you have a strong python background.

Trying some rando manylinux image from pytorch though, maybe it'll work (if GH decides to actually run that build...)

cromefire:
Well the manylinux image for rocm now seems to run into a disk space limit, could you maybe try and delete some caches or so? In theory it's much smaller (< 6GB) than the cuda (~10GB) image, which is why I'm slightly confused about it... Maybe it's running an parallel on the same VM?

.github/workflows/release.yml
cromefire: Should probably to be fixed with the manylinux image anyway, as soon as it builds.

wsxiaoys:
> Well the manylinux image for rocm now seems to run into a disk space limit, could you maybe try and delete some caches or so? In theory it's much smaller (< 6GB) than the cuda (~10GB) image, which is why I'm slightly confused about it... Maybe it's running an parallel on the same VM?

I'll suggest don't generate manylinux build atm, and checkin docker image in this PR.

cromefire:
I think it's just a fluke in the build and the build has to be restarted (as it worked before with a bigger container). Can't do that myself though.

cromefire:
Found the error: The size on docker hub is compressed and apparently ROCm is quite big, but compresses super well plus the manylinux and pytorch also added magma and MLOpen to it...

Got everything working by installing only the minimum needed on the fly. The biggest problem is that this adds ~4min to the ROCm build and I think the ROCm repo might also be slower than today sometimes...

As a solution I have prepared this: https://github.com/cromefire/hipblas-manylinux. Super basic docker build setup, but if you want me to switch to that, I'd recommend we move the repo to this org, so it's easier to maintain (basically to only thing required would be to upgrade to the latest ROCm version from time to time) as I probably won't always remember to upgrade it to the latest ROCm version.

.github/workflows/release.yml
cromefire: Kinda solved

.github/workflows/release.yml
cromefire: Using hipblas on top of manylinux, this should be fixed now

cromefire:
Also Windows build is possible as well (HIP install for windows can be found [here](https://rocm.docs.amd.com/en/latest/deploy/windows/cli/install.html)), but I'll probably just open an issue for that, because I don't think I have the bandwidth right now to deal with windows, maybe one day...

website/docs/faq.mdx
```diff
@@ -18,13 +19,26 @@ import CodeBlock from '@theme/CodeBlock';
     <p>
       To determine the mapping between the GPU card type and its compute capability, please visit <a href="https://developer.nvidia.com/cuda-gpus">this page</a>
     </p>
+    <p>
+      This also seems to be available on AMD Radeon™ GPUs, but it's unclear which cards besides RDNA3 support this.
+    </p>
```
wsxiaoys, line 24:
```suggestion
```

website/docs/installation/docker.mdx
```diff
@@ -6,14 +6,24 @@ sidebar_position: 0
 
 This guide explains how to launch Tabby using docker.
 
+
+
+
```
wsxiaoys, line 11:
```suggestion
```

website/docs/installation/docker-compose.mdx
```diff
@@ -5,6 +5,8 @@ sidebar_position: 1
 # Docker Compose
 This guide explains how to launch Tabby using docker-compose.
 
+
+
```
wsxiaoys, line 9:
```suggestion
```

website/docs/faq.mdx
```diff
@@ -24,7 +25,17 @@ import CodeBlock from '@theme/CodeBlock';
 <details>
   <summary>How to utilize multiple NVIDIA GPUs?</summary>
   <div>
-    <p>Tabby only supports the use of a single GPU. To utilize multiple GPUs, you can initiate multiple Tabby instances and set CUDA_VISIBLE_DEVICES accordingly.</p>
+    <p>Tabby only supports the use of a single GPU. To utilize multiple GPUs, you can initiate multiple Tabby instances and set CUDA_VISIBLE_DEVICES or HIP_VISIBLE_DEVICES accordingly.</p>
```
wsxiaoys, line 28:
```suggestion
    <p>Tabby only supports the use of a single GPU. To utilize multiple GPUs, you can initiate multiple Tabby instances and set CUDA_VISIBLE_DEVICES (for cuda) or HIP_VISIBLE_DEVICES (for rocm) accordingly.</p>
```

wsxiaoys:
Hey - I reverted most docker related part - given the manylinux build now works and I’m thinking of re-organize the docker image build to be around these manylinux binaries directly - will do it as a followup from my side.

The other part LGTM, thanks for your effort!

cromefire:
Yeah as said though, we probably want to transfer the image repo for ROCm to this org, so you don't have to wait on me for any upgrades and it's all under your control.

cromefire:
And also please at least restore the Dockerfiles themselves, as they are an easy way for people to build their own optimized images quickly.

cromefire:
Also because everything, including compilers, is old and crusty you'd probably actually want to have an optimized build in docker which can use all the newer tools there, including the latest compiler optimizations and mitigations, after all, you don't have to worry about compatibility.

cromefire:
It would also enable you to adjust the build, so it builds the ROCm libraries from source (they are open source after all), so you could build everything just for your GPU, which would probably shrink the container from like 10GB+ to more like 1-2GB.

wsxiaoys:
> For ROCm this is important as you may have seen these AMDGPU_TARGETS, and while it includes most GPUs, it doesn't include all, so having a way for people to just build it quickly for their specific GPU would be really important, that you I always built and tested all of this as well.

Since the user need to build and test it manually - I'll prefer just leave it in source code.

> Also because everything, including compilers, is old and crusty you'd probably actually want to have an optimized build in docker which can use all the newer tools there, including the latest compiler optimizations and mitigations, after all, you don't have to worry about compatibility.

Let me think about this a bit more - but I'm still a bit leaning to just copy manylinux binary to containers to make build process simpler

cromefire:
> Since the user need to build and test it manually - I'll prefer just leave it in source code.

Yeah, not saying you have to base your official docker builds on it, but it'd be good to at least have the Dockerfile there for people to use to build specific builds or just for development. As said I basically never compiled it on bare metal as I initially had some build issues and in docker it just worked. (And you'd also have to install ROCm on your machine directly)

cromefire:
Something like that.

PR #1017, opened 2023-12-10
chore: update rusqlite, rusqlite_migration and tokio-rusqlite

cljoly:
I’ve released a new beta version of rusqlite_migration today and I’ve
renamed the `async-tokio-rusqlite` feature that you use to
`alpha-async-tokio-rusqlite`. I suggest updating dependencies as done in
this PR. By the way, if you would like to share feedback on the async
APIs, feel free to use this conversation:
https://github.com/cljoly/rusqlite_migration/discussions/127

Reading the release notes from [tokio-rusqlite][1], the main breaking
change is to the method `Connection::call`:

> - **breaking:** `Connection::call` now takes `tokio_rusqlite::Result` instead
>   of `rusqlite::Result`.

My understanding is that we can either deal with a `Result<Result<…>>`
or call `Connection::call_unwrap`, which could panic if the connection
is closed. I’ve opted for using only `Connection::call` and,
effectively, flattening the `Result<Result<…>>`, to limit the potential
for panics.

[1]: https://github.com/programatik29/tokio-rusqlite/blob/master/CHANGELOG.md#050-25-nov-2023

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1012,9 +1012,9 @@ dependencies = [
 
 [[package]]
 name = "fallible-iterator"
-version = "0.2.0"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4443176a9f2c162692bd3d352d745ef9413eec5782a80d8fd6f8a1ac692a07f7"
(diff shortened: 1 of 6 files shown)


wsxiaoys:
Thanks for the PR, LGTM

PR #1021, opened 2023-12-11
feat(core): add --host option to specify the host Tabby server will listen on.

jasonharrison:
* Add --host option to specify the host Tabby server will listen on (default remains 0.0.0.0 if option is not specified)
* Correct small typo in startup message: "Starting server, this might ~takes~ **take** a few minutes..."
Screenshot 2023-12-11 at 00 38 28


wsxiaoys:
Thanks for the PR, LGTM

PR #1030, opened 2023-12-13
feat: Add repository context support for csharp

Blackclaws:
Adds repository context for csharp. I tried following existing conventions for how this is accomplished.


wsxiaoys:
done in https://github.com/TabbyML/tabby/pull/1355

PR #1044, opened 2023-12-14
fix: create event file by event timestamp, refactoring code

darknight:
For closing https://github.com/TabbyML/tabby/issues/1036

Changes include:

- remove `std::io`, apply `tokio::io` completely to prevent blocking
- create event log file based on the date when receive the event
- add tests

### DIFF
--- a/crates/tabby/src/services/event.rs
+++ b/crates/tabby/src/services/event.rs
@@ -115,20 +115,6 @@ mod tests {
         std::env::temp_dir().join(".tabby").join("events")
     }
 
-    async fn test_event_writer_normal_write() {
-        let mut event_wr = EventWriter::new(events_dir()).await;
-        for i in 0..100 {
-            event_wr.write_line(format!("line {}", i)).await;
-        }
-        event_wr.flush().await;
-
-        // we should be able to read target file successfully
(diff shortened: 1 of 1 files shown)


crates/tabby/src/services/event.rs
```diff
@@ -64,3 +99,83 @@ impl RawEventLogger for EventService {
...
+    }
+
+    async fn test_event_writer_normal_write() {
+        let mut event_wr = EventWriter::new(events_dir()).await;
+        for i in 0..100 {
+            event_wr.write_line(format!("line {}", i)).await;
+        }
+        event_wr.flush().await;
+
+        let now = Utc::now();
```
wsxiaoys, line 118: there's a rare chance this become a flacky test (e.g run at utc 23:59:59).

Is there a way to avoid it?

darknight: that's true...nice catch!

add a helper method to fix it.

wsxiaoys: line 129 could still break, right?

The only way to ensure it pass is to list all files writer has modified and count it in total has 100 lines?

I don't think it's actually something really useful. I'm fine to remove this test case and keep the swapping one

PR #1062, opened 2023-12-16
fix: pass ee root when call create_dir_all

darknight:
Follow up of https://github.com/TabbyML/tabby/pull/1057

In that PR, we pass the db file path for `create_dir_all` which will cause the exception when run the webserver.

This PR fix it.


ee/tabby-webserver/src/service/db/mod.rs
```diff
@@ -33,9 +36,8 @@ impl DbConn {
     }
 
     pub async fn new() -> Result<Self> {
-        let db_path = db_file();
-        tokio::fs::create_dir_all(db_path.as_path()).await?;
```
wsxiaoys, line 37: lets run db path‘s parent dir instead of passing ee root?

Issue #1037, opened 2023-12-14
Persist `tabby scheduler` run output in `~/.tabby/runs`


wsxiaoys:
Each run should be an dir dated with UTC timestamp, with following information:

```
meta.json
stdout
stderr
```

schema of meta.json (should be add to tabby-common)
```rust
struct SchedulerRunMeta {
  start_time: Datetime<Utc>,
  finish_time: Datetime<Utc>,
  success: boolean
};
```

PR #1064, opened 2023-12-17
feat: persist tabby scheduler run output into db

darknight:
For closing https://github.com/TabbyML/tabby/issues/1037

The logic is done with code refining work left, I want to raise first to collect feedback, make sure I'm on the correct direction.

What I've changed:

For **tabby-common**
- defined `SchedulerRunMeta`

For **tabby-scheduler**
- extract logic from `job1` & `job2` into separate functions, make them reused in new method `scheduler_job`
- define `ee` feature, `scheduler_job` is only present when `ee` feature is enabled

For **tabby**
- update `ee` feature which depends on both `tabby-webserver` & `ee` from `tabby-scheduler`
- add new cli subcommand with following format:
```
tabby.exe scheduler::job --run sync-repository
tabby.exe scheduler::job --run index-repository
```
- new subcommand is hided by default and only compiled when `ee` feature is turned on

For **tabby-webserver**
- originally there is a db job to delete expired refresh token, nothing changed for it, just split it into `job` mod.
- define `repository_job!` macro, this is where we run the new subcommand & implement persistence.
  - destination dir would be `~/.tabby/runs/<job name>/<run_datetime>/`, for example:
    - `runs/sync-repository/run_2023-12-17-08-38-35/meta.json`
    - `runs/index-repository/run_2023-12-17-08-38-35/stderr`
- the repository related jobs are run in a child process, which blocks until finish, so manage them in a different job scheduler (temporary solution, see below)

A few observations & concerns:
1. based on my tests, repository jobs are resource consuming, so perhaps do not run them too frequently?
2. we're using `kdam::Bar` to indicate progress during job running in the console, and they are captured into `stderr`. But the captured output contains every update of the bar from `0%` to `100%`, and since it is a simulation of a bar, the saved `stderr` file are not readable directly.
3. in the new `run_offline_job` method, I'm still using `tokio::spawn` which is incorrect, I'm thinking using `tokio::task::spawn_blocking` and the same synchronous version of job scheduler to handle this type of jobs.


crates/tabby-common/src/schema.rs
wsxiaoys, line 1: it seems to me that the entire `SchedulerRunMeta` can be put inside of `tabby-webserver` - since it's the producer / consumer of such run meta.

crates/aim-downloader/src/io.rs
```diff
@@ -8,13 +8,7 @@ fn get_output_file(path: &str, silent: bool) -> (Option<std::fs::File>, u64) {
...
                 println!("File exists. Resuming.");
             }
-            file = Some(
-                std::fs::OpenOptions::new()
-                    .write(true)
-                    .append(true)
-                    .open(path)
-                    .unwrap(),
-            );
+            file = Some(std::fs::OpenOptions::new().append(true).open(path).unwrap());
```
wsxiaoys, line 11: any reason to remove `write(true)`?

crates/tabby/src/main.rs
```diff
@@ -40,6 +40,11 @@ pub enum Commands {
     /// Run scheduler progress for cron jobs integrating external code repositories.
     Scheduler(SchedulerArgs),
 
+    /// Run specified job in background
+    #[cfg(feature = "ee")]
+    #[clap(name = "scheduler::job", hide = true)]
+    SchedulerJob(SchedulerJobArgs),
```
wsxiaoys, line 46: define new job like workers,

```rust
    #[cfg(feature = "ee")]
    #[clap(name = "job::sync")]
    JobSync(job::sync),
```

ee/tabby-webserver/src/path.rs
```diff
@@ -9,3 +9,7 @@ fn tabby_ee_root() -> PathBuf {
 pub fn db_file() -> PathBuf {
     tabby_ee_root().join("db.sqlite")
 }
+
+pub fn cron_job_runs_dir() -> PathBuf {
```
wsxiaoys, line 13:
```suggestion
pub fn job_runs_dir() -> PathBuf {
```

ee/tabby-webserver/src/path.rs
```diff
@@ -9,3 +9,7 @@ fn tabby_ee_root() -> PathBuf {
...
+    tabby_root().join("runs")
```
wsxiaoys, line 14:
```suggestion
    tabby_ee_root().join("runs")
```

ee/tabby-webserver/src/service/cron/job.rs
```diff
@@ -0,0 +1,68 @@
...
+            if let Err(e) = res {
+                error!("failed to delete expired token: {}", e);
+            }
+        })
+    })?;
+
+    Ok(job)
+}
+
+macro_rules! repository_job {
```
wsxiaoys, line 22: seems to me this could be a function, instead of a macro?

ee/tabby-webserver/src/service/cron/job.rs
```diff
@@ -0,0 +1,68 @@
+use anyhow::Result;
+use tokio_cron_scheduler::Job;
+use tracing::error;
+
+use crate::service::db::DbConn;
+
+pub(crate) async fn refresh_token_async(db_conn: DbConn) -> Result<Job> {
```
wsxiaoys, line 7: split to individual file (e.g db.rs)

ee/tabby-webserver/src/service/cron/job.rs
```diff
@@ -0,0 +1,68 @@
...
+            std::time::Duration::from_secs($interval),
+            |_, _| {
+                use tabby_common::schema::SchedulerRunMeta;
+
+                use crate::path::cron_job_runs_dir;
+
+                let start_time = chrono::Utc::now();
+
+                let exe = std::env::current_exe().unwrap();
+                let mut cmd = std::process::Command::new(exe);
```
wsxiaoys, line 34: will subprocess's stdout / stderr being piped to current stdout / stderr (we don't want that)

wsxiaoys (changes requested):

crates/tabby-scheduler/src/job.rs
wsxiaoys, line 1: Seems to me these two function could be put inside of `dataset.rs` / `index.rs`?

crates/aim-downloader/src/io.rs
darknight: wow...actually this change was made by `autofix-ci[bot]`, I didn't change irrelevant module.

Well, as I check the std doc, [it mentioned](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.append)
```
Note that setting .write(true).append(true) has the same effect as setting only .append(true).
```

I guess that's why the bot removed it.
Perhaps configure the bot to format code ONLY instead of optimizing it?

ee/tabby-webserver/src/service/cron/job.rs
darknight: I tested locally, if use `inherit` here, like:
```
            cmd.stdout(Stdio::inherit())
                .stderr(Stdio::inherit())
                .args(["job::sync", "--job", &job_name]);
```

The job run output will display on the console, and the collected `output.stderr` & `output.stdout` are both empty.

crates/aim-downloader/src/io.rs
```diff
@@ -10,7 +10,7 @@ fn get_output_file(path: &str, silent: bool) -> (Option<std::fs::File>, u64) {
             }
             file = Some(
                 std::fs::OpenOptions::new()
-                    .write(true)
```
darknight, line 13: I reverted this change in last commit, but the autofix bot removed it one more time...

crates/tabby/src/main.rs
```diff
@@ -40,6 +42,11 @@ pub enum Commands {
     /// Run scheduler progress for cron jobs integrating external code repositories.
     Scheduler(SchedulerArgs),
 
+    /// Run specified job in background
+    #[cfg(feature = "ee")]
+    #[clap(name = "job::sync", hide = true)]
+    JobSync(job::SyncArgs),
```
wsxiaoys, line 48: job::sync and job::index each should be individual command, just like WorkerCompletion/WorkerChat

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -138,6 +139,13 @@ pub struct User {
     pub created_at: DateTime<Utc>,
 }
 
+#[derive(Serialize, Deserialize, GraphQLObject)]
```
wsxiaoys, line 142: there's no need to be a `GraphQLObject`, at least for now.

ee/tabby-webserver/src/service/cron/job.rs
```diff
@@ -0,0 +1,44 @@
+use job_scheduler::Job;
+use tracing::error;
+
+use crate::{path::job_runs_dir, schema::SchedulerRunMeta};
+
+pub(crate) fn repository_job<'a>(job_name: &'a str, schedule: &'a str) -> Job<'a> {
```
wsxiaoys, line 6:
```suggestion
pub fn repository_job<'a>(job_name: &'a str, schedule: &'a str) -> Job<'a> {
```

ee/tabby-webserver/src/service/cron/db.rs
```diff
@@ -0,0 +1,19 @@
+use tokio_cron_scheduler::Job;
+use tracing::error;
+
+use crate::service::db::DbConn;
+
+pub(crate) async fn refresh_token_async(db_conn: DbConn) -> anyhow::Result<Job> {
```
wsxiaoys, line 6:
```suggestion
pub async fn refresh_token_async(db_conn: DbConn) -> anyhow::Result<Job> {
```

ee/tabby-webserver/src/service/cron/job.rs
```diff
@@ -0,0 +1,44 @@
...
+        let Ok(output) = output else {
+            error!("job `{}` failed: {:?}", &job_name, output.unwrap_err());
+            return;
+        };
+        let finish_time = chrono::Utc::now();
+
+        // prepare directory
+        let this_run_dir = job_runs_dir()
+            .join(&job_name)
+            .join(format!("run_{}", start_time.format("%Y-%m-%d-%H-%M-%S")));
```
wsxiaoys, line 27: add microseconds to avoid conflict

ee/tabby-webserver/src/schema/mod.rs
wsxiaoys: after a second thought - since we're creating / consuming SchedulerRunMeta in webserver anyway, would it be better if we simply make this a new table (e.g job_runs, with following columns: `job`, `start_ts`, `end_ts`, `success`, `stdin`, `stderr`), what do you think?

ee/tabby-webserver/src/service/cron/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod db;
+mod job;
+mod run_async;
+mod run_sync;
+
+pub use run_async::run_offline_job_async;
+pub use run_sync::run_offline_job;
```
wsxiaoys, line 7: I do not understand why there's a need to split run_offline_job_async / run_offline_job_sync.

I would suggest just expose a single function `run_cron`

wsxiaoys (changes requested):

ee/tabby-webserver/src/service/cron/job.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::{path::job_runs_dir, schema::SchedulerRunMeta};
+
+pub(crate) fn repository_job<'a>(job_name: &'a str, schedule: &'a str) -> Job<'a> {
+    let job_name = job_name.to_owned();
+
+    Job::new(schedule.parse().unwrap(), move || {
+        let start_time = chrono::Utc::now();
+
+        let exe = std::env::current_exe().unwrap();
+        let mut cmd = std::process::Command::new(exe);
```
wsxiaoys, line 13: you could use https://docs.rs/tokio/latest/tokio/process/index.html to make it async

wsxiaoys:
> we're using kdam::Bar to indicate progress during job running in the console, and they are captured into stderr. But the captured output contains every update of the bar from 0% to 100%, and since it is a simulation of a bar, the saved stderr file are not readable directly.

 we can detect whether the terminal is interactive to decide whether to use kdam::Bar (crate: `atty`)

ee/tabby-webserver/src/schema/mod.rs
wsxiaoys: let us go with db then

darknight:
> > based on my tests, repository jobs are resource consuming, so perhaps do not run them too frequently?
> > I'll comment on this one everything else is resolved.
> 
> > we're using kdam::Bar to indicate progress during job running in the console, and they are captured into stderr. But the captured output contains every update of the bar from 0% to 100%, and since it is a simulation of a bar, the saved stderr file are not readable directly.
> 
> we can detect whether the terminal is interactive to decide whether to use kdam::Bar (crate: `atty`)

How to achieve this?

> > in the new run_offline_job method, I'm still using tokio::spawn which is incorrect, I'm thinking using tokio::task::spawn_blocking and the same synchronous version of job scheduler to handle this type of jobs.
> 
> Replied inline.

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -138,6 +139,17 @@ pub struct User {
     pub created_at: DateTime<Utc>,
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct JobRunResult {
```
wsxiaoys, line 143:
```suggestion
pub struct JobRun {
```

ee/tabby-webserver/src/service/cron/db.rs
```diff
@@ -0,0 +1,58 @@
...
+            if let Err(e) = res {
+                error!("failed to delete expired token: {}", e);
+            }
+        })
+    })?;
+
+    Ok(job)
+}
+
+pub async fn repository_job(db_conn: DbConn, job_name: String, schedule: &str) -> Result<Job> {
```
wsxiaoys, line 22: `db.rs` here means it's about jobs doing db maintenance.

move to individual file, or just `mod.rs`.

ee/tabby-webserver/src/service/cron/db.rs
```diff
@@ -0,0 +1,58 @@
...
+
+            // save run result to db
+            let run = JobRunResult {
+                id: 0,
+                job_name,
+                start_time,
+                finish_time,
+                success: output.status.success(),
+                stdout: String::from_utf8_lossy(&output.stdout).to_string(),
+                stderr: String::from_utf8_lossy(&output.stderr).to_string(),
```
wsxiaoys, line 48: it'll be really nice to have stdout / stderr result being streammed update to database (e.g every lines trigger an update db operations.)

OK to do this in a following up PR

ee/tabby-webserver/migrations/05-job-runs-table/up.sql
```diff
@@ -0,0 +1,12 @@
...
+    job        VARCHAR(255) NOT NULL,
+    start_ts   TIMESTAMP NOT NULL,
+    end_ts     TIMESTAMP NOT NULL,
+    success    BOOLEAN NOT NULL,
+    stdout     TEXT NOT NULL,
+    stderr     TEXT NOT NULL,
+    created_at TIMESTAMP DEFAULT (DATETIME('now'))
+);
+
+CREATE INDEX idx_job ON job_runs (job);
```
wsxiaoys, line 12:
```suggestion
```

Let's consider indexing in future

wsxiaoys (changes requested):
Overall looks good, a few nits

ee/tabby-webserver/src/schema/mod.rs
wsxiaoys: Consider move this to job_runs.rs.

In future we'll still need to move this to `schema/mod.rs` (when we add it as part of graphql query), but it provides a better git history by putting to job_runs.rs for now.

ee/tabby-webserver/migrations/05-job-runs-table/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE job_runs (
+    id         INTEGER PRIMARY KEY AUTOINCREMENT,
+    job        VARCHAR(255) NOT NULL,
+    start_ts   TIMESTAMP NOT NULL,
+    end_ts     TIMESTAMP NOT NULL,
+    success    BOOLEAN NOT NULL,
```
wsxiaoys, line 6: exit_code int

ee/tabby-webserver/migrations/05-job-runs-table/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE job_runs (
+    id         INTEGER PRIMARY KEY AUTOINCREMENT,
+    job        VARCHAR(255) NOT NULL,
+    start_ts   TIMESTAMP NOT NULL,
+    end_ts     TIMESTAMP NOT NULL,
```
wsxiaoys, line 5: make this nullable (we wanna be able to track in-progress job as well)

ee/tabby-webserver/migrations/05-job-runs-table/up.sql
```diff
@@ -0,0 +1,12 @@
...
+    success    BOOLEAN NOT NULL,
+    stdout     TEXT NOT NULL,
+    stderr     TEXT NOT NULL,
+    created_at TIMESTAMP DEFAULT (DATETIME('now'))
```
wsxiaoys, line 9: add updated_at as well (since we plan to support stdout / stderr streamming)

ee/tabby-webserver/src/service/cron/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+pub fn run_cron(db_conn: DbConn) {
+    tokio::spawn(async move {
+        let Ok(job1) = db::refresh_token_job(db_conn.clone()).await else {
+            error!("failed to create db job");
+            return;
+        };
+        // run every 5 minutes
+        let Ok(job2) = repository_job(db_conn.clone(), "sync".to_owned(), "0 1/5 * * * * *").await
```
darknight, line 26: I want to point out one more time that IMHO the sync job schedules too frequent.

ee/tabby-webserver/src/service/cron/db.rs
darknight: Please help create a ticket for future improvement.

wsxiaoys (approved):

wsxiaoys:
Please also update the PR description

PR #1095, opened 2023-12-21
fix(clients): online documentation links

hduelme:
Currently the links to the extensions configuration result in `Page Not Found`. I added the missing `s` .


icycodes (approved):
Thank you for the fix!

PR #1104, opened 2023-12-22
docs: Replace dead health.rs link

Lash-L:
Hello! I noticed that the configuration docs points to a non-existing file for the health.rs configuration. It was incorrectly 'serve' instead of 'services'


wsxiaoys:
Thanks for the fix!

PR #1121, opened 2023-12-26
feat(core): add TABBY_MODEL_CACHE_ROOT to allow individually set tabb…

wsxiaoys:
…y model cache dir


linpan:
where default model path is in the docker ?

Issue #538, opened 2023-10-11
[TAB-438] Add a source code browser for ~/.tabby/repositories


wsxiaoys:
**Please describe the feature you desire:**

To implement the desired feature, modify the `~/.tabby/config.toml` configuration file as shown below:

```toml
[[repositories]]
git_url = "https://github.com/TabbyML/tabby.git"
```

After executing `tabby scheduler`, it will generate a `~/.tabby/repositories` directory with the following content:

```bash
meng@Mengs-MacBook-Pro:~/.tabby/repositories % ls
https_github.com_TabbyML_tabby.git
```

#### with Monaco
Example application: [React Monaco File Tree](https://codesandbox.io/s/react-monaco-file-tree-ww9kis?file=/src/App.tsx)

- Backend is will implemented by #815 

- [HoverProvider](https://stackoverflow.com/questions/49449401/monaco-editor-hoverprovider-get-the-word-mouse-hovers-over) to provide intelligence display for tags.

- [DocumentSymbolProvider](https://microsoft.github.io/monaco-editor/typedoc/interfaces/languages.DocumentSymbolProvider.html) to provides symbol navigation functionality

#### with CodeMirror

Example application: https://github.com/vizhub-core/vzcode

- [Hover Tooltip](https://codemirror.net/examples/tooltip/)

<sub>[TAB-438](https://linear.app/tabbyml/issue/TAB-438/add-a-source-code-browser-for-tabbyrepositories)</sub>

wsxiaoys:
Blocked by #815

wsxiaoys:
Looks slick https://x.com/musabdev/status/1727125032214950282?s=61

wsxiaoys:
TBD:

1. Directory view @jueliang

PR #1123, opened 2023-12-26
docs: Q1 2024 Roadmap

wsxiaoys:


wsxiaoys:
cc @gyxlucy for some comments

website/docs/roadmap.md
```diff
@@ -2,8 +2,20 @@
 
 We continuously work on updating our roadmap and we love to discuss those with our community. Feel encouraged to participate.
 
+## Q1 2024
+
+* 🎊 Supports embedding api for document context related use cases: https://github.com/TabbyML/tabby/issues/790
+* 🔐 SSO / OAuth support in Tabby EE: https://github.com/TabbyML/tabby/issues/1039
```
gyxlucy, line 8: How about `authorization` for EE? Like owner/admin/member access permissions. Feasible for Q1?

website/docs/roadmap.md
```diff
@@ -2,8 +2,20 @@
 
 We continuously work on updating our roadmap and we love to discuss those with our community. Feel encouraged to participate.
 
+## Q1 2024
+
+* 🎊 Supports embedding api for document context related use cases: https://github.com/TabbyML/tabby/issues/790
```
gyxlucy, line 7: Do we consider integrating this API into chat playground in Q1 as well? Once the API is ready, I suppose the integrations should be rather straightforward.

website/docs/roadmap.md
wsxiaoys: should be feasible - so far we only have two tiers of authorization (admin / non-admin)

website/docs/roadmap.md
wsxiaoys: This one is a bit tricky - it's still not clear how these context should be integrated.

One nice PoC built by sourcegraph folks is https://opencodegraph.org/ - I feel at least we could do some integration with our code browser (once it's implemented in #538).

The code browser could use some chat ui, probably.

gyxlucy: Cool let's research and discuss more in Q1 then.

icycodes (approved):

PR #1131, opened 2023-12-27
refactor(webserver): deprecate invitations / users api. add delete_in…

wsxiaoys:
…vitation_next

cc @darknight

### DIFF
--- a/ee/tabby-webserver/src/schema/auth.rs
+++ b/ee/tabby-webserver/src/schema/auth.rs
@@ -367,9 +367,7 @@ pub trait AuthenticationService: Send + Sync {
 
     async fn reset_user_auth_token(&self, email: &str) -> Result<()>;
 
-    async fn list_users(&self) -> Result<Vec<User>>;
-
-    async fn list_users_in_page(
+    async fn list_users(
         &self,
         after: Option<String>,
         before: Option<String>,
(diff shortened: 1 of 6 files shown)


ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -361,23 +362,20 @@ pub trait AuthenticationService: Send + Sync {
     async fn is_admin_initialized(&self) -> Result<bool>;
     async fn get_user_by_email(&self, email: &str) -> Result<User>;
 
-    async fn create_invitation(&self, email: String) -> Result<i32>;
-    async fn list_invitations(&self) -> Result<Vec<Invitation>>;
-    async fn delete_invitation(&self, id: i32) -> Result<i32>;
+    async fn create_invitation(&self, email: String) -> Result<ID>;
+    async fn delete_invitation(&self, id: ID) -> Result<ID>;
```
wsxiaoys, line 366: We should always use DTO objects (in this case, schema::InvitationNext / schema::User / ID), instead of DAO objects (db::Invitation / db::User / i32)

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -290,21 +303,40 @@ impl Mutation {
         ctx.locator.auth().refresh_token(refresh_token).await
     }
 
-    async fn create_invitation(ctx: &Context, email: String) -> Result<i32> {
+    async fn create_invitation(ctx: &Context, email: String) -> Result<ID> {
```
wsxiaoys, line 306: client side doesn't consumed the returned id, so we can change the return type directly (without creating a new create_invitation_next).

PR #1137, opened 2023-12-28
fix: use document.execCommand for non-secure context

### LINKED ISSUE #1134: Copy to clipboard doesn't work for non-secure context (e.g http://127.0.0.1:8080)
Location: https://github.com/TabbyML/tabby/blob/3339a6c967adbbf782299073d9c064352675c0f1/ee/tabby-ui/lib/hooks/use-copy-to-clipboard.tsx#L15

Related discussion: https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined

Need a workaround

liangfung:
Close #1134


ee/tabby-ui/lib/hooks/use-copy-to-clipboard.tsx
```diff
@@ -1,32 +1,50 @@
 'use client'
 
 import * as React from 'react'
+import copy from 'copy-to-clipboard'
+import { toast } from 'react-hot-toast'
```
wsxiaoys, line 5: shall we use shadn's toast component

ee/tabby-ui/app/layout.tsx
```diff
@@ -1,5 +1,6 @@
 import { Metadata } from 'next'
-import { Toaster } from 'react-hot-toast'
+
+import { Toaster } from '@/components/ui/toaster'
```
wsxiaoys, line 3: https://ui.shadcn.com/docs/components/sonner
https://ui.shadcn.com/docs/components/toast

Which one would you prefer? (in case you missed sonner)

liangfung: Yes, I missed sonner, sonner is better!

PR #1140, opened 2023-12-28
fix(webserver): ignore .git for /repositories resolve api

darknight:
Follow up of issue: https://github.com/TabbyML/tabby/issues/815#issuecomment-1870146269

- ignore `.git` directory from resolve api, return `404` directly
- add fallback handling

### DIFF
--- a/ee/tabby-webserver/src/repositories/mod.rs
+++ b/ee/tabby-webserver/src/repositories/mod.rs
@@ -23,10 +23,13 @@ use crate::{
 
 pub fn routes(locator: Arc<dyn ServiceLocator>) -> Router {
     Router::new()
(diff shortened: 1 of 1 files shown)


ee/tabby-webserver/src/repositories/mod.rs
```diff
@@ -23,10 +23,13 @@ use crate::{
 
 pub fn routes(locator: Arc<dyn ServiceLocator>) -> Router {
     Router::new()
+        .route("/:name/resolve/.git/", routing::get(not_found))
+        .route("/:name/resolve/.git/*path", routing::get(not_found))
```
wsxiaoys, line 27: nice trick

PR #1144, opened 2023-12-29
feat: Add a new event type:`dismiss`. Add `view_id` and `elapsed` in event body.

icycodes:
Complete TAB-382

### DIFF
--- a/crates/tabby-common/src/api/event.rs
+++ b/crates/tabby-common/src/api/event.rs
@@ -54,7 +54,7 @@ pub enum Event {
     Dismiss {
         completion_id: String,
         choice_index: u32,
-        
+
         #[serde(skip_serializing_if = "Option::is_none")]
         view_id: Option<String>,
         #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


wsxiaoys:
#1148

wsxiaoys:
is this ready for review?

wsxiaoys (approved):

crates/tabby-common/src/api/event.rs
```diff
@@ -3,14 +3,18 @@ use utoipa::ToSchema;
 
 #[derive(Serialize, Deserialize, ToSchema, Clone, Debug)]
 pub struct LogEventRequest {
-    /// Event type, should be `view` or `select`.
+    /// Event type, should be `view`, `select` or `dismiss`.
```
boxbeam, line 6: Maybe we should use an enum to model this, or `Discriminant<Event>`? I believe I forgot to modify this comment when adding the `Completion` variant - using `Discriminant<Event>` would make it clear through the type system what this is, and not require updating the comment every time a new event is added.

wsxiaoys: there's no need to modify this, completion event are not logged through this route.

boxbeam (approved):

PR #1145, opened 2023-12-30
fix(webserver): support resolve all types of repos

darknight:
Follow up of issue: https://github.com/TabbyML/tabby/issues/815#issuecomment-1870146269

- do not assume repo is located in `~/.tabby/repositories/`, instead, identify repo location based on the config (so load config lazily)
- on windows platform, the filepath in `~/.tabby/dataset/files.jsonl` is backslash, like:
```
{"git_url":"https://github.com/TabbyML/tabby.git","filepath":"python\\tabby-loadtest\\server.py","content":"......"", .....}
```
so add `os_path` method based on target os & apply it when build DatasetKey for querying meta info


ee/tabby-webserver/src/repositories/mod.rs
```diff
@@ -62,8 +67,11 @@ async fn not_found() -> StatusCode {
 
 #[instrument(skip(repo))]
 async fn resolve(Path(repo): Path<ResolveParams>) -> Result<Response, StatusCode> {
-    let root = repositories_dir().join(repo.name_str());
-    let full_path = root.join(repo.path_str());
+    let Some(conf) = CONF.find_repository(repo.name_str()) else {
```
wsxiaoys, line 70: instead of using a global variable, Conf can be passed in `attach_webserver` and being formed as a state here.

ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -47,6 +47,15 @@ impl ResolveParams {
...
         self.path.as_deref().unwrap_or("")
     }
+
+    #[cfg(not(target_os = "windows"))]
+    pub fn os_path(&self) -> String {
+        self.path.clone().unwrap_or_default()
+    }
+    #[cfg(target_os = "windows")]
+    pub fn os_path(&self) -> String {
+        self.path.clone().unwrap_or_default().replace("/", r"\")
```
wsxiaoys, line 57: is there a standard library function to handle such conversion?

wsxiaoys (commented):
Note for the comment in
> /repositories/resolve should returns all repositories, in DirEntry format.

It actual also means that we wanna return all repositories, in format of DirEntry, when user access `GET /repositories/resolve`

ee/tabby-webserver/src/repositories/resolve.rs
darknight: No such function from std AFAIK.

Actually, I was planning to use [path_slash](https://docs.rs/path-slash/latest/path_slash/) crate, but later thought that it wasn't worthwhile to import another dep for such simple case. But you can take a look at it and let me know your thoughts.

darknight:
> Note for the comment in
> 
> > /repositories/resolve should returns all repositories, in DirEntry format.
> 
> It actual also means that we wanna return all repositories, in format of DirEntry, when user access `GET /repositories/resolve`

Missed this point at first, now added this API.

crates/tabby/src/serve.rs
```diff
@@ -101,7 +101,7 @@ pub struct ServeArgs {
     webserver: bool,
 }
 
-pub async fn main(config: &Config, args: &ServeArgs) {
+pub async fn main(config: Config, args: &ServeArgs) {
```
wsxiaoys, line 104:
```suggestion
pub async fn main(config: &Config, args: &ServeArgs) {
```

crates/tabby/src/serve.rs
```diff
@@ -115,14 +115,15 @@ pub async fn main(config: &Config, args: &ServeArgs) {
 
     let logger = Arc::new(create_logger());
     let code = Arc::new(create_code_search());
+    let config = Arc::new(config);
 
-    let api = api_router(args, config, logger.clone(), code.clone()).await;
+    let api = api_router(args, config.as_ref(), logger.clone(), code.clone()).await;
```
wsxiaoys, line 120: Let's not creating config as Arc here

ee/tabby-webserver/src/handler.rs
```diff
@@ -22,6 +25,7 @@ pub async fn attach_webserver(
     ui: Router,
     logger: Arc<dyn RawEventLogger>,
     code: Arc<dyn CodeSearch>,
+    conf: Arc<Config>,
 ) -> (Router, Router) {
     let ctx = create_service_locator(logger, code).await;
     let schema = Arc::new(create_schema());
```
wsxiaoys, line 31: You might create Vec<Repositories> as `Arc` here, or create a wrapper struct, e.g
```rust
struct ResolveState {
  repositories: Vec<Repositories>
}
```

You can easily attach find_repository logic to this struct

ee/tabby-webserver/src/handler.rs
```diff
@@ -22,6 +25,7 @@ pub async fn attach_webserver(
     ui: Router,
     logger: Arc<dyn RawEventLogger>,
     code: Arc<dyn CodeSearch>,
+    conf: Arc<Config>,
```
wsxiaoys, line 28: avoid use abbr for short words, just use config
```suggestion
    config: &Config,
```

ee/tabby-webserver/src/repositories/mod.rs
```diff
@@ -12,23 +12,28 @@ use axum::{
...
+    repositories::resolve::{
+        resolve_all, resolve_dir, resolve_file, resolve_meta, Meta, ResolveParams,
+    },
     schema::ServiceLocator,
 };
 
-pub fn routes(locator: Arc<dyn ServiceLocator>) -> Router {
+pub fn routes(conf: Arc<Config>, locator: Arc<dyn ServiceLocator>) -> Router {
     Router::new()
+        .route("/resolve/", routing::get(resolve))
```
wsxiaoys, line 27: will `/resolve` resolve to this path?

ee/tabby-webserver/src/repositories/resolve.rs
wsxiaoys: sg. Then let's use `cfg! statement`

```rust
if cfg!(target_os != "windows") {
   ...
} else {
  ...
}
```

wsxiaoys (changes requested):

ee/tabby-webserver/src/repositories/mod.rs
darknight: no, it causes `404`. I added both paths (with & without) to prevent this issue.

ee/tabby-webserver/src/handler.rs
```diff
@@ -8,7 +8,10 @@ use axum::{
 };
 use hyper::Body;
```
darknight, line 9: I was also wondering why `handler.rs` duplicates the code from `lib.rs`, why not keep only one copy?

wsxiaoys: it's a mistake - we shall only keep the content inside `handler.rs`

ee/tabby-webserver/src/handler.rs
```diff
@@ -22,9 +25,13 @@ pub async fn attach_webserver(
     ui: Router,
     logger: Arc<dyn RawEventLogger>,
     code: Arc<dyn CodeSearch>,
+    conf: &Config,
```
wsxiaoys, line 28:
```suggestion
    config: &Config,
```

wsxiaoys:
Just one single nit, otherwise LGTM

Issue #1039, opened 2023-12-14
[TAB-440] OAuth SSO Support 


wsxiaoys:
At first step, we'd like to support following identity providers:

* GIthub
* Google

---

- [X] Create sql table to store Google / Github authentication credentials (each identity provider should have its own table).
- [X] Implement graphql API to read / set credentials.
- [X] Frontend settings page to modify the credentials.
- [X] Implement backend oauth authentication flow callback endpoints.
- [X] Frontend oauth flow implementation.

<sub>From [SyncLinear.com](https://synclinear.com) | [TAB-440](https://linear.app/tabbyml/issue/TAB-440/oauth-sso-support)</sub>

PR #1160, opened 2024-01-04
feat(webserver): Add github oauth support

darknight:

The scope of original ticket is big, so I'm gonna break it down, raise smaller PR to implement the target bit by bit.

Changes in this PR:
- add db schema, db r/w operations for saving `Github` oauth credential
  - considering there's only one row in the table, I just felt we can combine `insert` & `update` together, so I implemented `upsert_github_oauth_credential`
- similar to `Worker` & `WorkerKind` for worker service, I added `OAuthProvider` & `OAuthCredential` as new graphql objects
- defined 3 new graphql api for `read, write, delete` credential respectively
- updated `AuthenticationService`, added 3 interfaces to glue db DAO & public graphql api

### DIFF
--- a/ee/tabby-db/src/github_oauth_credential.rs
+++ b/ee/tabby-db/src/github_oauth_credential.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use chrono::{DateTime, Utc};
-use rusqlite::{params, OptionalExtension};
+use rusqlite::{params, OptionalExtension, named_params};
 
 use super::DbConn;
 
@@ -41,15 +41,16 @@ impl DbConn {
             .call(move |c| {
                 let mut stmt = c.prepare(
                     r#"INSERT INTO github_oauth_credential (id, client_id, client_secret)
-                    VALUES (?, ?, ?) ON CONFLICT(id) DO UPDATE
-                    SET client_id = ?, client_secret = ?, active = ?, updated_at = datetime('now')
+                    VALUES (:id, :cid, :secret) ON CONFLICT(id) DO UPDATE
+                    SET client_id = :cid, client_secret = :secret, active = :active, updated_at = datetime('now')
                     WHERE id = ?"#,
                 )?;
-                stmt.insert(params![
-                    GITHUB_OAUTH_CREDENTIAL_ROW_ID, client_id, client_secret,
-                    client_id, client_secret, active,
-                    GITHUB_OAUTH_CREDENTIAL_ROW_ID
-                ])?;
+                stmt.insert(named_params! {
+                    ":id": GITHUB_OAUTH_CREDENTIAL_ROW_ID,
+                    ":cid": client_id,
+                    ":secret": client_secret,
+                    ":active": active,
+                })?;
                 Ok(())
             })
             .await?;
@@ -89,16 +90,15 @@ mod tests {
         let res = conn.read_github_oauth_credential().await.unwrap().unwrap();
         assert_eq!(res.client_id, "client_id");
         assert_eq!(res.client_secret, "client_secret");
-        assert_eq!(res.active, true);
+        assert!(res.active);
 
         // test update
-        let id = conn
-            .update_github_oauth_credential("client_id", "client_secret_2", false)
+        conn.update_github_oauth_credential("client_id", "client_secret_2", false)
             .await
             .unwrap();
         let res = conn.read_github_oauth_credential().await.unwrap().unwrap();
         assert_eq!(res.client_id, "client_id");
         assert_eq!(res.client_secret, "client_secret_2");
-        assert_eq!(res.active, false);
+        assert!(!res.active);
     }
 }
--- a/ee/tabby-db/src/github_oauth_credential.rs
+++ b/ee/tabby-db/src/github_oauth_credential.rs
@@ -1,0 +1,106 @@
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use rusqlite::{params, OptionalExtension};
+
+use super::DbConn;
+
+const GITHUB_OAUTH_CREDENTIAL_ROW_ID: i32 = 1;
+
+pub struct GithubOAuthCredentialDAO {
+    pub client_id: String,
+    pub client_secret: String,
+    pub active: bool,
+    pub created_at: DateTime<Utc>,
+    pub updated_at: DateTime<Utc>,
+}
+
+impl GithubOAuthCredentialDAO {
+    fn from_row(row: &rusqlite::Row<'_>) -> std::result::Result<Self, rusqlite::Error> {
+        Ok(Self {
+            client_id: row.get(0)?,
+            client_secret: row.get(1)?,
+            active: row.get(2)?,
+            created_at: row.get(3)?,
+            updated_at: row.get(4)?,
+        })
+    }
+}
+
+/// db read/write operations for `github_oauth_credential` table
+impl DbConn {
+    pub async fn update_github_oauth_credential(
+        &self,
+        client_id: &str,
+        client_secret: &str,
+        active: bool,
+    ) -> Result<()> {
+        let client_id = client_id.to_string();
+        let client_secret = client_secret.to_string();
+
+        self.conn
+            .call(move |c| {
+                let mut stmt = c.prepare(
+                    r#"INSERT INTO github_oauth_credential (id, client_id, client_secret)
+                    VALUES (?, ?, ?) ON CONFLICT(id) DO UPDATE
+                    SET client_id = ?, client_secret = ?, active = ?, updated_at = datetime('now')
+                    WHERE id = ?"#,
+                )?;
+                stmt.insert(params![
+                    GITHUB_OAUTH_CREDENTIAL_ROW_ID, client_id, client_secret,
+                    client_id, client_secret, active,
+                    GITHUB_OAUTH_CREDENTIAL_ROW_ID
+                ])?;
+                Ok(())
+            })
+            .await?;
+
+        Ok(())
+    }
+
+    pub async fn read_github_oauth_credential(&self) -> Result<Option<GithubOAuthCredentialDAO>> {
+        let token = self
+            .conn
+            .call(|conn| {
+                Ok(conn
+                    .query_row(
+                        r#"SELECT client_id, client_secret, active, created_at, updated_at FROM github_oauth_credential WHERE id = ?"#,
+                        [GITHUB_OAUTH_CREDENTIAL_ROW_ID],
+                        GithubOAuthCredentialDAO::from_row,
+                    )
+                    .optional())
+            })
+            .await?;
+
+        Ok(token?)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_operations_on_github_oauth_credential() {
+        // test insert
+        let conn = DbConn::new().await.unwrap();
+        let id = conn
+            .update_github_oauth_credential("client_id", "client_secret")
+            .await
+            .unwrap();
+        assert_eq!(id, 1);
+
+        // test update
+        let id = conn
+            .update_github_oauth_credential("client_id", "client_secret_2")
+            .await
+            .unwrap();
+        let res = conn.read_github_oauth_credential().await.unwrap().unwrap();
+        assert_eq!(res.id, id);
+        assert_eq!(res.client_secret, "client_secret_2");
+
+        // test delete
+        conn._github_oauth_credential().await.unwrap();
+        let res = conn.read_github_oauth_credential().await.unwrap();
+        assert!(res.is_none());
+    }
+}
--- a/ee/tabby-db/src/github_oauth_credential.rs
+++ b/ee/tabby-db/src/github_oauth_credential.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use chrono::{DateTime, Utc};
-use rusqlite::{params, OptionalExtension, named_params};
+use rusqlite::{named_params, OptionalExtension};
 
 use super::DbConn;
 
--- a/ee/tabby-db/src/github_oauth_credential.rs
+++ b/ee/tabby-db/src/github_oauth_credential.rs
@@ -43,7 +43,7 @@ impl DbConn {
                     r#"INSERT INTO github_oauth_credential (id, client_id, client_secret)
                     VALUES (:id, :cid, :secret) ON CONFLICT(id) DO UPDATE
                     SET client_id = :cid, client_secret = :secret, active = :active, updated_at = datetime('now')
-                    WHERE id = ?"#,
+                    WHERE id = :id"#,
                 )?;
                 stmt.insert(named_params! {
                     ":id": GITHUB_OAUTH_CREDENTIAL_ROW_ID,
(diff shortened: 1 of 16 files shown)


ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -0,0 +1,123 @@
...
+        Ok(id as i32)
+    }
+
+    pub async fn get_github_oauth_credential(&self) -> Result<Option<GithubOAuthCredentialDAO>> {
+        let token = self
+            .conn
+            .call(|conn| {
+                Ok(conn
+                    .query_row(
+                        r#"SELECT id, client_id, client_secret, created_at, updated_at FROM github_oauth_credential"#,
```
wsxiaoys, line 71: I feel it’s better to have a static variable with `GITHUB_OAUTH_CREDENTIAL_ROW_ID = 1` to make query / updating explicit.

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -0,0 +1,123 @@
...
+                        GithubOAuthCredentialDAO::from_row,
+                    )
+                    .optional())
+            })
+            .await?;
+
+        Ok(token?)
+    }
+
+    pub async fn delete_github_oauth_credential(&self) -> Result<()> {
```
wsxiaoys, line 82: instead of deleting, i feel it’s better to have a boolean column in table `active` to mark whether the github oauth is enabled.

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -0,0 +1,123 @@
...
+            client_secret: row.get(2)?,
+            created_at: row.get(3)?,
+            updated_at: row.get(4)?,
+        })
+    }
+}
+
+/// db read/write operations for `github_oauth_credential` table
+impl DbConn {
+    pub async fn upsert_github_oauth_credential(
```
wsxiaoys, line 29:
```suggestion
    pub async fn update_github_oauth_credential(
```

I feel the semantic here can be restrict to single row operations

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -0,0 +1,123 @@
...
+                )?;
+                let row_id = stmt.insert(params![client_id, client_secret])?;
+                Ok(row_id)
+            })
+            .await?;
+
+        Ok(id as i32)
+    }
+
+    pub async fn get_github_oauth_credential(&self) -> Result<Option<GithubOAuthCredentialDAO>> {
```
wsxiaoys, line 65:
```suggestion
    pub async fn read_github_oauth_credential(&self) -> Result<Option<GithubOAuthCredentialDAO>> {
```

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -0,0 +1,123 @@
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use rusqlite::{params, OptionalExtension};
+
+use super::DbConn;
+
+pub struct GithubOAuthCredentialDAO {
+    pub id: i32,
```
wsxiaoys, line 8:
```suggestion
```

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -0,0 +1,123 @@
...
+                stmt.execute(params![])?;
+                Ok(())
+            })
+            .await?;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
```
wsxiaoys, line 95: In future, i recommend add unit test after the interface got reviewed

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -354,6 +370,15 @@ pub trait AuthenticationService: Send + Sync {
...
     ) -> Result<Vec<InvitationNext>>;
+
+    async fn upsert_oauth_credential(
+        &self,
+        provider: OAuthProvider,
+        client_id: String,
+        client_secret: String,
+    ) -> Result<OAuthCredential>;
+    async fn delete_oauth_credential(&self, provider: OAuthProvider) -> Result<()>;
+    async fn get_oauth_credential(&self, provider: OAuthProvider) -> Result<OAuthCredential>;
```
wsxiaoys, line 381: I’ll suggest skip the graphql api for now. So you can complete the end to end authorization flow by:
1. implementing the oauth callback.
2. set credentials by editing the sqlite database directly.

Once we finished 1 & 2, we can get back and review the graphql design.

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -1,23 +1,28 @@
 use anyhow::Result;
 use chrono::{DateTime, Utc};
+use lazy_static::lazy_static;
 use rusqlite::{params, OptionalExtension};
 
 use super::DbConn;
 
+lazy_static! {
```
wsxiaoys, line 8: no need to use lazy_static since it's barely a i32

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -26,50 +31,50 @@ impl GithubOAuthCredentialDAO {
...
         client_id: &str,
         client_secret: &str,
-    ) -> Result<i32> {
+        active: bool,
+    ) -> Result<()> {
         let client_id = client_id.to_string();
         let client_secret = client_secret.to_string();
 
-        if let Ok(Some(cred)) = self.get_github_oauth_credential().await {
+        if self.read_github_oauth_credential().await?.is_some() {
```
wsxiaoys, line 43: can we just use `UPSERT`? https://www.sqlite.org/lang_upsert.html

ee/tabby-webserver/src/schema/auth.rs
darknight: For the callback workflow, I've added new api
`/oauth/github/callback?code=ba327a......04a14db8`
and it produces following response:
```
{
    "access_token": "gho_G4ar.................1nmqYM",
    "scope": "user:email",
    "token_type": "bearer",
    "error": "",
    "error_description": "",
    "error_uri": ""
}
```

Is this supposed to expose to frontend? or we should fetch user info with the access token on the server side? @wsxiaoys

ee/tabby-webserver/src/oauth/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+#[derive(Deserialize)]
+struct GithubCallbackParams {
+    code: String,
+    state: Option<String>,
+}
+
+async fn github_callback(
+    State(state): State<Arc<dyn ServiceLocator>>,
```
wsxiaoys, line 26: I feel that it might be better to just inject `DbConn` directly, instead of relying on auth service, on this particular route.

So you don't have to do a wrapper layer in AuthenticationService. (AuthenticationService's target is to decouple graphql api to their dependencies, while for other routes, i feel it's not necessary).

ee/tabby-webserver/src/schema/auth.rs
wsxiaoys: As suggested above, I feel the github callback route can just use DbConn.

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -21,9 +24,8 @@ use tracing::error;
 use validator::ValidationErrors;
 use worker::{Worker, WorkerService};
 
-use crate::schema::auth::{OAuthCredential, OAuthProvider};
-
 pub trait ServiceLocator: Send + Sync {
+    fn https_client(&self) -> Client<HttpsConnector<HttpConnector>>;
```
wsxiaoys, line 28: as suggested above, I feel this client class should be a member of state struct specific to oauth route, instead of puting inside service locator (which is specific for graphql apis)

ee/tabby-webserver/src/handler.rs
```diff
@@ -48,7 +43,8 @@ pub async fn attach_webserver(
         .nest(
             "/repositories",
             repositories::routes(rs.clone(), ctx.clone()),
-        );
+        )
+        .nest("/oauth", oauth::routes(ctx.clone()));
```
wsxiaoys, line 47: `/oauth_callback/{provider}`

ee/tabby-webserver/src/oauth/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+    Query(params): Query<GithubCallbackParams>,
+) -> Result<Json<GithubOAuthResponse>, StatusCode> {
+    let resp = github::access_token(state, params.code)
```
wsxiaoys, line 29: need to check whether the github oauth is actual active (return 404 when active = false)

ee/tabby-webserver/src/handler.rs
wsxiaoys: please also update https://github.com/TabbyML/tabby/blob/main/ee/tabby-webserver/docs/api_spec.md

ee/tabby-webserver/src/oauth/github.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    #[serde(default)]
+    error: String,
+    #[serde(default)]
+    error_description: String,
+    #[serde(default)]
+    error_uri: String,
+}
+
+pub async fn access_token(
```
wsxiaoys, line 25: I feel we should just read the email directly from github to handle the "authentication", what do you think? (it can be done in a followup PR) after acquiring the access token.

ee/tabby-webserver/src/schema/auth.rs
wsxiaoys: It shouldn't be passed to frontend, instead, we should generate a redirect request to `/auth/signin?refreshToken=XXXXX` to finish signin process (the frontend will only see the new refreshtoken)

More comments here: https://github.com/TabbyML/tabby/pull/1160/files#r1442595004

ee/tabby-webserver/src/oauth/mod.rs
wsxiaoys: I'll suggest let ServiceLocator returns `DbConn` directly, to avoid creating a dummy wrapper inside of AuthenticationService.

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -354,6 +370,8 @@ pub trait AuthenticationService: Send + Sync {
         first: Option<usize>,
         last: Option<usize>,
     ) -> Result<Vec<InvitationNext>>;
+
+    async fn get_oauth_credential(&self, provider: OAuthProvider) -> Result<OAuthCredential>;
```
wsxiaoys, line 374: instead of reading oauth here, I'll suggest just creating an function to actual wrap the authentication logic, e.g

```rust
fn github_auth(...) ->  std::result::Result<TokenAuthResponse, TokenAuthError> {
}
```

ee/tabby-webserver/docs/api_spec.md
```diff
@@ -141,3 +141,27 @@ The `Content-Type` for successful response is always `application/json`.
...
+**Method:** `GET`
+
+**Request example:**
+
+```shell
+curl --request GET \
+  --url http://localhost:8080/oauth_callback/github?code=1234567890
+```
+
+**Response example:**
```
wsxiaoys, line 160: response should be an 302 temp redirect with accessToken / refreshToken query parameters.

ee/tabby-webserver/src/oauth/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+use tracing::error;
+
+use crate::schema::{
+    auth::{AuthenticationService, GithubAuthError, GithubAuthResponse},
+    ServiceLocator,
+};
+
+pub(crate) mod github;
+
+#[derive(Clone)]
```
wsxiaoys, line 20: instead of marking this as clone, does it better to wrap this state inside of `Arc`?

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -145,6 +147,30 @@ pub enum TokenAuthError {
...
+    refresh_token: String,
+}
+
+#[derive(Error, Debug)]
+pub enum GithubAuthError {
+    #[error("The code passed is incorrect or expired")]
+    InvalidVerificationCode,
+
+    #[error("The Github credential is not found")]
+    CredentialNotFound,
```
wsxiaoys, line 162: seems can be merged with CredentialNotActive

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -310,6 +336,22 @@ impl relay::NodeType for InvitationNext {
...
+#[derive(GraphQLObject)]
+pub struct OAuthCredential {
+    pub provider: OAuthProvider,
+    pub client_id: String,
+    pub client_secret: String,
+    pub active: bool,
+    pub created_at: DateTime<Utc>,
+    pub updated_at: DateTime<Utc>,
+}
+
```
wsxiaoys, line 354:
```suggestion
```

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -339,6 +345,27 @@ impl AuthenticationService for DbConn {
...
+        let credential: OAuthCredential = self
+            .read_github_oauth_credential()
+            .await?
+            .ok_or(GithubAuthError::CredentialNotFound)?
+            .into();
+        if !credential.active {
+            return Err(GithubAuthError::CredentialNotActive);
+        }
+
+        let _email = github::fetch_user_email(code, credential, client).await?;
```
wsxiaoys, line 363: just pass GithubOAuthCredentialDAO

ee/tabby-webserver/src/oauth/github.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[derive(Debug, Deserialize)]
+struct GithubUserEmail {
+    email: String,
+    primary: bool,
+    verified: bool,
+    visibility: String,
+}
+
+async fn exchange_access_token(
```
wsxiaoys, line 33: is it better to wrap an GithubClient around the Client<HttpsClient<...>>, and implement fetch_user_email as its member function?

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        self.conn
+            .call(move |c| {
+                let mut stmt = c.prepare(
+                    r#"INSERT INTO github_oauth_credential (id, client_id, client_secret)
+                    VALUES (?, ?, ?) ON CONFLICT(id) DO UPDATE
+                    SET client_id = ?, client_secret = ?, active = ?, updated_at = datetime('now')
+                    WHERE id = ?"#,
+                )?;
+                stmt.insert(params![
```
wsxiaoys, line 48: you could use positioned parameter or named parameter to avoid passing args multiple times

https://docs.rs/rusqlite/latest/rusqlite/trait.Params.html#example-positional

ee/tabby-webserver/src/oauth/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    schema::{
+        auth::{AuthenticationService, GithubAuthError},
+        ServiceLocator,
+    },
+};
+
+pub mod github;
+
+#[non_exhaustive]
+struct OAuthState {
```
wsxiaoys, line 23: since all fields of this struct are already `Arc`, I suggest just derive `Clone` for the struct

ee/tabby-webserver/src/service/mod.rs
wsxiaoys, line 1: revert change

wsxiaoys (commented):
Otherwise LGTM

ee/tabby-db/src/github_oauth_credential.rs
darknight: This is really helpful! 👍

PR #1161, opened 2024-01-04
Pass logger through to chat service and log completion chunks as they are generated

### LINKED ISSUE #1149: Add event logging for `/v1beta/chat/completions`
Add event logging for `/v1beta/chat/completions`. Logic should be similar to what we do for logging `/v1/completion`.

References:

Example for logging `/v1/completions` events: 
- Schema: https://github.com/TabbyML/tabby/blob/1e9ce3e2c658fdc32790a11c4e029898df03fa94/crates/tabby-common/src/api/event.rs#L42
- Logging: https://github.com/TabbyML/tabby/blob/1e9ce3e2c658fdc32790a11c4e029898df03fa94/crates/tabby/src/services/completion.rs#L244

boxbeam:
Closes #1149 

Not sure my approach for getting the prompt was correct, please check that and let me know if there's a better method.

### DIFF
--- a/crates/tabby-common/src/api/event.rs
+++ b/crates/tabby-common/src/api/event.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
@@ -48,6 +50,11 @@ pub enum Event {
         choices: Vec<Choice>,
         #[serde(skip_serializing_if = "Option::is_none")]
         user: Option<String>,
+    },
+    ChatCompletion {
+        completion_id: String,
+        prompt: Arc<String>,
+        content: String,
     },
 }
 
--- a/crates/tabby/Cargo.toml
+++ b/crates/tabby/Cargo.toml
@@ -24,7 +24,7 @@ hyper = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 utoipa = { workspace = true, features = ["axum_extras", "preserve_order"] }
 utoipa-swagger-ui = { version = "3.1", features = ["axum"] }
-serde = { workspace = true }
+serde = { workspace = true, features = ["rc"] }
 serdeconv = { workspace = true }
 serde_json = { workspace = true }
 tower-http = { version = "0.4.0", features = ["cors", "timeout"] }
--- a/crates/tabby/src/routes/chat.rs
+++ b/crates/tabby/src/routes/chat.rs
@@ -7,6 +7,7 @@ use axum::{
     response::{IntoResponse, Response},
     Json,
 };
+use tabby_common::api::event::Event::ChatCompletion;
 use tracing::instrument;
 
 use crate::services::chat::{ChatCompletionRequest, ChatService};
@@ -30,8 +31,11 @@ pub async fn chat_completions(
     State(state): State<Arc<ChatService>>,
     Json(request): Json<ChatCompletionRequest>,
 ) -> Response {
+    let prompt = Arc::new(request.prompt());
     let s = stream! {
         for await content in state.generate(&request).await {
+
+            let id = content.id.clone();
             let content = match serde_json::to_string(&content) {
                 Ok(s) => s,
                 Err(e) => {
@@ -41,6 +45,9 @@ pub async fn chat_completions(
             };
 
             let content = format!("data: {}\n\n", content);
+            let event = ChatCompletion { completion_id: id, prompt: prompt.clone(), content: content.clone() };
+            state.logger.log(event);
+
             yield Ok(content)
         }
     };
--- a/crates/tabby/src/services/chat.rs
+++ b/crates/tabby/src/services/chat.rs
@@ -6,6 +6,7 @@ use async_stream::stream;
 use chat_prompt::ChatPromptBuilder;
 use futures::stream::BoxStream;
 use serde::{Deserialize, Serialize};
+use tabby_common::api::event::EventLogger;
 use tabby_inference::{TextGeneration, TextGenerationOptions, TextGenerationOptionsBuilder};
 use tracing::debug;
 use utoipa::ToSchema;
@@ -34,7 +35,7 @@ pub struct Message {
 
 #[derive(Serialize, Deserialize, ToSchema, Clone, Debug)]
 pub struct ChatCompletionChunk {
-    id: String,
+    pub(crate) id: String,
     created: u64,
     system_fingerprint: String,
     object: &'static str,
@@ -75,13 +76,26 @@ impl ChatCompletionChunk {
 
 pub struct ChatService {
     engine: Arc<dyn TextGeneration>,
+    pub(crate) logger: Arc<dyn EventLogger>,
     prompt_builder: ChatPromptBuilder,
 }
 
+impl ChatCompletionRequest {
+    pub fn prompt(&self) -> String {
+        // Doubt this is the right way to find the prompt.
+        self.messages.iter().map(|m| &*m.content).collect()
+    }
+}
+
 impl ChatService {
-    fn new(engine: Arc<dyn TextGeneration>, chat_template: String) -> Self {
+    fn new(
+        engine: Arc<dyn TextGeneration>,
+        logger: Arc<dyn EventLogger>,
+        chat_template: String,
+    ) -> Self {
         Self {
             engine,
+            logger,
             prompt_builder: ChatPromptBuilder::new(chat_template),
         }
     }
@@ -118,7 +132,12 @@ impl ChatService {
     }
 }
 
-pub async fn create_chat_service(model: &str, device: &Device, parallelism: u8) -> ChatService {
+pub async fn create_chat_service(
+    logger: Arc<dyn EventLogger>,
+    model: &str,
+    device: &Device,
+    parallelism: u8,
+) -> ChatService {
     let (engine, model::PromptInfo { chat_template, .. }) =
         model::load_text_generation(model, device, parallelism).await;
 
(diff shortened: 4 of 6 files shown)


crates/tabby-common/src/api/event.rs
```diff
@@ -49,6 +51,11 @@ pub enum Event {
         #[serde(skip_serializing_if = "Option::is_none")]
         user: Option<String>,
     },
+    ChatCompletion {
+        completion_id: String,
+        prompt: Arc<String>,
```
wsxiaoys, line 56: log input prompt as format of `Vec<Message>`, and output in format of `Message`

https://github.com/TabbyML/tabby/blob/0977b78e6555a3faa7ec825046f461c15fd9315d/crates/tabby/src/services/chat.rs#L26

crates/tabby/src/routes/chat.rs
```diff
@@ -41,6 +45,9 @@ pub async fn chat_completions(
             };
 
             let content = format!("data: {}\n\n", content);
+            let event = ChatCompletion { completion_id: id, prompt: prompt.clone(), content: content.clone() };
```
wsxiaoys, line 48: the logged event should be per request, instead of per chunk.

You might use a string accumulator to collect the output message, e.g
```rust
let output_message = “”

…
output_message += content
…

// use output_message in event
```

wsxiaoys (changes requested):

crates/tabby-common/src/api/event.rs
boxbeam: Can't use `Vec<Message>` in the event because `Event` is in `tabby-common` and `Message` is in `tabby`.

crates/tabby/src/services/chat.rs
```diff
@@ -75,13 +76,26 @@ impl ChatCompletionChunk {
...
+        engine: Arc<dyn TextGeneration>,
+        logger: Arc<dyn EventLogger>,
+        chat_template: String,
+    ) -> Self {
         Self {
             engine,
+            logger,
             prompt_builder: ChatPromptBuilder::new(chat_template),
         }
     }
```
wsxiaoys, line 101: i recommend implement the logging logic inside of `generate` function here, so you don’t have to expose logger as public.

This is also what we done for `completion.rs` service

boxbeam: Looking into this and finding that the issue is we'll have to serialize each chunk twice. Once in the generator to create the log output lines, and once in the endpoint to create the response. Is this an acceptable overhead?

boxbeam: For now I will just push the string content into a Vec.

crates/tabby-common/src/api/event.rs
wsxiaoys: You could duplicate the `Message` here, just like what we did for `Segments` (which is also a duplication of what defined in `tabby`)

This is also a best practice when storing serialized json - to separate the schema definition between what filled in request / what got logged.

wsxiaoys:
When finishing above implementation, please test a few prompts in `/playground` and verify the logged events locates at `~/.tabby/events` (you might paste a few logged json lines here)

crates/tabby-common/src/api/event.rs
```diff
@@ -49,6 +49,17 @@ pub enum Event {
         #[serde(skip_serializing_if = "Option::is_none")]
         user: Option<String>,
     },
+    ChatCompletion {
+        completion_id: String,
+        prompt: Vec<Message>,
```
wsxiaoys, line 54:
```suggestion
        input: Vec<Message>,
```

crates/tabby-common/src/api/event.rs
```diff
@@ -49,6 +49,17 @@ pub enum Event {
...
+        content: Vec<String>,
```
wsxiaoys, line 55:
```suggestion
        output: Message,
```

crates/tabby-common/src/api/event.rs
```diff
@@ -49,6 +49,17 @@ pub enum Event {
         #[serde(skip_serializing_if = "Option::is_none")]
         user: Option<String>,
     },
+    ChatCompletion {
+        completion_id: String,
+        input: Vec<Message>,
+        output: Vec<String>,
```
wsxiaoys, line 55: output format should be `Message`, as precious comment suggested

wsxiaoys (changes requested):

crates/tabby/src/services/chat.rs
```diff
@@ -34,7 +35,7 @@ pub struct Message {
 
 #[derive(Serialize, Deserialize, ToSchema, Clone, Debug)]
 pub struct ChatCompletionChunk {
-    id: String,
+    pub(crate) id: String,
```
wsxiaoys, line 38: no longer needed?

crates/tabby-common/src/api/event.rs
boxbeam: The only extra data associated with that is the role, which is not generated by the stream. There are multiple input messages, each which holds its own role. Will they ever differ, or are they all the same? If they're all the same, I can just store the first role that is received and associate it with each output chunk.

crates/tabby/src/services/chat.rs
```diff
@@ -99,32 +106,52 @@ impl ChatService {
         &self,
         request: &ChatCompletionRequest,
     ) -> BoxStream<ChatCompletionChunk> {
+        let mut log_output = vec![];
```
wsxiaoys, line 109: consider create an helper struct `ChatCompletionEventBuilder` to handle the appending / event building logic

wsxiaoys (changes requested):

crates/tabby/src/services/chat.rs
boxbeam: Started doing that, it wasn't looking particularly helpful since the only real logic that needs to be pulled out is converting the message list, so I wrote a helper function for that instead.

boxbeam:
Sample log output:
```
{"ts":1704386657477,"event":{"chat_completion":{"completion_id":"chatcmpl-a76384bc-e009-48a4-b7e3-f8a05de4791a","input":[{"role":"user","content":"hello world in python"}],"output":{"role":"assistant","content":"```python\nprint(\"hello world\")\n```"}}}}
```

crates/tabby/src/routes/chat.rs
wsxiaoys, line 1: revert changes to this file

crates/tabby/src/services/chat.rs
```diff
@@ -99,32 +106,65 @@ impl ChatService {
...
+            let event = Event::ChatCompletion { completion_id: id, input: log_prompt, output: assistant_message(log_output) };
+            self.logger.log(event);
         };
 
         Box::pin(s)
     }
 }
 
-pub async fn create_chat_service(model: &str, device: &Device, parallelism: u8) -> ChatService {
+fn assistant_message(string: String) -> tabby_common::api::event::Message {
```
wsxiaoys, line 136:
```suggestion
fn create_assistant_message(string: String) -> tabby_common::api::event::Message {
```

crates/tabby/src/services/chat.rs
```diff
@@ -99,32 +106,65 @@ impl ChatService {
         &self,
         request: &ChatCompletionRequest,
     ) -> BoxStream<ChatCompletionChunk> {
+        let mut log_output = String::new();
```
wsxiaoys, line 109:
```suggestion
        let mut event_output = String::new();
```

crates/tabby/src/services/chat.rs
```diff
@@ -99,32 +106,65 @@ impl ChatService {
...
+        let log_prompt = convert_messages(&request.messages);
```
wsxiaoys, line 110:
```suggestion
        let event_input = convert_messages(&request.messages);
```

wsxiaoys (changes requested):
Otherwise LGTM

crates/tabby/Cargo.toml
```diff
@@ -24,7 +24,7 @@ hyper = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 utoipa = { workspace = true, features = ["axum_extras", "preserve_order"] }
 utoipa-swagger-ui = { version = "3.1", features = ["axum"] }
-serde = { workspace = true }
+serde = { workspace = true, features = ["rc"] }
```
wsxiaoys, line 27: is this necessary?

crates/tabby/src/services/chat.rs
```diff
@@ -99,32 +106,65 @@ impl ChatService {
...
         let s = stream! {
             for await content in self.engine.generate_stream(&prompt, options).await {
+                log_output.push_str(&content);
                 yield ChatCompletionChunk::new(content, id.clone(), created, false)
             }
-            yield ChatCompletionChunk::new("".into(), id, created, true)
+            yield ChatCompletionChunk::new("".into(), id.clone(), created, true);
+
+            let event = Event::ChatCompletion { completion_id: id, input: log_prompt, output: assistant_message(log_output) };
+            self.logger.log(event);
```
wsxiaoys, line 129:
```suggestion
           self.logger.log(Event::ChatCompletion {
             ...
           })
```

crates/tabby/src/services/chat.rs
```diff
@@ -99,32 +106,65 @@ impl ChatService {
...
+fn convert_messages(input: &Vec<Message>) -> Vec<tabby_common::api::event::Message> {
+    let mut output = vec![];
+
+    for message in input {
+        output.push(tabby_common::api::event::Message {
+            content: message.content.clone(),
+            role: message.role.clone(),
+        })
+    }
+
```
wsxiaoys, line 152:
```suggestion
  return input.iter().map(|x| ...).collect()
```

crates/tabby/Cargo.toml
boxbeam: Not anymore, I had used it in a previous version but removed the rc that was stored inside a serialized struct

wsxiaoys (approved):

PR #1168, opened 2024-01-05
docs: Crate doc comments

boxbeam:
Add basic descriptions of most tabby crates in doc comments

### DIFF
--- a/crates/llama-cpp-bindings/src/lib.rs
+++ b/crates/llama-cpp-bindings/src/lib.rs
@@ -1,3 +1,4 @@
+//! Bindings to raw C++ LLaMA implementation. Used alongside tabby-inference to generate text and code.
 mod llama;
 mod utils;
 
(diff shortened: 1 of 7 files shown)


crates/llama-cpp-bindings/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+//! Bindings to raw C++ LLaMA implementation. Used alongside tabby-inference to generate text and code.
```
wsxiaoys, line 1:
```suggestion
//! Bindings to raw C++ LLaMA implementation. implements the TextGeneration trait and being used in tabby to generate text / code.
```

crates/tabby-scheduler/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+//! Responsible for scheduling all of the background tasks for tabby.
```
wsxiaoys, line 1:
```suggestion
//! Responsible for scheduling all of the background jobs for tabby.
```

used `jobs` since we have subcommand concept `job::` for backend jobs.

ee/tabby-webserver/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+//! Defines behavior for the tabby webserver which allows users to interact with enterprise features.
+//! Using a chat bot in the playground requires using this module with the `--webserver` flag on the command line.
```
wsxiaoys, line 2:
```suggestion
//! Using the web interface (e.g chat playground) requires using this module with the `--webserver` flag on the command line.
```

wsxiaoys (changes requested):

PR #1172, opened 2024-01-08
feat(agent): add lsp support to agent.

icycodes:
Add basic lsp support.
Complete TAB-388

Add command line arg `--lsp` will run agent as a lsp server.

Supported client -> server method:

- initialize
- shutdown
- exit
- textDocument/completion

Used server -> client method:
- window/showMessage
  - show warning when disconnected

### DIFF
--- a/clients/tabby-agent/README.md
+++ b/clients/tabby-agent/README.md
@@ -1,0 +1,80 @@
+# Tabby agent
+
+The Tabby Agent is an agent used for communication with the Tabby server. It is based on Node.js v18 and implements several common features to enhance code completion, including caching, debouncing, and post-processing.
+
+This package exports both libraries and a CLI tool.
+
+## Language Server (Unreleased)
+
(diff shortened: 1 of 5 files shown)


wsxiaoys:
could you compare this against https://github.com/huggingface/llm-ls briefly?

icycodes:
> could you compare this against https://github.com/huggingface/llm-ls briefly?

The main feature provided by both `tabby-agent` and `llm-ls` is the same, but implemented LSP methods may differ.
- `llm-ls` uses custom methods `llm-ls/getCompletions`, `llm-ls/acceptCompletions`, and `llm-ls/rejectCompletion` to provide code completion. This requires the use of a custom client like `llm-vscode` or `llm.nvim`.
- `tabby-agent` provides code completion through the standard `textDocument/completion` method. This allows it to be directly connected to widely-used LSP clients without the need for a custom client. But, since we haven't implemented `postEvent` in LSP mode, the `view/select` event telemetry feature is currently not available.

PR #1177, opened 2024-01-08
feat(webserver): Add graphql api for oauth credential management

darknight:

Added 2 GraphQL API:

- query: `oauthCredential`
- mutation: `resetOauthCredential`


ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -384,6 +400,19 @@ pub trait AuthenticationService: Send + Sync {
         code: String,
         client: Arc<GithubClient>,
     ) -> std::result::Result<GithubAuthResponse, GithubAuthError>;
+
+    async fn get_oauth_credential(
+        &self,
+        provider: OAuthProvider,
+    ) -> Result<Option<OAuthCredential>>;
```
wsxiaoys, line 407: do you think it make sense to not returns the actual `client_secret` in the read interface? The front end should just display a dummy masked value?

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -334,6 +334,22 @@ impl relay::NodeType for InvitationNext {
     }
 }
 
+#[derive(GraphQLEnum, Clone)]
+pub enum OAuthProvider {
+    Github,
+    Google,
```
wsxiaoys, line 340:
```suggestion
```

Remove google since it's not supported yet.

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -341,6 +358,34 @@ impl Mutation {
             "Only admin is able to delete invitation",
         ))
     }
+
+    /// `active` is optional, if omit, it'll be set as **true**
+    async fn reset_oauth_credential(
```
wsxiaoys, line 363:
```suggestion
    async fn update_oauth_credential(
```

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -341,6 +358,34 @@ impl Mutation {
...
+        ctx: &Context,
+        provider: OAuthProvider,
+        client_id: String,
+        client_secret: String,
+        active: Option<bool>,
```
wsxiaoys, line 368:
```suggestion
        active: bool,
```

ee/tabby-webserver/src/schema/auth.rs
wsxiaoys: I'l suggest just make client-secret Optional in read / update request.

wsxiaoys (changes requested):

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -31,20 +31,25 @@ impl DbConn {
     pub async fn update_github_oauth_credential(
         &self,
         client_id: &str,
-        client_secret: &str,
+        client_secret: &Option<String>,
```
wsxiaoys, line 34: can this be `Option<&str>` ?

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -240,6 +242,28 @@ impl Query {
             "Only admin is able to query job runs",
         )))
     }
+
+    async fn oauth_credential(
+        ctx: &Context,
+        provider: OAuthProvider,
+        include_secret: bool,
```
wsxiaoys, line 249:
```suggestion
```

Always do not return secret.

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -387,6 +387,32 @@ impl AuthenticationService for DbConn {
         };
         Ok(resp)
     }
+
+    async fn get_oauth_credential(
```
wsxiaoys, line 391:
```suggestion
    async fn read_oauth_credential(
```

wsxiaoys (changes requested):
Otherwise LGTM

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -31,20 +31,25 @@ impl DbConn {
...
+        if client_secret.is_some() {
+            sql += r#"SET client_id = :cid, client_secret = :secret, active = :active, updated_at = datetime('now')
+                    WHERE id = :id"#;
+        } else {
+            sql += r#"SET client_id = :cid, active = :active, updated_at = datetime('now')
+                    WHERE id = :id"#;
+        }
 
+        let client_id = client_id.to_string();
+        let client_secret = client_secret.unwrap_or_default().to_owned();
```
wsxiaoys, line 49: shall we avoid unwrap but restructure the condition in a different way?

darknight: what's different way, specifically?

wsxiaoys:
```rust
let (param, query) = if let Some(client_secret) = client_secret {
  // construct sql query and param object with client_secret and return.
} else {
  // construct sql query and param object without client_secret and return.
}

execute sql query and param...
```

darknight: As I tried, return `param` from `if let ... else ...` can not compile, it gives error:
```
temporary value dropped while borrowed
```
the `named_params!` macro generates value with type `&[(&str, &dyn ToSql)]`, which will be freed at the end of `if else` statement.

darknight: also, we need to do `INSERT INTO github_oauth_credential (id, client_id, client_secret)` for first write.
If sceret is not provided, we need the default empty string to be filled, as db schema requires client_secret is `NOT NULL`.

wsxiaoys: > the named_params! macro generates value with type &[(&str, &dyn ToSql)], which will be freed at the end of if else statement.

in that case, let's just write everything down, and try extract common structures between each branch.

> If sceret is not provided, we need the default empty string to be filled, as db schema requires client_secret is NOT NULL.

but when client_secret is empty and client_secret is empty, we should just probably just explicit returns an error?

darknight: > > the named_params! macro generates value with type &[(&str, &dyn ToSql)], which will be freed at the end of if else statement.
> 
> in that case, let's just write everything down, and try extract common structures between each branch.

I don't see much value here to extract `params` structure, either `client_secret` is some string or empty, we need these 4 parameters to execute the sql statement.

> > If sceret is not provided, we need the default empty string to be filled, as db schema requires client_secret is NOT NULL.
> 
> but when client_secret is empty and client_secret is empty, we should just probably just explicit returns an error?

Just like `register` and `tokenAuth` we can add input validation to return user-friendly error message for your concern, do you want to add it?

wsxiaoys (approved):

Issue #1143, opened 2023-12-29
Implement GraphQL api for repository configurations


wsxiaoys:
- [x] Implement schema change
- [x] Implement DAO
- [x] Implement GraphQL api

Example:
* DB Schema change: https://github.com/TabbyML/tabby/tree/main/ee/tabby-db/migrations/03-invitations-table
* DAO implementation: https://github.com/TabbyML/tabby/blob/main/ee/tabby-db/src/invitations.rs
* GraphQL API:
  - List: https://github.com/TabbyML/tabby/blob/main/ee/tabby-webserver/src/schema/mod.rs#L176
  - Create: https://github.com/TabbyML/tabby/blob/main/ee/tabby-webserver/src/schema/mod.rs#L306
  - Delete: https://github.com/TabbyML/tabby/blob/main/ee/tabby-webserver/src/schema/mod.rs#L334

PR #1178, opened 2024-01-08
feat(db): Implement DAO for repositories

boxbeam:
Related to #1143

### DIFF
--- a/ee/tabby-db/migrations/08-repositories-table/up.sql
+++ b/ee/tabby-db/migrations/08-repositories-table/up.sql
@@ -1,0 +1,4 @@
+CREATE TABLE repositories (
+    name VARCHAR(255) PRIMARY KEY,
+    git_url VARCHAR(255),
+);
--- a/ee/tabby-db/migrations/08-repositories-table/up.sql
+++ b/ee/tabby-db/migrations/08-repositories-table/up.sql
@@ -1,4 +1,7 @@
 CREATE TABLE repositories (
-    name VARCHAR(255) PRIMARY KEY,
-    git_url VARCHAR(255),
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    name VARCHAR(255) NOT NULL,
+    git_url VARCHAR(255) NOT NULL,
+    CONSTRAINT `idx_name` UNIQUE (`name`)
+    CONSTRAINT `idx_git_url` UNIQUE (`git_url`)
 );
--- a/ee/tabby-db/src/repositories.rs
+++ b/ee/tabby-db/src/repositories.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use rusqlite::{OptionalExtension, Row};
 
-use crate::DbConn;
+use crate::{DbConn, TabbyDBError};
 
 pub struct RepositoryDAO {
     pub id: i32,
@@ -83,17 +83,22 @@ impl DbConn {
             .await?)
     }
 
-    pub async fn update_repository_url(&self, id: i32, git_url: String) -> Result<bool> {
-        Ok(self
+    pub async fn update_repository(&self, id: i32, name: String, git_url: String) -> Result<()> {
+        let updated = self
             .conn
             .call(move |c| {
-                let updated = c.execute(
-                    "UPDATE repositories SET git_url=? WHERE id=?",
-                    (git_url, id),
+                let update_count = c.execute(
+                    "UPDATE repositories SET git_url=?, name=? WHERE id=?",
+                    (git_url, name, id),
                 )?;
-                Ok(updated == 1)
+                Ok(update_count == 1)
             })
-            .await?)
+            .await?;
+        if updated {
+            Ok(())
+        } else {
+            Err(TabbyDBError::NoSuchElement.into())
+        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


ee/tabby-db/migrations/08-repositories-table/up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE repositories (
```
wsxiaoys, line 1: Again, please follow example in issue to define the table:
https://github.com/TabbyML/tabby/issues/1143

ee/tabby-db/src/repositories.rs
```diff
@@ -0,0 +1,107 @@
...
+                    .prepare("SELECT name, git_url FROM repositories")?
+                    .query_map([], RepositoryDAO::from_row)?
+                    .map(|r| r.map_err(Into::into))
+                    .collect();
+                Ok(thing)
+            })
+            .await?
+    }
+
+    pub async fn get_repository(&self, name: String) -> Result<Option<RepositoryDAO>> {
```
wsxiaoys, line 35: Single item CRUD should be operating on id

ee/tabby-db/migrations/08-repositories-table/up.sql
```diff
@@ -0,0 +1,4 @@
...
+    name VARCHAR(255) PRIMARY KEY,
+    git_url VARCHAR(255),
```
wsxiaoys, line 3: Add unique constraint to `git_url`

wsxiaoys (changes requested):

ee/tabby-db/src/repositories.rs
```diff
@@ -68,13 +83,13 @@ impl DbConn {
             .await?)
     }
 
-    pub async fn update_repository_url(&self, name: String, git_url: String) -> Result<bool> {
+    pub async fn update_repository_url(&self, id: i32, git_url: String) -> Result<bool> {
```
wsxiaoys, line 86:
```suggestion
    pub async fn update_repository(&self, id: i32, name: String, git_url: String) -> Result<()> {
```

ee/tabby-db/src/repositories.rs
```diff
@@ -68,13 +83,13 @@ impl DbConn {
...
             .conn
-            .call(|c| {
+            .call(move |c| {
                 let updated = c.execute(
-                    "UPDATE repositories SET git_url=? WHERE name=?",
-                    [git_url, name],
+                    "UPDATE repositories SET git_url=? WHERE id=?",
+                    (git_url, id),
                 )?;
                 Ok(updated == 1)
```
wsxiaoys, line 94: When updated is not 1, shall we generate an error? or rusqlite already generate an error for us?

wsxiaoys (changes requested):

wsxiaoys:
When this PR is ready for review, please remove the draft status

ee/tabby-db/src/repositories.rs
boxbeam: Do we want to have one function that updates both the name and URL? Or one for each? This one was intended to be just for updating the URL. Also, this function returns a boolean to indicate whether the value was successfully updated - do we want to discard that information?

ee/tabby-db/src/repositories.rs
boxbeam: SQL won't error if an update query affects no rows. I wasn't sure if that should be an error, do you think it would be preferable to returning false?

ee/tabby-db/src/repositories.rs
wsxiaoys: for DAO functions, i suggest by default just provide an update function with all necessary parameters. We can always add new functions in future.

> Also, this function returns a boolean to indicate whether the value was successfully updated - do we want to discard that information?

See another comments, return error when updated != 1

ee/tabby-db/migrations/08-repositories-table/up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE repositories (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    name VARCHAR(255) NOT NULL,
+    git_url VARCHAR(255) NOT NULL,
```
wsxiaoys, line 4: let's still add `created_at` just for sorting.

ee/tabby-db/src/repositories.rs
```diff
@@ -0,0 +1,137 @@
...
+            .call(move |c| {
+                Ok(c.query_row(
+                    "SELECT name, git_url FROM repositories WHERE name=?",
+                    [name],
+                    RepositoryDAO::from_row,
+                ))
+            })
+            .await?
+            .optional()?)
+    }
```
wsxiaoys, line 62:
```suggestion
```

ee/tabby-db/src/repositories.rs
```diff
@@ -0,0 +1,137 @@
...
+            .call(move |c| {
+                Ok(c.query_row(
+                    "SELECT name, git_url FROM repositories WHERE id=?",
+                    [id],
+                    RepositoryDAO::from_row,
+                ))
+            })
+            .await?
+            .optional()?)
+    }
```
wsxiaoys, line 48:
```suggestion
```

ee/tabby-db/src/repositories.rs
```diff
@@ -0,0 +1,137 @@
...
+            .call(move |c| {
+                let deleted = c
+                    .execute("DELETE FROM repositories WHERE id=?", [id])
+                    .is_ok();
+                Ok(deleted)
+            })
+            .await?)
+    }
+
+    pub async fn add_repository(&self, name: String, git_url: String) -> Result<()> {
```
wsxiaoys, line 76:
```suggestion
    pub async fn create_repository(&self, name: String, git_url: String) -> Result<()> {
```

ee/tabby-db/migrations/08-repositories-table/up.sql
boxbeam: If it's just for sorting, wouldn't sorting by ID still order it the same?

wsxiaoys:
Please fix unit test

PR #1193, opened 2024-01-11
feat: Remove 'active' field from OAuth credentials

boxbeam:
Removes the active field and determines active status based on whether the credential is present or not

### DIFF
--- a/ee/tabby-db/src/github_oauth_credential.rs
+++ b/ee/tabby-db/src/github_oauth_credential.rs
@@ -133,6 +133,10 @@ mod tests {
         assert_eq!(res.client_id, "client_id");
         assert_eq!(res.client_secret, "client_secret_2");
 
+        // test delete
+        conn.delete_github_oauth_credential().await.unwrap();
+        assert!(conn.read_github_oauth_credential().await.unwrap().is_none());
+
         // test update without client_secret
         conn.update_github_oauth_credential("client_id_2", None)
             .await
--- a/ee/tabby-db/src/github_oauth_credential.rs
+++ b/ee/tabby-db/src/github_oauth_credential.rs
@@ -29,49 +29,26 @@ impl DbConn {
     pub async fn update_github_oauth_credential(
         &self,
         client_id: &str,
-        client_secret: Option<&str>,
+        client_secret: &str,
     ) -> Result<()> {
         let client_id = client_id.to_string();
-        if let Some(client_secret) = client_secret {
-            let client_secret = client_secret.to_string();
-            let sql = r#"INSERT INTO github_oauth_credential (id, client_id, client_secret)
+        let client_secret = client_secret.to_string();
+        let sql = r#"INSERT INTO github_oauth_credential (id, client_id, client_secret)
                                 VALUES (:id, :cid, :secret) ON CONFLICT(id) DO UPDATE
                                 SET client_id = :cid, client_secret = :secret, updated_at = datetime('now')
                                 WHERE id = :id"#;
-            self.conn
-                .call(move |c| {
-                    let mut stmt = c.prepare(sql)?;
-                    stmt.insert(named_params! {
-                    ":id": GITHUB_OAUTH_CREDENTIAL_ROW_ID,
-                    ":cid": client_id,
-                    ":secret": client_secret,
-                    })?;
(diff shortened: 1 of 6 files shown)


wsxiaoys:
add @darknight for review

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -410,6 +409,19 @@ impl Mutation {
             .await?;
         Ok(true)
     }
+
+    async fn delete_email_service_credential(ctx: &Context) -> FieldResult<bool> {
```
darknight, line 413: remove unrelated method

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -410,6 +409,19 @@ impl Mutation {
...
+        ctx.locator
+            .email_service_credential()
+            .delete_email_service_credential()
+            .await?;
+        Ok(true)
+    }
+
+    async fn delete_oauth_credential(ctx: &Context, provider: OAuthProvider) -> FieldResult<bool> {
```
darknight, line 421: Use defined `type Result<T, E = CoreError>` if there's no particular reason.

ee/tabby-webserver/src/schema/mod.rs
wsxiaoys: it's related (delete means deactive)

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -48,15 +45,14 @@ impl DbConn {
...
-                    ":active": active,
                     })?;
                     Ok(())
                 })
                 .await?;
             Ok(())
         } else {
             let sql = r#"
-            UPDATE github_oauth_credential SET client_id = :cid, active = :active, updated_at = datetime('now')
+            UPDATE github_oauth_credential SET client_id = :cid, updated_at = datetime('now')
```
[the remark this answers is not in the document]
wsxiaoys: i think it's still necessary to support Option<&str> - for example, an user might update smtp server url / port / protocol without updating client secret.

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -363,7 +363,7 @@ impl Mutation {
         provider: OAuthProvider,
         client_id: String,
         client_secret: Option<String>,
-        active: bool,
+        _active: bool,
```
wsxiaoys, line 366: remove this

wsxiaoys (changes requested):

ee/tabby-db/src/github_oauth_credential.rs
```diff
@@ -79,13 +74,26 @@ impl DbConn {
         }
     }
 
+    pub async fn delete_github_oauth_credential(&self) -> Result<()> {
```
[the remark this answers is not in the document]
boxbeam: Added a test case

ee/tabby-db/src/github_oauth_credential.rs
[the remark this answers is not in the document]
darknight: `smtp`? so we want to manage `smtp credential` & `oauth credential` together?

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -403,15 +400,20 @@ impl AuthenticationService for DbConn {
         &self,
         provider: OAuthProvider,
         client_id: String,
-        client_secret: Option<String>,
-        active: bool,
+        client_secret: String,
     ) -> Result<()> {
         match provider {
             OAuthProvider::Github => Ok(self
-                .update_github_oauth_credential(&client_id, client_secret.as_deref(), active)
```
darknight, line 411: overall looks good to me. but you may need to revert this change if we still need to pass `Option<&str>`.

wsxiaoys: for this specific case (github oauth), i think we don't really have a case that need to update client_id only, without updating the client_secret. thus I think it's fine to make client_secret a required field in updating, after removing active.

darknight: sounds good, no more comments from my side.

PR #1194, opened 2024-01-11
feat(webserver): Add Google oauth support

darknight:

Similar to Github oauth, add following changes

- Add `google_oauth_credential` db table, and read/write operations
  - it's slightly different from `github_oauth_credential`, we need to keep `redirect_uri` because it's a `required` parameter when exchange access token (unlike github which is optional)
- Add `google` client, implement token exhange and user email fetching
- Add `/oauth_callback/google` route, extract common response handling logic into `match_auth_result` method
- google oauth is very simliar to github oauth, so refactor `github_auth` method to take care of both providers
- update graphql api to update/return `redirect_uri` field, this field is defined as optional, so should break nothing

### DIFF
--- a/ee/tabby-db/src/google_oauth_credential.rs
+++ b/ee/tabby-db/src/google_oauth_credential.rs
@@ -1,0 +1,172 @@
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use rusqlite::{named_params, OptionalExtension};
+
+use super::DbConn;
+
+const GOOGLE_OAUTH_CREDENTIAL_ROW_ID: i32 = 1;
+
+pub struct GoogleOAuthCredentialDAO {
+    pub client_id: String,
+    pub client_secret: String,
+    pub redirect_uri: String,
+    pub created_at: DateTime<Utc>,
+    pub updated_at: DateTime<Utc>,
+}
+
+impl GoogleOAuthCredentialDAO {
+    fn from_row(row: &rusqlite::Row<'_>) -> std::result::Result<Self, rusqlite::Error> {
+        Ok(Self {
+            client_id: row.get(0)?,
+            client_secret: row.get(1)?,
+            redirect_uri: row.get(2)?,
+            created_at: row.get(3)?,
+            updated_at: row.get(4)?,
+        })
+    }
+}
+
+/// db read/write operations for `google_oauth_credential` table
+impl DbConn {
+    pub async fn update_google_oauth_credential(
+        &self,
+        client_id: &str,
+        client_secret: &str,
+        redirect_uri: Option<&str>,
+    ) -> Result<()> {
+        let redirect_uri = redirect_uri.unwrap_or_default().to_string();
+        if !client_secret.is_empty() {
+            let client_id = client_id.to_string();
+            let client_secret = client_secret.to_string();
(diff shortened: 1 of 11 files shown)


darknight:
blocked by https://github.com/TabbyML/tabby/pull/1193

ee/tabby-db/src/google_oauth_credential.rs
```diff
@@ -0,0 +1,172 @@
...
+/// db read/write operations for `google_oauth_credential` table
+impl DbConn {
+    pub async fn update_google_oauth_credential(
+        &self,
+        client_id: &str,
+        client_secret: &str,
+        redirect_uri: Option<&str>,
+    ) -> Result<()> {
+        let redirect_uri = redirect_uri.unwrap_or_default().to_string();
+        if !client_secret.is_empty() {
```
boxbeam, line 38: Not a big fan of this pattern, since it really balloons if we ever have more than one optional field. Maybe you could use a `format!` to construct the query, having it do `SET client_secret = :client_secret` when the secret is provided, and `SET client_secret = client_secret` for a no-op when it is not provided? That way just a tiny piece of data would be conditional, and the rest of the logic would have no branching and no additional queries.

ee/tabby-webserver/src/oauth/mod.rs
```diff
@@ -48,23 +58,54 @@ async fn github_callback(
...
+        state
+            .auth
+            .oauth(param.code, OAuthClient::Github(state.github_client.clone()))
+            .await,
+    )
+}
+
+#[derive(Deserialize)]
+#[allow(dead_code)]
+struct GoogleCallbackParam {
```
boxbeam, line 71: Purpose unclear, maybe leave a comment explaining what this is for?

ee/tabby-webserver/src/oauth/mod.rs
```diff
@@ -48,23 +58,54 @@ async fn github_callback(
...
+    #[serde(default)]
+    code: String,
+    #[serde(default)]
+    scope: String,
+    #[serde(default)]
+    error: String,
+}
+
+async fn google_callback(
```
boxbeam, line 80: Also unclear, this function could use a more descriptive name and/or comment

boxbeam (changes requested):

ee/tabby-db/src/google_oauth_credential.rs
darknight: 1. be aware that we never return `client_secret` to FE, so if passed secret is non-empty, very likely it's (re)set by the admin. So it goes to the `if` branch.
2. multi optional fields are fine, they all go to the `else` branch if the admin just wants to update them, e.g. `redirect_url` for google oauth
3. about your suggestion, I found there is one case that we might insert a record `(1, '', '', '')` because the schema requires `NOT NULL` for id, secret, redirect_url. So the insert statement is fixed as `insert into google_oauth_credential (id,client_id,client_secret,redirect_uri) values (...)`, we can only tune the `do update part`. But for current implementation, if secret is empty and no existing record, the `update` in else branch will fail as expected.

boxbeam: Sounds good, then. Just wish there was a better way to do this.

boxbeam (approved):

PR #1198, opened 2024-01-12
feat(core): improve completion error handling

### LINKED ISSUE #1108: Improve error handling in /v1beta/chat/completions
Instead of panic when input messages is invalid in chat routes, we should generate better response code (e.g 422) to reflect error status.

Code location:
* https://github.com/TabbyML/tabby/blob/main/crates/tabby/src/services/chat/chat_prompt.rs#L26
* https://github.com/TabbyML/tabby/blob/main/crates/tabby/src/services/chat.rs#L61

We should propagate the error to chat completion routes and return 422 code.

boxbeam:
Closes #1108

### DIFF
--- a/crates/tabby/src/routes/chat.rs
+++ b/crates/tabby/src/routes/chat.rs
@@ -31,11 +31,11 @@ pub async fn chat_completions(
     Json(request): Json<ChatCompletionRequest>,
 ) -> Response {
     let s = stream! {
-        for await content in state.generate(&request).await {
+        for await content in state.generate(&request).await? {
             let content = match serde_json::to_string(&content) {
                 Ok(s) => s,
                 Err(e) => {
-                    yield Err(e);
+                    yield Err::<_, anyhow::Error>(e.into());
                     continue
                 }
             };
--- a/crates/tabby/src/routes/chat.rs
+++ b/crates/tabby/src/routes/chat.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use async_stream::stream;
+
 use axum::{
     body::StreamBody,
     extract::State,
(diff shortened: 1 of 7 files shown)


crates/tabby/src/services/chat/chat_prompt.rs
```diff
@@ -16,14 +16,10 @@ impl ChatPromptBuilder {
...
 
-    pub fn build(&self, messages: &[Message]) -> String {
-        self.env
-            .get_template("prompt")
-            .unwrap()
-            .render(context!(
-                    messages => messages
-            ))
-            .expect("Failed to evaluate")
+    pub fn build(&self, messages: &[Message]) -> Result<String, minijinja::Error> {
```
wsxiaoys, line 19: define custom error type with thiserror https://docs.rs/thiserror/latest/thiserror/

wsxiaoys (changes requested):

crates/tabby/src/services/error.rs
```diff
@@ -0,0 +1,7 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CompletionError {
```
wsxiaoys, line 4: Move this to `Chat` subcrate? It's part of interface of the chat service.

crates/tabby/src/routes/chat.rs
```diff
@@ -31,17 +31,10 @@ pub async fn chat_completions(
...
-        for await content in state.generate(&request).await {
-            let content = match serde_json::to_string(&content) {
-                Ok(s) => s,
-                Err(e) => {
-                    yield Err(e);
-                    continue
-                }
-            };
-
+        for await content in state.generate(&request).await? {
```
wsxiaoys, line 34: We should handle the error type in state.generate, with a match clause.

e.g for Minijinja error, we'd like to returns something like https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

You'll also need to update the status code responses in line 22.

PR #1207, opened 2024-01-15
feat: Database refactor demonstration

boxbeam:
Refactor the repository service to use sqlx instead of rusqlite as an example, to see if we want to move away from rusqlite


.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=sqlite:///home/redempt/.tabby/ee/db.sqlite
```
boxbeam, line 1: This is temporary. We will need to either:

1. Have people update this file with their local path to the tabby DB
2. Use the build.rs to automatically generate this file with the correct path
3. Include a SQLite db in the tabby repository storing only the schema

This is for sqlx's checked queries.

ee/tabby-db/src/repositories.rs
```diff
@@ -1,14 +1,94 @@
...
+    ) -> Result<Vec<RepositoryDAO>> {
+        let query = DbConn::make_pagination_query(
+            "repositories",
+            &["id", "name", "git_url"],
+            limit,
+            skip_id,
+            backwards,
+        );
+        let rows = sqlx::query(&query)
+            .try_map(|r| FromRow::from_row(&r))
```
boxbeam, line 73: Using the `sqlx-macros` crate, we can derive the `FromRow` trait, which prevents us from having to manually implement our `from_row` deserialization method on each DAO type. Our `from_row` implementation is currently error prone since it uses positional row getters to fetch data. Tests should catch failures, but I believe sqlx uses column names to deserialize and so should be much safer while also requiring less code.

.env
boxbeam, line 1: Currently I am keeping the sqlite database inside the repository, and we can update it as we modify its schema. Another option would be to generate the database automatically using our migration files.

PR #1222, opened 2024-01-17
fix(webserver): filter out .git dir for resolve api

darknight:


ee/tabby-webserver/src/repositories/resolve.rs
```diff
@@ -160,6 +160,10 @@ pub async fn resolve_dir(repo_name: &str, root: PathBuf, full_path: PathBuf) ->
             // Skip others.
             continue;
         };
+        // filter out .git directory
+        if kind == DirEntryKind::Dir && basename == ".git" {
```
wsxiaoys, line 164: should we filter out all hidden files?

wsxiaoys, line 164: should we only filter out `.git` directory at root?

wsxiaoys:
cc @liangfung

PR #1229, opened 2024-01-17
feat(core): Add support for passing temperature as a parameter

### LINKED ISSUE #1216: [TAB-433] Add seed / temperature (i64) parameter in code completion request
So the client side have the chance to control randomness and generating a different completion.

Code location:

Request: [https://github.com/TabbyML/tabby/blob/3fdfe652f98b80b5631f45f34cd20ead69111f09/crates/tabby/src/services/completion.rs#L36](https://github.com/TabbyML/tabby/blob/3fdfe652f98b80b5631f45f34cd20ead69111f09/crates/tabby/src/services/completion.rs#L36)

Sampling:
[https://github.com/TabbyML/tabby/blob/3fdfe652f98b80b5631f45f34cd20ead69111f09/crates/llama-cpp-bindings/src/engine.cc#L217](https://github.com/TabbyML/tabby/blob/3fdfe652f98b80b5631f45f34cd20ead69111f09/crates/llama-cpp-bindings/src/engine.cc#L217)

Instead of selecting token with max probs, we need to apply temperature based sampling (with seed): [https://lukesalamone.github.io/posts/what-is-temperature/](https://lukesalamone.github.io/posts/what-is-temperature/)

Might consider reuse [https://github.com/ggerganov/llama.cpp/blob/c37b3474e61d609d43cccc3bde5d559e80e4f5d1/llama.h#L734](https://github.com/ggerganov/llama.cpp/blob/c37b3474e61d609d43cccc3bde5d559e80e4f5d1/llama.h#L734)

<sub>From [SyncLinear.com](https://synclinear.com) | [TAB-433](https://linear.app/tabbyml/issue/TAB-433/add-seed-temperature-i64-parameter-in-code-completion-request-1211)</sub>

boxbeam:
Closes #1216

### DIFF
--- a/crates/llama-cpp-bindings/src/engine.cc
+++ b/crates/llama-cpp-bindings/src/engine.cc
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <functional>
+#include <random>
 #include <vector>
 #include <cmath>
 #include <deque>
@@ -97,13 +98,13 @@ float softmax(float* nums, size_t len, float_t temperature) {
   return sum;
 }
 
-size_t weighted_random(float* nums, size_t len) {
+size_t weighted_random(float* nums, size_t len, std::mt19937 rng) {
   float_t sum = 0;
   for (size_t i = 0; i < len; i++) {
     sum += nums[i];
   }
 
-  float random = static_cast<float> (rand()) / static_cast<float>(RAND_MAX);
+  float random = std::uniform_real_distribution<float>(0, sum)(rng);
   random *= sum;
   sum = 0;
   for (size_t i = 0; i < len; i++) {
@@ -249,12 +250,12 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
         if ((request.i_batch < i) || (request.i_batch >= (i + n_tokens))) {
           continue;
         }
-        srand(request.seed);
+        std::mt19937 rng(request.seed);
 
         int32_t i_batch = request.i_batch - i;
         float* logits = llama_get_logits_ith(ctx, i_batch);
         softmax(logits, n_vocab, request.temperature);
-        auto next_token = weighted_random(logits, n_vocab);
+        auto next_token = weighted_random(logits, n_vocab, rng);
         request.n_past += request.tokens.size();
 
         request.tokens.clear();
--- a/crates/llama-cpp-bindings/src/engine.cc
+++ b/crates/llama-cpp-bindings/src/engine.cc
@@ -101,7 +101,6 @@ size_t weighted_random(float* nums, size_t len) {
     sum += nums[i];
   }
 
-  srand(time(NULL));
   float random = static_cast<float> (rand()) / static_cast<float>(RAND_MAX);
   random *= sum;
   sum = 0;
@@ -126,6 +125,7 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
       batch_ = llama_batch_init(N_CTX * parallelism, 0, 1);
       // warm up
       {
+        srand(time(NULL));
         batch_.n_tokens = 16;
         for (int i = 0; i < batch_.n_tokens; ++i) {
           batch_.token[i] = 0;
--- a/crates/llama-cpp-bindings/src/engine.cc
+++ b/crates/llama-cpp-bindings/src/engine.cc
@@ -19,15 +19,17 @@ namespace {
 constexpr size_t N_BATCH = 512;  // # per batch inference.
 constexpr size_t N_CTX = 4096;   // # max kv history.
 struct Request {
-  Request(size_t request_id, std::vector<llama_token> input_token_ids, float_t temperature) :
+  Request(size_t request_id, std::vector<llama_token> input_token_ids, float_t temperature, uint64_t seed) :
     id(request_id),
     tokens(input_token_ids.begin(), input_token_ids.end()),
-    temperature(temperature) {
+    temperature(temperature),
+    seed(seed) {
     }
 
   uint32_t id = -1;
   llama_seq_id seq_id = -1;
   float_t temperature = 0;
+  uint64_t seed = 0;
 
   std::vector<llama_token> tokens;
   size_t i_batch = -1;
@@ -146,13 +148,13 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
     llama_batch_free(batch_);
   }
 
-  virtual void add_request(uint32_t request_id, rust::Str text, size_t max_input_length, float_t temperature) override {
+  virtual void add_request(uint32_t request_id, rust::Str text, size_t max_input_length, float_t temperature, uint64_t seed) override {
     auto tokens = llama_tokenize(llama_get_model(ctx_.get()), text, false, true);
     if (tokens.size() > max_input_length) {
       int start = tokens.size() - max_input_length;
       tokens = std::vector<llama_token>(tokens.begin() + start, tokens.end());
     }
-    pending_requests_.push_back(Request(request_id, tokens, temperature));
+    pending_requests_.push_back(Request(request_id, tokens, temperature, seed));
   }
 
   void stop_request(uint32_t request_id) override {
@@ -247,6 +249,7 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
         if ((request.i_batch < i) || (request.i_batch >= (i + n_tokens))) {
           continue;
         }
+        srand(request.seed);
 
         int32_t i_batch = request.i_batch - i;
         float* logits = llama_get_logits_ith(ctx, i_batch);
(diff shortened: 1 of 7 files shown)


crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -110,13 +148,13 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
     llama_batch_free(batch_);
   }
 
-  virtual void add_request(uint32_t request_id, rust::Str text, size_t max_input_length) override {
+  virtual void add_request(uint32_t request_id, rust::Str text, size_t max_input_length, float_t temperature, uint64_t seed) override {
```
wsxiaoys, line 151: breaking line to make it easier to read

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -211,11 +249,12 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
         if ((request.i_batch < i) || (request.i_batch >= (i + n_tokens))) {
           continue;
         }
+        srand(request.seed);
 
         int32_t i_batch = request.i_batch - i;
-        auto logits = llama_get_logits_ith(ctx, i_batch);
-        auto next_token = std::distance(logits, std::max_element(logits, logits + n_vocab));
-
+        float* logits = llama_get_logits_ith(ctx, i_batch);
```
wsxiaoys, line 255:
```suggestion
        const float* logits = llama_get_logits_ith(ctx, i_batch);
```

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -77,6 +83,38 @@ std::string string_format(const std::string& format, Args ... args)
 	return std::string(buf.get(), buf.get() + size - 1); // We don't want the '\0' inside
 }
 
+float softmax(float* nums, size_t len, float_t temperature) {
```
wsxiaoys, line 86:
```suggestion
float softmax(const float* nums, size_t len, float temperature) {
```

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -77,6 +83,38 @@ std::string string_format(const std::string& format, Args ... args)
...
+    nums[i] = std::exp(nums[i] / temperature);
+    sum += nums[i];
+  }
+  for (size_t i = 0; i < len; i++) {
+      nums[i] /= sum;
+  }
+  return sum;
+}
+
+size_t weighted_random(float* nums, size_t len) {
```
wsxiaoys, line 100: ditto

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -211,11 +249,12 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
         if ((request.i_batch < i) || (request.i_batch >= (i + n_tokens))) {
           continue;
         }
+        srand(request.seed);
```
wsxiaoys, line 252: srand itself is not thread safe, let's use std::mt19937 to create a random generator in stack.

Example: https://github.com/ggerganov/llama.cpp/blob/b46757735d30f5c6ed4f20ebeccc684e02d4f3bf/examples/main/main.cpp#L171

wsxiaoys (changes requested):

wsxiaoys:
Please always document the component your pr changed in semantic PR, in this case, it should be 
```
feat(core): ...
```

crates/llama-cpp-bindings/src/engine.cc
boxbeam: I can't make these const because my `softmax` function modifies the array in-place rather than allocating a new one. Would you rather I allocate one?

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -211,11 +252,12 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
...
         }
+        std::mt19937 rng(request.seed);
 
         int32_t i_batch = request.i_batch - i;
-        auto logits = llama_get_logits_ith(ctx, i_batch);
-        auto next_token = std::distance(logits, std::max_element(logits, logits + n_vocab));
-
+        float* logits = llama_get_logits_ith(ctx, i_batch);
+        softmax(logits, n_vocab, request.temperature);
+        auto next_token = weighted_random(logits, n_vocab, rng);
```
wsxiaoys, line 260: pass request.seed as arguments

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -77,6 +84,38 @@ std::string string_format(const std::string& format, Args ... args)
 	return std::string(buf.get(), buf.get() + size - 1); // We don't want the '\0' inside
 }
 
+float softmax(float* nums, size_t len, float temperature) {
```
wsxiaoys, line 87:
```suggestion
float compute_softmax_inplace(float* nums, size_t len, float temperature) {
```

crates/tabby/src/services/completion.rs
```diff
@@ -216,7 +227,16 @@ impl CompletionService {
     ) -> Result<CompletionResponse, CompletionError> {
         let completion_id = format!("cmpl-{}", uuid::Uuid::new_v4());
         let language = request.language_or_unknown();
-        let options = Self::text_generation_options(language.as_str());
+        let options = Self::text_generation_options(
+            language.as_str(),
+            request.temperature.unwrap_or(0.1),
+            request.seed.unwrap_or_else(|| {
```
wsxiaoys, line 233: could you also add this seed setting logic to `/chat` api? (probably add temperature to that path as well)

boxbeam: Could that be done in a separate PR?

wsxiaoys: sounds good

boxbeam: #1238

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -77,6 +84,39 @@ std::string string_format(const std::string& format, Args ... args)
...
+  }
+  for (size_t i = 0; i < len; i++) {
+      nums[i] /= sum;
+  }
+  return sum;
+}
+
+size_t weighted_random(const float* nums, size_t len, uint64_t seed) {
+  std::mt19937 rng(seed);
+  float sum = 0;
```
wsxiaoys, line 103: doesn't softmax above guranteed the `sum` being computed here ends up = `1.0`?

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -77,6 +84,39 @@ std::string string_format(const std::string& format, Args ... args)
...
+  for (size_t i = 0; i < len; i++) {
+    sum += nums[i];
+  }
+
+  float random = std::uniform_real_distribution<float>(0, sum)(rng);
+  random *= sum;
```
wsxiaoys, line 109: why need to `*= sum` here?

boxbeam: We don't, good catch.

crates/llama-cpp-bindings/src/engine.cc
wsxiaoys: make sense, while since we compute the sum in `compute_softmax_in_place`, let's just pass it here to avoid recomputing?

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -77,6 +84,39 @@ std::string string_format(const std::string& format, Args ... args)
...
+  float random = std::uniform_real_distribution<float>(0, sum)(rng);
+  sum = 0;
+  size_t i;
+  for (i = 0; i < len; i++) {
+    sum += nums[i];
+    if (sum >= random) {
+      return i;
+    }
+  }
+  return i;
```
wsxiaoys, line 117:
```suggestion
  size_t i;
  for (i = 0; i < len; i++) {
    random -= nums[i];
    if (random < 0) {
      return i;
    }
  }
  return i;
```

wsxiaoys (changes requested):
Please have this revisited in after the golden test change for cpu.

crates/tabby/src/services/completion.rs
wsxiaoys, line 1: Please also add temperature / seed to `chat_completion`, adopting openai's interface.

boxbeam: We had agreed to do this in a separate PR

wsxiaoys: ah - sorry I forgot that discussion. True, it can be done in a separate PR.

wsxiaoys:
cc @icycodes

PR #1235, opened 2024-01-18
fix(core): Make should-panic test use unwrap instead of ignoring result

boxbeam:


wsxiaoys:
please add component in semantic message, in this case, it should be
`fix(core)`

PR #1244, opened 2024-01-19
fix(agent): Use /usr/bin/env instead of /bin/env

anoldguy:
macOS does not have `/bin/env`, so let's point to `/usr/bin/env` like the other instances.

### DIFF
--- a/clients/intellij/node_scripts/tabby-agent.js
+++ b/clients/intellij/node_scripts/tabby-agent.js
@@ -1,4 +1,4 @@
-#!/bin/env node
+#!/usr/bin/env node
 'use strict';
 
 var events = require('events');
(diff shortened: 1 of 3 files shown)


anoldguy:
Thanks @boxbeam / @icycodes! This is my first interaction with this codebase, so please let me know if I need to change anything, I'll be happy to.

boxbeam:
Please rename your commit via `git commit --amend` to: `fix(agent): Use /usr/bin/env instead of /bin/env` and force-push, then you should be good to mark this ready for review. Thanks for the PR!

icycodes (approved):

anoldguy:
@icycodes Will there be a new npm version bumped for this? I didn't know whether to put that in this PR or not and how npm releases were managed.

icycodes:
> @icycodes Will there be a new npm version bumped for this? I didn't know whether to put that in this PR or not and how npm releases were managed.

Hi @anoldguy, 
I apologize for the delay in my response. I am releasing a version later today. 
Thank you for your contribution!

PR #1249, opened 2024-01-19
fix(ui, tags): add utf16 column range for ease of tag rendering in javascript

wsxiaoys:
related discussion: https://github.com/tree-sitter/tree-sitter/issues/792
related test case: https://github.com/tree-sitter/tree-sitter/blob/660481dbf71413eba5a928b0b0ab8da50c1109e0/cli/src/tests/tags_test.rs#L186


wsxiaoys:
cc @liangfung please give this PR a try to see this `utf16_column_range` could help fix the issue we're encountering

liangfung (approved):

PR #1250, opened 2024-01-19
test(tabby): Fix golden tests and add golden tests for CPU

### LINKED ISSUE #1248: [TAB-448] add golden test (chat + completion) for cpu
<sub>From [SyncLinear.com](https://synclinear.com) | [TAB-448](https://linear.app/tabbyml/issue/TAB-448/add-golden-test-chat-completion-for-rocm)</sub>

boxbeam:
Closes #1248

Our golden test crate, `insta`, seems to use the expression inside the macro to uniquely identify the test. We were using a function that invokes this macro, which doesn't work because the expression inside the macro is always the same. I have converted `assert_golden` from a function to a macro to fix this, and added a test for CPU.

Because of our prior golden tests being incorrect, they all had to be deleted and will need to be regenerated.


crates/tabby/tests/goldentests.rs
```diff
@@ -16,9 +15,12 @@ lazy_static! {
             .arg("TabbyML/StarCoder-1B")
             .arg("--port")
             .arg("9090")
-            .arg("--device")
-            .arg("metal")
             .kill_on_drop(true);
```
wsxiaoys, line 18: refactor above block into a function, instead of a static variable.

Thus we will have freedom to pass in different device enum in different test.

crates/tabby/tests/goldentests.rs
```diff
@@ -84,29 +86,51 @@ async fn golden_test(body: serde_json::Value) -> serde_json::Value {
...
             "language": "python",
             "segments": {
                 "prefix": "import datetime\n\ndef parse_expenses(expenses_string):\n    \"\"\"Parse the list of expenses and return the list of triples (date, value, currency).\n    Ignore lines starting with #.\n    Parse the date using datetime.\n    Example expenses_string:\n        2016-01-02 -34.01 USD\n        2016-01-03 2.59 DKK\n        2016-01-03 -2.72 EUR\n    \"\"\"\n    for line in expenses_string.split('\\n'):\n        "
             }
-    })).await;
+    }));
+}
+
+#[cfg(not(all(target_os = "macos", target_arch = "aarch64")))]
+#[tokio::test]
```
wsxiaoys, line 119:
```suggestion
```

run cpu test on ios as well

crates/tabby/tests/goldentests_chat.rs
```diff
@@ -103,31 +106,32 @@ async fn golden_test(body: serde_json::Value) -> String {
...
+    assert_golden!(json!({
             "messages": [
                 {
                     "role": "user",
                     "content": "How to parse email address with regex"
                 }
             ]
-    }))
-    .await;
+    }));
```
wsxiaoys, line 136: add test for cpu

wsxiaoys (changes requested):

crates/tabby/tests/goldentests.rs
```diff
@@ -84,29 +86,50 @@ async fn golden_test(body: serde_json::Value) -> serde_json::Value {
...
             "segments": {
                 "prefix": "import datetime\n\ndef parse_expenses(expenses_string):\n    \"\"\"Parse the list of expenses and return the list of triples (date, value, currency).\n    Ignore lines starting with #.\n    Parse the date using datetime.\n    Example expenses_string:\n        2016-01-02 -34.01 USD\n        2016-01-03 2.59 DKK\n        2016-01-03 -2.72 EUR\n    \"\"\"\n    for line in expenses_string.split('\\n'):\n        "
             }
-    })).await;
+    }));
+}
+
+#[tokio::test]
+async fn run_golden_tests_cpu() {
+    wait_for_server(None).await;
```
wsxiaoys, line 120: let's specify this explicitly to avoid future change of default flag (e.g we might set metla to default on mac m1/m2)

crates/tabby/tests/goldentests.rs
```diff
@@ -84,29 +86,50 @@ async fn golden_test(body: serde_json::Value) -> serde_json::Value {
...
+    assert_golden!(json!({
             "language": "python",
             "segments": {
                 "prefix": "import datetime\n\ndef parse_expenses(expenses_string):\n    \"\"\"Parse the list of expenses and return the list of triples (date, value, currency).\n    Ignore lines starting with #.\n    Parse the date using datetime.\n    Example expenses_string:\n        2016-01-02 -34.01 USD\n        2016-01-03 2.59 DKK\n        2016-01-03 -2.72 EUR\n    \"\"\"\n    for line in expenses_string.split('\\n'):\n        "
             }
-    })).await;
+    }));
+}
+
+#[tokio::test]
```
wsxiaoys, line 118: is there a way to ensure the test only runs sequentially within the file? On my system (mac m1/m2), it tries to start `run_golden_tests` and `run_golden_tests_cpu` in parallel, thus caused the port confliction.

wsxiaoys (changes requested):

wsxiaoys:
otherwise LGTM

crates/tabby/tests/goldentests.rs
boxbeam: This is really just specifying what to pass for the `--device` flag. The default behavior is to not pass the flag at all, and I think that makes sense. Are you saying you'd rather make a new enum, like `Device::CPU|Cuda|Rocm|Metal`?

crates/tabby/tests/goldentests.rs
boxbeam: I'm not quite sure how to do that - especially since I think this conflict would also apply between the `goldentests_chat` file and `goldentests` file, since they both use the same port. We could change the port to fix that, though. I'll see if I can figure out how to run sequentially within an individual file.

boxbeam: Attempted this using the `serial_test` crate, could you please rerun and let me know if this is working for you now?

wsxiaoys: > especially since I think this conflict would also apply between the goldentests_chat file and goldentests file,

test between different files are sequential.

> Attempted this using the serial_test crate, could you please rerun and let me know if this is working for you now?
works for me.

crates/tabby/tests/goldentests.rs
boxbeam: Oh, so `--device cpu` works too? Wasn't sure; I'll pass that.

PR #1262, opened 2024-01-22
fix(agent): fix postprocess remove duplicated block closing line.

icycodes:
thanks to contribution of @erfanium


erfanium:
wow, that was really fast! thanks
I will report other postprocess issues if i face.

Issue #1171, opened 2024-01-07
[TAB-443] Provide a way to deactivate a user


PeterTucker:
**Please describe the feature you want**
There is no way to  deactivate a user (member). I looked in Team Management and there are no options.

**Additional context**

![New Project](https://uploads.linear.app/36e8c471-0050-4ed1-a34b-9a4f22609cd1/78a3671d-b669-4877-88ae-1b7b9325e46f/48564dc3-04cb-4432-81e3-9dbcd7e6df91?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzM2ZThjNDcxLTAwNTAtNGVkMS1hMzRiLTlhNGYyMjYwOWNkMS83OGEzNjcxZC1iNjY5LTQ4NzctODhhZS0xYjdiOTMyNWU0NmYvNDg1NjRkYzMtMDRjYi00NDMyLTgxZTMtOWRiY2Q3ZTZkZjkxIiwiaWF0IjoxNzA1NTExODAxLCJleHAiOjE3MDU1OTgyMDF9.m5cUEcgDvDIeuuZkVl_eo_NoLBUEVgyb62Hgzz33C2c)

- [X] Backend support: @jmarcuse
- [ ] Frontend support: @jueliang

<sub>From [SyncLinear.com](https://synclinear.com) | [TAB-443](https://linear.app/tabbyml/issue/TAB-443/provide-a-way-to-deactivate-a-user)</sub>

liangfung:
@jmarcuse `schema.graphql` has not been updated yet.
@wsxiaoys Woud it be possible to consider automating the execution within some hooks?

wsxiaoys:
offline resolved

PR #1269, opened 2024-01-22
feat(db, webserver): Allow disabling users

boxbeam:
Relates to #1171


ee/tabby-db/migrations/02-users-table/up.sql
```diff
@@ -6,7 +6,8 @@ CREATE TABLE users (
     created_at         TIMESTAMP DEFAULT (DATETIME('now')),
     updated_at         TIMESTAMP DEFAULT (DATETIME('now')),
     auth_token         VARCHAR(128) NOT NULL,
+    active             BOOLEAN NOT NULL DEFAULT 1,
```
wsxiaoys, line 9: you cannot modify an released migration.

Create a new migration for this.

ee/tabby-db/src/users.rs
```diff
@@ -203,6 +206,16 @@ impl DbConn {
 
         Ok(())
     }
+
+    pub async fn update_user_active(&self, id: i32, active: bool) -> Result<()> {
```
wsxiaoys, line 210:
```suggestion
    pub async fn toggle_user_active(&self, id: i32) -> Result<()> {
```

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -326,6 +326,11 @@ impl Mutation {
         }
     }
 
+    async fn update_user_active(ctx: &Context, id: i32, active: bool) -> Result<bool> {
```
wsxiaoys, line 329:
```suggestion
    async fn update_user_active(ctx: &Context, id: ID, active: bool) -> Result<bool> {
```

wsxiaoys (changes requested):

ee/tabby-db/src/users.rs
boxbeam: There's a few reasons I think `update` should be preferred to `toggle`. A toggle requires two database operations, one to read the current value and one to write. The frontend will already have the information about whether the user is disabled, and can just invert that and call update. This can also help keep things in sync; if two admins have an active user on their screen and both click a button to toggle whether that user is active, the user would be toggled twice, left in the same state it started in, and this would be unexpected behavior for both admins. If each clicks it and they both call update instead of toggle, the user will be left disable, and I believe this behavior is more in line with what an admin would expect.

wsxiaoys: you can do update in a single database operation? something like `UPDATE .... SET active = !active`?

> if two admins have an active user on their screen and both click a button to toggle whether that user is active, the user would be toggled twice

Good point, we could pass in the expect current value of user for the purpose (e.g `UPDATE ... SET active = !active WHERE ...`)

boxbeam: At that point, we're back to `update_user_active` except you would be supplying the *opposite* of the value you want set. I think it'd be more clear to just leave it as passing the value you want set.

wsxiaoys: that works, but i'd really like to throw an error when you're setting same value for `active` status. That usually means there's an frontend displaying error for an critical operations like this. I'm fine to making it named as `update_user_active`, as long as it checks the current value.

boxbeam: Sounds good, I'll do that in a bit

ee/tabby-db/src/users.rs
```diff
@@ -203,6 +206,23 @@ impl DbConn {
...
+        let new_active = self
+            .conn
+            .call(move |c| {
+                c.execute("UPDATE users SET active=? WHERE id=?", (active, id))?;
```
wsxiaoys, line 214: can this being something like `UPDATE users SET active=? WHRERE id=? and active=?, (active, id, !active)`? So you could decide whether to throw error based on number of rows being changed (returned by query)

ee/tabby-db/src/users.rs
```diff
@@ -208,16 +208,17 @@ impl DbConn {
     }
 
     pub async fn update_user_active(&self, id: i32, active: bool) -> Result<()> {
```
wsxiaoys, line 210: please also add a unit test on this

wsxiaoys: (expecting error if there's no change)

boxbeam: I was pushing the test as you left this comment :')

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -66,6 +66,9 @@ pub enum CoreError {
 
     #[error(transparent)]
     Other(#[from] anyhow::Error),
+
+    #[error("{0}")]
+    ParseInt(#[from] ParseIntError),
```
wsxiaoys, line 71:
```suggestion
    #[error("ID is not vallid due to ....")]
    InvalidError(#[from] ParseIntError),
```

Issue #1261, opened 2024-01-21
[TAB-456] Provide a solution to collect repositories from tabby webserver and run the cron job (instead of reading from config.toml)


wsxiaoys:
Related code:

1. add method to expose repositories in Hub service.

[https://github.com/TabbyML/tabby/blob/f220ba9b06b69525d6aa7dfbd213cc28220f6c45/ee/tabby-webserver/src/hub/api.rs#L15](https://github.com/TabbyML/tabby/blob/f220ba9b06b69525d6aa7dfbd213cc28220f6c45/ee/tabby-webserver/src/hub/api.rs#L15)

1. collect repositories from Hub service in scheduler

[https://github.com/TabbyML/tabby/blob/cb0e3b154232accf3d40cf8f778e0b8506d2d3de/crates/tabby/src/worker.rs#L102](https://github.com/TabbyML/tabby/blob/cb0e3b154232accf3d40cf8f778e0b8506d2d3de/crates/tabby/src/worker.rs#L102)

<sub>From [SyncLinear.com](https://synclinear.com) | [TAB-456](https://linear.app/tabbyml/issue/TAB-456/provide-a-solution-to-collect-repositories-from-tabby-webserver-and)</sub>

PR #1271, opened 2024-01-22
feat(webserver): Add endpoint to fetch repositories from hub server

boxbeam:
Relates to #1261

### DIFF
--- a/ee/tabby-webserver/src/hub/api.rs
+++ b/ee/tabby-webserver/src/hub/api.rs
@@ -27,7 +27,7 @@ pub trait Hub {
         offset: usize,
     ) -> SearchResponse;
 
-    async fn get_repositories() -> Result<Vec<RepositoryConfig>, String>;
+    async fn get_repositories() -> Vec<RepositoryConfig>;
 }
 
 pub fn tracing_context() -> tarpc::context::Context {
--- a/ee/tabby-webserver/src/hub/api.rs
+++ b/ee/tabby-webserver/src/hub/api.rs
@@ -2,9 +2,12 @@ use async_trait::async_trait;
 use axum::{headers::Header, http::HeaderName};
 use hyper::Request;
 use serde::{Deserialize, Serialize};
-use tabby_common::api::{
-    code::{CodeSearch, CodeSearchError, SearchResponse},
-    event::RawEventLogger,
+use tabby_common::{
+    api::{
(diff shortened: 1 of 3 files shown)


ee/tabby-webserver/src/hub/mod.rs
```diff
@@ -133,4 +136,8 @@ impl Hub for Arc<HubImpl> {
             }
         }
     }
+
+    async fn get_repositories(self, _context: tarpc::context::Context) -> RepositoryCache {
+        META.clone()
```
wsxiaoys, line 141: instead of reading from META, we should really reading this from sqlite with [RepositoryService](https://github.com/TabbyML/tabby/blob/84f27ce67d2ec63cdc9e476e988d54fdf4ebe65d/ee/tabby-webserver/src/schema/repository.rs#L33)

wsxiaoys (changes requested):

ee/tabby-webserver/src/hub/mod.rs
```diff
@@ -137,7 +134,14 @@ impl Hub for Arc<HubImpl> {
         }
     }
 
-    async fn get_repositories(self, _context: tarpc::context::Context) -> RepositoryCache {
-        META.clone()
+    async fn get_repositories(
+        self,
+        _context: tarpc::context::Context,
+    ) -> Result<Vec<Repository>, String> {
```
wsxiaoys, line 140: we should returns `tabby_common::Repository`, since it's the struct visible to `tabby`

boxbeam: I'm assuming you meant RepositoryConfig since there's no Repository struct in tabby_common

wsxiaoys: Right - RepositoryConfig

ee/tabby-webserver/src/hub/api.rs
```diff
@@ -23,6 +26,8 @@ pub trait Hub {
         limit: usize,
         offset: usize,
     ) -> SearchResponse;
+
+    async fn get_repositories() -> Result<Vec<RepositoryConfig>, String>;
```
wsxiaoys, line 30: Let’s just release `Vec<RepositoryConfig>` for simplicity - you could handle error and return empty vec instead.

PR #1274, opened 2024-01-22
fix(webserver): disable scheduler jobs in current upcoming release

wsxiaoys:


ee/tabby-webserver/src/service/cron/mod.rs
```diff
@@ -16,29 +16,38 @@ async fn new_job_scheduler(jobs: Vec<Job>) -> anyhow::Result<JobScheduler> {
...
-        // run every 5 hours
-        let Ok(job3) =
-            job_utils::run_job(db_conn.clone(), "index".to_owned(), "0 0 1/5 * * * *").await
-        else {
-            error!("failed to create index job");
-            return;
-        };
+        jobs.push(job1);
+
+        if (turn_on_scheduler_jobs) {
```
boxbeam: Looks fine for now, but if we end up adding more feature-gated scheduled jobs, we might want to try to take a more declarative approach to this

boxbeam (approved):

ee/tabby-webserver/src/service/cron/mod.rs
wsxiaoys: sure - a env flag / config flag based approach could be adopt in future

PR #1284, opened 2024-01-23
fix(webserver): Remove redundant scheduler code

### LINKED ISSUE #1283: Scheduler has redundant logic
The Tabby scheduler has redundant logic that we can simplify.

https://github.com/TabbyML/tabby/blob/673768e4cb2fab6dfa6c987176aa6c9fdc02da80/ee/tabby-webserver/src/service/cron/mod.rs#L15 - This function spawns a coroutine that will tick the scheduler every 500ms

https://github.com/TabbyML/tabby/blob/673768e4cb2fab6dfa6c987176aa6c9fdc02da80/ee/tabby-webserver/src/service/cron/mod.rs#L50 - Here we use our function to get this scheduler which is already running in the background, and spawn another loop running in the background to tick the scheduler

We should be able to get rid of the loop entirely, and just rely on `.start()`.

boxbeam:
Closes #1283

### DIFF
--- a/ee/tabby-webserver/src/service/cron/mod.rs
+++ b/ee/tabby-webserver/src/service/cron/mod.rs
@@ -1,8 +1,6 @@
 mod db;
 mod job_utils;
 
-use std::time::Duration;
-
 use tabby_db::DbConn;
 use tokio_cron_scheduler::{Job, JobScheduler};
 use tracing::error;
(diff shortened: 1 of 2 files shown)


wsxiaoys:
cc @darknight for review

darknight:
looks good to me!

PR #1285, opened 2024-01-24
fix(ui): Update condition logic in did-auth-error

liangfung:
Related #1279

### DIFF
--- a/ee/tabby-ui/lib/tabby/auth.tsx
+++ b/ee/tabby-ui/lib/tabby/auth.tsx
@@ -75,10 +75,10 @@ const saveAuthToken = (authData: AuthData) => {
 const clearAuthToken = () => {
   localStorage.removeItem(AUTH_TOKEN_KEY)
   window.dispatchEvent(
-    new StorageEvent("storage", {
+    new StorageEvent('storage', {
       storageArea: window.localStorage,
       url: window.location.href,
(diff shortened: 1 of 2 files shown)


ee/tabby-ui/lib/tabby/auth.tsx
```diff
@@ -74,6 +74,13 @@ const saveAuthToken = (authData: AuthData) => {
 }
 const clearAuthToken = () => {
   localStorage.removeItem(AUTH_TOKEN_KEY)
+  window.dispatchEvent(
```
wsxiaoys, line 77: do we need to manually trigger this? i thought removeItem will trigger such event automatically?

wsxiaoys: got it - in tabby-ui, which part of code relying on this event? could you point me the link?

ee/tabby-ui/lib/tabby/auth.tsx
```diff
@@ -74,6 +74,9 @@ const saveAuthToken = (authData: AuthData) => {
 }
 const clearAuthToken = () => {
   localStorage.removeItem(AUTH_TOKEN_KEY)
+  // FIXME
```
wsxiaoys, line 77:
```suggestion
  // FIXME(liangfung)
```

PR #1294, opened 2024-01-25
test(agent): add a fail case test for postprocess

erfanium:
image

I'm not sure about test description and file name, feel free to change it

### DIFF
--- a/clients/tabby-agent/src/postprocess/golden/remove_duplication/duplicated_line_suffix.toml
+++ b/clients/tabby-agent/src/postprocess/golden/remove_duplication/duplicated_line_suffix.toml
@@ -4,7 +4,7 @@ description = 'Remove duplication: dupicated line suffix'
 # use default config
 
 [context]
-filepath = 'checks.js'
+filepath = 'checks.ts'
 language = 'typescript'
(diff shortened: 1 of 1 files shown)


clients/tabby-agent/src/postprocess/golden/remove_duplication/duplicated_line_suffix.toml
```diff
@@ -0,0 +1,23 @@
...
+}
+'''
+
+[expected]
+# text = '''
+# const log = ({ ok }: { ok:├ boolean┤ }) => {
+#   console.log(ok);
+# }
+# '''
+unchanged = true
```
icycodes, line 22:
```suggestion
text = '''
const log = ({ ok }: { ok:├ boolean┤ }) => {
  console.log(ok);
}
'''
```

It is ok to add what is expected here, no need to comment it out. The `notEqual` flag will assert that the output is not equal to the expected result.

erfanium:
> However, it seems to be a case that is much more complex than what post-processing currently can handle.

One idea is to compare ending of `last non-blank line of completion` with `context.currentLineSuffix`. if the endings share the same sequence, the we can trim it.

PR #1296, opened 2024-01-25
feat(tabby): Allow passing temperature and seed to chat completion requests

### LINKED ISSUE #1238: Add support for seed and temperature in the chat_completion API

Allow seed and temperature to be specified in the chat completion API, similar to #1216 

---

boxbeam:
Closes #1238

### DIFF
--- a/crates/tabby-inference/src/lib.rs
+++ b/crates/tabby-inference/src/lib.rs
@@ -24,6 +24,17 @@ pub struct TextGenerationOptions {
     pub language: Option<&'static Language>,
 }
 
+impl TextGenerationOptions {
+    pub const DEFAULT_TEMPERATURE: f32 = 0.1;
+
+    pub fn default_seed() -> u64 {
+        std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap()
+            .as_millis() as u64
+    }
+}
+
 #[async_trait]
 pub trait TextGeneration: Sync + Send {
     async fn generate(&self, prompt: &str, options: TextGenerationOptions) -> String;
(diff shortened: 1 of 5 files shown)


boxbeam:
Golden tests passing

crates/tabby/src/services/mod.rs
```diff
@@ -4,3 +4,12 @@ pub mod completion;
 pub mod event;
 pub mod health;
 pub mod model;
+
+pub const DEFAULT_TEMPERATURE: f32 = 0.1;
+
+pub fn default_seed() -> u64 {
```
wsxiaoys, line 10: I feel this better being a static method of `TextInferenceEngine`

crates/tabby/src/services/mod.rs
```diff
@@ -4,3 +4,12 @@ pub mod completion;
 pub mod event;
 pub mod health;
 pub mod model;
+
+pub const DEFAULT_TEMPERATURE: f32 = 0.1;
```
wsxiaoys, line 8: same as above

crates/tabby/src/services/mod.rs
boxbeam: Put it on `TextGenerationOptions`; `TextInferenceEngine` is not accessible from where it's needed for that

crates/tabby-inference/src/lib.rs
```diff
@@ -24,6 +24,17 @@ pub struct TextGenerationOptions {
     pub language: Option<&'static Language>,
 }
 
+impl TextGenerationOptions {
+    pub const DEFAULT_TEMPERATURE: f32 = 0.1;
```
wsxiaoys, line 28: do we still need to have this since we set default value to be 0.1 in `builder`?

You might remove it and cleanup the code in service

wsxiaoys (changes requested):
Otherwise LGTM

wsxiaoys (approved):

PR #1308, opened 2024-01-26
fix(swagger): fix swagger ui by adding missing types

wsxiaoys:


katopz:
i got this from `0.7.0`
```
can't parse JSON.  Raw result:

[object Blob]
```

image

Not sure is this fix that issue or related?

wsxiaoys:
> i got this from `0.7.0`
> 
> ```
> 
> can't parse JSON.  Raw result:
> 
> 
> 
> [object Blob]
> 
> ```
> 
> 
> 
> image
> 
> 
> 
> Not sure is this fix that issue or related?

it is irrelevant.

Swagger UI is not able to display streaming output.

PR #1322, opened 2024-01-29
fix(juniper_axum): Fix how has_previous_page and has_next_page are computed

boxbeam:
Remaining flaw: `has_previous_page` will always be true if `before` or `after` are specified, even if all possible values are returned. This cannot be resolved with the current setup of functions.

### DIFF
--- a/crates/juniper-axum/src/relay/connection.rs
+++ b/crates/juniper-axum/src/relay/connection.rs
@@ -33,7 +33,8 @@ where
         first: Option<usize>,
         last: Option<usize>,
     ) -> Self {
-        let has_next_page = first.or(last).is_some_and(|count| nodes.len() > count);
+        let selected_count = first.or(last).unwrap_or(nodes.len());
+        let has_next_page = selected_count < nodes.len();
         let has_previous_page = after || before;
         let len = nodes.len();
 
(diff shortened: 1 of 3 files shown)


crates/juniper-axum/src/relay/connection.rs
```diff
@@ -27,8 +27,8 @@ where
     }
 
     pub fn build_connection(nodes: Vec<Node>, first: Option<usize>, last: Option<usize>) -> Self {
-        let has_next_page = first.map_or(false, |first| nodes.len() > first);
-        let has_previous_page = last.map_or(false, |last| nodes.len() > last);
+        let has_next_page = first.is_some_and(|first| nodes.len() > first);
+        let has_previous_page = first.is_some_and(|first| first > 0);
```
wsxiaoys, line 31: This will return true even for the first page, while it should actually being false.

boxbeam: The first element has ID 1?

wsxiaoys: first is not offset, it means pick `first` cnt of elements after `cursor`

liangfung (approved):

PR #1325, opened 2024-01-30
fix(ui): enrich the content of acticating/deactivating toast

liangfung:
1. enrich the content of toast
2. try to fix the bg color bug of toast component


liangfung:
https://jam.dev/c/5a64ab81-31a4-481a-b1b8-a632036de80c

PR #1332, opened 2024-01-31
doc(oauth): add spec for api `/oauth/callback/github`

liangfung:


ee/tabby-webserver/docs/api_spec.md
```diff
@@ -191,7 +191,7 @@ Redirect to oauth provider for signin
 
 ```shell
```
wsxiaoys, line 192: can we merge the example of github / google? they seems to be quite similar to me.

PR #1339, opened 2024-01-31
docs: update online docs for clients config about timeout limit.

icycodes:
Remove documentation about the client-side timeout limit, as the timeout limit has been removed in clients after version 1.3.0.


wsxiaoys:
cc @gyxlucy

PR #1341, opened 2024-01-31
feat(webserver): Populate the user field in completions requests if the user authenticated

### LINKED ISSUE #1324: [TAB-471] Fill the user field in request for an authenticated instance
When authorization field is enabled, email address of user should be filled for requests of `/v1/completion`

Code Location: [https://github.com/TabbyML/tabby/blob/5dc3303f73a60beb5867f1e242f88d509fc44606/ee/tabby-webserver/src/service/mod.rs#L153](https://github.com/TabbyML/tabby/blob/5dc3303f73a60beb5867f1e242f88d509fc44606/ee/tabby-webserver/src/service/mod.rs#L153)

<sub>From [SyncLinear.com](https://synclinear.com) | [TAB-471](https://linear.app/tabbyml/issue/TAB-471/fill-the-user-field-in-request-for-an-authenticated-instance)</sub>

boxbeam:
Closes #1324

### DIFF
--- a/crates/tabby/src/routes/completions.rs
+++ b/crates/tabby/src/routes/completions.rs
@@ -24,9 +24,12 @@ use crate::services::completion::{CompletionRequest, CompletionResponse, Complet
 #[instrument(skip(state, request))]
 pub async fn completions(
     State(state): State<Arc<CompletionService>>,
+    TypedHeader(MaybeUser(user)): TypedHeader<MaybeUser>,
     Json(mut request): Json<CompletionRequest>,
-    TypedHeader(user): TypedHeader<MaybeUser>,
 ) -> Result<Json<CompletionResponse>, StatusCode> {
+    if let Some(user) = user {
+        request.user.replace(user);
+    }
     match state.generate(&request).await {
         Ok(resp) => Ok(Json(resp)),
         Err(err) => {
@@ -37,7 +40,7 @@ pub async fn completions(
 }
 
 #[derive(Debug)]
-struct MaybeUser(Option<String>);
+pub struct MaybeUser(Option<String>);
 
 impl Header for MaybeUser {
     fn name() -> &'static axum::http::HeaderName {
--- a/crates/tabby/src/routes/completions.rs
+++ b/crates/tabby/src/routes/completions.rs
@@ -1,7 +1,8 @@
 use std::sync::Arc;
 
-use axum::{extract::State, Json};
+use axum::{extract::State, headers::Header, Json, TypedHeader};
 use hyper::StatusCode;
+use tabby_webserver::public::USER_HEADER_FIELD_NAME;
 use tracing::{instrument, warn};
 
 use crate::services::completion::{CompletionRequest, CompletionResponse, CompletionService};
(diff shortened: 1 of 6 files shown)


crates/tabby/src/routes/completions.rs
```diff
@@ -23,7 +24,8 @@ use crate::services::completion::{CompletionRequest, CompletionResponse, Complet
 #[instrument(skip(state, request))]
 pub async fn completions(
     State(state): State<Arc<CompletionService>>,
-    Json(request): Json<CompletionRequest>,
+    Json(mut request): Json<CompletionRequest>,
+    TypedHeader(user): TypedHeader<MaybeUser>,
```
wsxiaoys, line 28: line 28 need to be place before  line 27.

ee/tabby-webserver/src/service/mod.rs
```diff
@@ -60,41 +63,31 @@ impl ServerContext {
         }
     }
 
-    async fn authorize_request(&self, request: &Request<Body>) -> bool {
+    async fn authorize_request(&self, request: &Request<Body>) -> (bool, Option<String>) {
```
wsxiaoys, line 66: nit: just return `Option<String>`? None means invalid authentication

wsxiaoys (approved):
Otherwise LGTM

ee/tabby-webserver/src/service/mod.rs
boxbeam: I'm following the logic of how this was implemented before. There were some cases (like for an access token) where it could be authenticated successfully, but we wouldn't be able to determine a user from it. You'll note that there are paths that return `(true, None)` here. I didn't want to change the authentication logic, so I just refactored a little and made it also return the user if possible.

ee/tabby-webserver/src/service/mod.rs
```diff
@@ -60,41 +63,31 @@ impl ServerContext {
...
+            .and_then(Result::ok);
+
+        let token = authorization
+            .and_then(|s| s.split_once(' '))
+            .map(|(_bearer, token)| token);
+
+        let Some(token) = token else {
+            return (false, None);
+        };
+        if self.db_conn.verify_access_token(token).await.is_ok() {
```
wsxiaoys, line 85: it's always possible to extract email from access token, it's being filled in the `sub` field:

https://github.com/TabbyML/tabby/blob/1f51a0946a9aa9f2b3bf52931b9da5ca65225439/ee/tabby-webserver/src/schema/auth.rs#L254

boxbeam: Isn't the return being `(true, None)` for unauthenticated routes still required, though?

wsxiaoys: ah, you're right, we still to maintain them (everything other than `/v1/` `/v1beta/` don't have to do authorization through this middle ware).

wsxiaoys: let's fix the verify_access_token and this PR is good to go

boxbeam: Already did with [07a0124](https://github.com/TabbyML/tabby/pull/1341/commits/07a0124434f17b1073f63747df39b984cb48bb92)

PR #1359, opened 2024-02-02
feat(db): Split prod and dev db, default to dev

### LINKED ISSUE #1272: [TAB-461] Refactor database code to use sqlx instead of rusqlite
Switch all database code to use sqlx instead of rusqlite.

- [X] Create full db schema matching current spec using sqlx commands (`cargo sqlx`)
- [X] Remove rusqlite dependency and add sqlx dependency
- [X] Migrate all code and tests to use sqlx instead
- [X] Ensure migration path for users upgrading to the sqlx version
- [ ] Split Prod / Dev environment for tabby webserver

Reference: [https://github.com/TabbyML/tabby/pull/1207](https://github.com/TabbyML/tabby/pull/1207)

<sub>From [SyncLinear.com](https://synclinear.com) | [TAB-461](https://linear.app/tabbyml/issue/TAB-461/refactor-database-code-to-use-sqlx-instead-of-rusqlite)</sub>

boxbeam:
Closes #1272 

NOTE: This will immediately present a breaking change for anyone who currently has a prod db and is expecting it to be used on startup. We might want to document this, but if so, which document(s)?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,17 +35,17 @@ jobs:
             target: x86_64-unknown-linux-gnu
             binary: x86_64-manylinux2014
             container: quay.io/pypa/manylinux2014_x86_64
-            build_args: --features static-ssl
+            build_args: --features static-ssl,prod
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
             binary: x86_64-manylinux2014-cuda117
             container: sameli/manylinux2014_x86_64_cuda_11.7
-            build_args: --features static-ssl --features cuda
+            build_args: --features static-ssl --features cuda,prod
           - os: ubuntu-latest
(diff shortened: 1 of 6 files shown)


wsxiaoys:
Please also change:
1. https://github.com/TabbyML/tabby/blob/e1312c7b035c3db8c9eed607a58aa35cbacb2fce/.github/workflows/release.yml#L38
2. https://github.com/TabbyML/tabby/blob/e1312c7b035c3db8c9eed607a58aa35cbacb2fce/Dockerfile#L41

wsxiaoys:
> NOTE: This will immediately present a breaking change for anyone who currently has a prod db and is expecting it to be used on startup. We might want to document this, but if so, which document(s)?

It won't - since the only one we break are those running from our source code repository

PR #1364, opened 2024-02-02
fix(agent): add removeDuplicatedLineSuffix postprocess filter

erfanium:
here is my attempt to fix `clients/tabby-agent/src/postprocess/golden/remove_duplication/duplicated_line_suffix.toml` test.
I also tested this on a daily usage. the fix just works fine.

but currently not all the tests are passing, because of a logical interference in `removeDuplicatedLineSuffix` and `calculateReplaceRange` filters. `removeDuplicatedLineSuffix` approach is to trim the input. but `calculateReplaceRange` approach is to keep the current input and modify the replaceRange.

should i refactor this filter and move it under replaceRange filter categories?

### DIFF
--- a/clients/tabby-agent/src/postprocess/golden/remove_duplication/duplicated_line_suffix.toml
+++ b/clients/tabby-agent/src/postprocess/golden/remove_duplication/duplicated_line_suffix.toml
@@ -8,7 +8,7 @@ filepath = 'checks.ts'
 language = 'typescript'
 # indentation = '  ' # not specified
 text = '''
-const log = ({ ok }: { ok:├ boolean }) => { ┤ }) => {
+const log = ({ ok }: { ok:├ boolean }) => {┤ }) => {
   console.log(ok);
 }
 '''
@@ -19,4 +19,3 @@ const log = ({ ok }: { ok:├ boolean┤ }) => {
   console.log(ok);
 }
 '''
-notEqual = true  # FIXME: fix bad case
--- a/clients/tabby-agent/src/postprocess/index.ts
+++ b/clients/tabby-agent/src/postprocess/index.ts
@@ -12,6 +12,7 @@ import { trimMultiLineInSingleLineMode } from "./trimMultiLineInSingleLineMode";
 import { dropDuplicated } from "./dropDuplicated";
 import { dropBlank } from "./dropBlank";
 import { calculateReplaceRange } from "./calculateReplaceRange";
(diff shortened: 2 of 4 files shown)


icycodes:
Hi, @erfanium!

Thank you for your PR!

I noticed that the filter `removeDuplicatedLineSuffix` seems to handle many cases that the `calculateReplaceRange` or `trimMultiLineInSingleLineMode` should handle.

Let me explain the current filters:

The `calculateReplaceRange`: 
If the current line suffix characters are all auto-closing characters like `)` and `]`, which should be deleted when applying the suggestion, I think they should be included in the replace range.

The `trimMultiLineInSingleLineMode`: 
If the current line suffix characters contain non-auto-closing characters, it means that the user is filling in a line. Therefore, the completion is expected to be a single line. 
In this case, if the server-generated completion is multi-line, I only keep the first line and remove the duplicated suffix. If the server-generated completion is just one line, it mostly appears to be a good FIM case, so I keep it as is.

I remember that you mentioned in the Slack channel that you forked the tabby server source and added checking for multi-line or single-line generation, with `\n` as a stop character for the single-line case. I think this is the only case that `removeDuplicatedLineSuffix` can handle but `trimMultiLineInSingleLineMode` can not.

I'm not entirely sure about this, so please correct me if I'm misunderstanding.

wsxiaoys:
Closing as this is stale

erfanium:
@icycodes I'm so sorry. I had totally forgotten this PR.

>  I think this is the only case that removeDuplicatedLineSuffix can handle but trimMultiLineInSingleLineMode can not.

do you think tabby-agent should handle this scenario (server returning only the first line)? If yes, I can fix this PR based on your suggestions

icycodes:
> @icycodes I'm so sorry. I had totally forgotten about this PR.
> 
> > I think this is the only case that removeDuplicatedLineSuffix can handle but trimMultiLineInSingleLineMode can not.
> 
> do you think tabby-agent should handle this scenario (server returning only the first line)? If yes, I can fix this PR based on your suggestions

I think it should be handled. 
When the server returned only one line, I assumed it was a good FIM result and skipped the post-processing. However, it seems that this assume is not correct. 

It's not recommended to add a new filter because `trimMultiLineInSingleLineMode` has already processing for `mode == "fill-in-line" && inputLines.length > 1`. I think that adding some code here for `inputLines.length == 1` would be a simpler approach. We could rename this filter to something like `trimSuffixInSingleLineMode`.

erfanium:
going to reopen a new pr for this

PR #1382, opened 2024-02-05
feat(webserver): Use hash IDs rather than numeric row IDs in external API

boxbeam:
Closes TAB-470


ee/tabby-webserver/src/lib.rs
```diff
@@ -24,3 +29,25 @@ pub mod public {
...
 }
+
+static HASHER: OnceLock<HashIds> = OnceLock::new();
+
+#[derive(thiserror::Error, Debug)]
+#[error("Invalid ID")]
+pub struct InvalidIDError;
+
+fn hasher() -> &'static HashIds {
+    HASHER.get_or_init(|| HashIds::builder().with_salt("tabby-id-serializer").finish())
```
wsxiaoys, line 40: seems you could just use `lazy_static` for the purpose?

wsxiaoys (approved):

ee/tabby-webserver/src/lib.rs
wsxiaoys, line 1: revert this file?

PR #1387, opened 2024-02-05
feat(db): Add new email settings

boxbeam:


ee/tabby-db/src/email_setting.rs
wsxiaoys, line 1: update daos as well?

wsxiaoys, line 1: add params to update_email_setting as well?

ee/tabby-db/src/email_setting.rs
```diff
@@ -10,13 +10,16 @@ pub struct EmailSettingDAO {
     pub smtp_username: String,
     pub smtp_password: String,
     pub smtp_server: String,
+    pub from_address: Option<String>,
+    pub encryption: String,
```
wsxiaoys, line 14: use an enum

ee/tabby-db/src/email_setting.rs
```diff
@@ -10,13 +10,16 @@ pub struct EmailSettingDAO {
...
+    pub auth_method: String,
```
wsxiaoys, line 15: ditto

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,16 @@
...
 
-#[derive(GraphQLObject)]
+#[derive(GraphQLObject, Validate)]
 pub struct EmailSetting {
     pub smtp_username: String,
+    #[validate(url)]
     pub smtp_server: String,
+    pub from_address: String,
+    pub encryption: String,
+    pub auth_method: String,
```
wsxiaoys, line 13: 12/13 should be enum

ee/tabby-db/src/email_setting.rs
```diff
@@ -10,13 +10,16 @@ pub struct EmailSettingDAO {
     pub smtp_username: String,
     pub smtp_password: String,
     pub smtp_server: String,
+    pub from_address: Option<String>,
```
wsxiaoys, line 13: still wanna add port field?

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,16 @@
 use anyhow::Result;
 use async_trait::async_trait;
 use juniper::GraphQLObject;
+use validator::Validate;
 
-#[derive(GraphQLObject)]
+#[derive(GraphQLObject, Validate)]
```
wsxiaoys, line 6: create separate struct for validation, eg EnailSettingInput

wsxiaoys (changes requested):

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,120 @@
...
+    field: &'static str,
+    value: &T,
+) -> ValidationError {
+    let mut err = ValidationError::new(msg);
+    err.add_param(field.into(), value);
+    err
+}
+
+impl EmailSetting {
+    pub fn new_validate(
```
wsxiaoys, line 63: I don't think we should do validate in this direction (the data is already in db, thus no need to implement it as a ValidationError).

I'll suggest:
1. keep the implementation, but return an `anyhow::Error`
2. In EmailService::get_email_setting, whenever a readed record is not valid (new_validate failed):
    - delete the email setting record from database.
    - return None instead.

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,120 @@
...
+        Ok(EmailSetting {
+            smtp_username,
+            smtp_server,
+            from_address,
+            encryption,
+            auth_method,
+        })
+    }
+
+    pub fn convert_encryption(encryption: &str) -> Result<Encryption, ValidationError> {
```
wsxiaoys, line 91: I suggest just use serde serialize / deserialize to parse / serialize the enums. (though still return an anyhow error if deserialize failed)

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -434,10 +435,20 @@ impl Mutation {
         smtp_username: String,
         smtp_password: Option<String>,
         smtp_server: String,
+        from_address: Option<String>,
+        encryption: Encryption,
+        auth_method: AuthMethod,
     ) -> Result<bool> {
```
wsxiaoys, line 441: require admin.

also for `email_setting` query and `delete_email_setting` mutation.

wsxiaoys (changes requested):

ee/tabby-webserver/src/schema/email.rs
boxbeam: I don't think using serde here really makes sense. We would need to implement a custom Serializer/Deserializer. I considered using strum as well, but pulling that in just for this also seems overkill. I think it's fine to have a few string-based match statements for now until we have enough of them that it warrants conversion to strum, which I think we should prefer over serde for direct string <-> enum mapping.

crates/tabby-common/src/enum_conversion.rs
wsxiaoys, line 1: Move this to tabby-webserver (schema/macros.rs)

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,70 @@
...
+        auth_method: String,
+    ) -> Result<Self> {
+        if !validate_url(&smtp_server) {
+            return Err(anyhow!("Invalid smtp server address"));
+        }
+
+        let encryption = Encryption::from_db_str(&encryption)?;
+        let auth_method = AuthMethod::from_db_str(&auth_method)?;
+
+        let from_address = from_address.unwrap_or_else(|| smtp_username.clone());
```
wsxiaoys, line 59: when from_address is null, return error.

ee/tabby-db/migrations/0011_new-email-settings.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE email_setting ADD COLUMN from_address VARCHAR(255);
+ALTER TABLE email_setting ADD COLUMN encryption VARCHAR(255) NOT NULL DEFAULT 'SSL/TLS';
```
wsxiaoys: consistent with enum?

ee/tabby-db/migrations/0011_new-email-settings.up.sql
```diff
@@ -0,0 +1,3 @@
...
+ALTER TABLE email_setting ADD COLUMN auth_method VARCHAR(255) NOT NULL DEFAULT 'PLAIN';
```
wsxiaoys: ditto

ee/tabby-webserver/src/service/email.rs
```diff
@@ -76,22 +108,40 @@ pub async fn new_email_service(db: DbConn) -> Result<impl EmailService> {
...
+        };
+        Ok(Some(creds))
     }
 
     async fn update_email_setting(
         &self,
         smtp_username: String,
         smtp_password: Option<String>,
         smtp_server: String,
+        from_address: Option<String>,
```
wsxiaoys, line 129: this should be required field from service / graphql api perspective

wsxiaoys (changes requested):

crates/tabby-common/src/enum_conversion.rs
```diff
@@ -0,0 +1,30 @@
+use anyhow::Result;
+
+pub trait DatabaseSerializable: Sized {
+    fn as_db_str(&self) -> &'static str;
+    fn from_db_str(s: &str) -> Result<Self>;
+}
+
+#[macro_export]
```
boxbeam, line 8: Could also have it generate it using the name of the variant as the string, but I wanted to avoid that in case we have variants that use a different db name for casing or formatting

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,70 @@
...
+use validator::validate_url;
+
+#[derive(GraphQLEnum, Clone, Debug)]
+pub enum Encryption {
+    StartTls,
+    SslTls,
+    None,
+}
+
+enum_mapping!(
```
wsxiaoys, line 14: put this in `dao.rs`

ee/tabby-webserver/src/schema/email.rs
boxbeam: Why not just take `String` instead of `Option<String>` then?

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,70 @@
...
+);
+
+#[derive(GraphQLEnum, Clone, Debug)]
+pub enum AuthMethod {
+    Plain,
+    Login,
+    XOAuth2,
+}
+
+enum_mapping!(
```
wsxiaoys, line 28: ditto

ee/tabby-db/migrations/0011_new-email-settings.up.sql
boxbeam: They are now

ee/tabby-webserver/src/schema/email.rs
wsxiaoys: that works as well. but since new_validate takes data from db side, it does have a chance being null (and you'll need to check it before calling try_from then)

boxbeam: Right. I had thought we wanted to allow this to be None too, and maybe just emit a warning about it rather than an error. An error would cause the entire settings object to be discarded and no email could be sent, so if we make it an error the migration immediately breaks your settings.

ee/tabby-db/src/conversions.rs
wsxiaoys, line 1: rename to enum.rs

ee/tabby-db/src/lib.rs
```diff
@@ -7,6 +7,7 @@ pub use repositories::RepositoryDAO;
 use sqlx::{query, query_scalar, Pool, Sqlite, SqlitePool};
 pub use users::UserDAO;
 
+pub mod conversions;
```
wsxiaoys: export DatabaseSerializable / enum_mapping directly

ee/tabby-db/src/conversions.rs
```diff
@@ -0,0 +1,28 @@
+pub trait DatabaseSerializable: Sized {
```
wsxiaoys:
```suggestion
pub trait DbEnum: Sized {
```

ee/tabby-db/src/conversions.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn as_db_str(&self) -> &'static str;
+    fn from_db_str(s: &str) -> anyhow::Result<Self>;
+}
+
+#[macro_export]
+macro_rules! enum_mapping {
```
wsxiaoys: 1. if convinent, can this being implemented as an derivation macro? (e.g DbSerializable)
2. If not, I prefer making the traits implementation explicit at this stage, just for better readability

ee/tabby-webserver/src/schema/email.rs
wsxiaoys: discussed in slack

ee/tabby-webserver/src/service/email.rs
```diff
@@ -76,22 +107,38 @@ pub async fn new_email_service(db: DbConn) -> Result<impl EmailService> {
...
     async fn get_email_setting(&self) -> Result<Option<EmailSetting>> {
         let creds = self.db.read_email_setting().await?;
-        Ok(creds.map(Into::into))
+        let Some(creds) = creds else {
+            return Ok(None);
+        };
+        let creds = creds.try_into();
+        let Ok(creds) = creds else {
+            warn!(
+                "Email settings are corrupt, please inspect or reset them with deleteEmailSetting"
```
wsxiaoys, line 116: let's delete the record on behalf of user. (call self.db.delete_email_setting here)

ee/tabby-db/src/conversions.rs
```diff
@@ -0,0 +1,28 @@
+pub trait DatabaseSerializable: Sized {
+    fn as_db_str(&self) -> &'static str;
```
wsxiaoys:
```suggestion
    fn as_enum_str(&self) -> &'static str;
```

ee/tabby-db/src/conversions.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn from_db_str(s: &str) -> anyhow::Result<Self>;
```
wsxiaoys:
```suggestion
    fn from_enum_str(s: &str) -> anyhow::Result<Self>;
```

wsxiaoys (changes requested):

ee/tabby-db/src/conversions.rs
boxbeam: I could certainly do that, we'd just have to define a new crate for it. Maybe we could make a `tabby-macros` crate?

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,55 @@
...
 pub struct EmailSetting {
     pub smtp_username: String,
     pub smtp_server: String,
+    pub from_address: String,
+    pub encryption: Encryption,
+    pub auth_method: AuthMethod,
+}
+
+impl EmailSetting {
+    pub fn new_validate(
```
wsxiaoys, line 31: as discussed, move this to dao.rs so you don't have to include tabby_db herer.

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,55 @@
...
 
 #[derive(GraphQLObject)]
 pub struct EmailSetting {
     pub smtp_username: String,
     pub smtp_server: String,
+    pub from_address: String,
+    pub encryption: Encryption,
+    pub auth_method: AuthMethod,
+}
+
```
wsxiaoys, line 29: add EmailSettingInput oject

wsxiaoys (changes requested):

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,39 @@
...
+#[derive(GraphQLObject, Validate)]
+pub struct EmailSettingInput {
+    #[validate(email)]
```
wsxiaoys, line 31: smtp_username doesn't have to be email address,

example for aws:

image

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -1,11 +1,39 @@
...
+    pub smtp_username: String,
+    #[validate(email)]
+    pub from_address: String,
+    #[validate(url)]
+    pub smtp_server: String,
```
wsxiaoys, line 36: missing smtp_password field (optional), authentication field, ...

ee/tabby-webserver/src/schema/email.rs
```diff
@@ -16,6 +44,9 @@ pub trait EmailService: Send + Sync {
         smtp_username: String,
         smtp_password: Option<String>,
         smtp_server: String,
+        from_address: String,
+        encryption: Encryption,
+        auth_method: AuthMethod,
```
wsxiaoys, line 49:
```suggestion
        input: EmailSettingInput
```

Please follow the example in https://github.com/TabbyML/tabby/pull/1419 on `*Input`

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -468,6 +469,9 @@ impl Mutation {
         smtp_username: String,
         smtp_password: Option<String>,
         smtp_server: String,
+        from_address: String,
+        encryption: Encryption,
+        auth_method: AuthMethod,
```
wsxiaoys, line 474:
```suggestion
        input: EmailSettingInput
```

ditto

ee/tabby-webserver/src/service/dao.rs
```diff
@@ -148,3 +155,66 @@ impl AsID for i32 {
...
+            "plain" => Ok(AuthMethod::Plain),
+            "login" => Ok(AuthMethod::Login),
+            "xoauth2" => Ok(AuthMethod::XOAuth2),
+            _ => Err(anyhow!("{s} is not a valid value for AuthMethod")),
+        }
+    }
+}
+
+impl EmailSetting {
+    pub fn new_validate(
```
wsxiaoys, line 198: I suggest simply moving the content of this function in side of `TryFrom`

ee/tabby-webserver/src/service/email.rs
```diff
@@ -117,13 +179,13 @@ impl EmailService for EmailServiceImpl {
...
-        // TODO: Include invitation link
+        let network_setting = self.db.read_network_setting().await?;
+        let external_url = network_setting.external_url;
         self.send_mail(
             email,
             "You've been invited to join a Tabby workspace!".into(),
             format!("Welcome to Tabby! You have been invited to join a Tabby instance, where you can tap into\
-                AI-driven code completions and chat assistants. Your invite code is {code}, use this at the link\
-                you were given to join the organization."),
+                AI-driven code completions and chat assistants. Your invite code is {code}, go to {external_url} to join!"),
```
wsxiaoys, line 188: Generate the url directly, in format of `{external_url}/auth/signup?invitationCode={code}`

wsxiaoys (changes requested):

ee/tabby-webserver/src/service/email.rs
```diff
@@ -76,23 +111,44 @@ pub async fn new_email_service(db: DbConn) -> Result<impl EmailService> {
...
-        )
-        .await?;
-        // TODO: make from address being configurable in EmailSettings.
-        *self.from.write().await = smtp_username.clone();
+        let input = EmailSettingInput {
+            smtp_username,
+            from_address,
+            smtp_server,
+        };
+        input.validate()?;
```
wsxiaoys, line 140: Please follow the example in https://github.com/TabbyML/tabby/pull/1419 on *Input

validate should be called at graphql api layer to properly construct ValidationErrors

boxbeam: This PR calls it in the service layer: https://github.com/TabbyML/tabby/pull/1419/files#diff-e0b0fc890e043329c44fcced75dee758af00d7637d9bd0a81e93cce1346935d4R17

wsxiaoys (commented):
Otherwise LGTM

PR #1376, opened 2024-02-04
bugfix: com.intellij.serviceContainer.AlreadyDisposedException

moqimoqidea:
I often use Intellij IDEA to open multiple projects, and frequently encounter the following stack errors when using plugins, which cause the entire software to freeze and unable to work.

```txt
ProcessCanceledException must not be thrown from document listeners for real document
   
   java.lang.Throwable: com.intellij.openapi.progress.ProcessCanceledException: com.intellij.serviceContainer.AlreadyDisposedException: Cannot create com.intellij.ui.docking.DockManager because container is already disposed (container=Project(name=my-project, containerState=DISPOSE_COMPLETED, componentStore=/Users/moqi/code/my-project) (disposed))
	at com.intellij.openapi.editor.impl.DocumentImpl$DelayedExceptions.register(DocumentImpl.java:837)
	at com.intellij.openapi.editor.impl.DocumentImpl.lambda$changedUpdate$1(DocumentImpl.java:916)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$executeNonCancelableSection$3(CoreProgressManager.java:222)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:679)
	at com.intellij.openapi.progress.impl.CoreProgressManager.computeUnderProgress(CoreProgressManager.java:635)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$computeInNonCancelableSection$4(CoreProgressManager.java:230)
	at com.intellij.openapi.progress.Cancellation.computeInNonCancelableSection(Cancellation.java:64)
	at com.intellij.openapi.progress.impl.CoreProgressManager.computeInNonCancelableSection(CoreProgressManager.java:230)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeNonCancelableSection(CoreProgressManager.java:221)
	at com.intellij.openapi.editor.impl.DocumentImpl.changedUpdate(DocumentImpl.java:910)
	at com.intellij.openapi.editor.impl.DocumentImpl.updateText(DocumentImpl.java:814)
	at com.intellij.openapi.editor.impl.DocumentImpl.insertString(DocumentImpl.java:545)
	at com.intellij.openapi.editor.actions.EnterAction.insertNewLineAtCaret(EnterAction.java:62)
	at com.intellij.openapi.editor.actions.EnterAction$Handler.executeWriteAction(EnterAction.java:29)
[...]

Issue #917, opened 2023-11-29
Pycharm text types in reverse (


0x26res:
**Describe the bug**

I've installed the TabbyML plugin for PyCharm (v1.1.1).
After a while (several days), when I type anything on my pycharm it is spelled backward, because as I type the cursor moves backward.
The only solution is to disable the plugin and restart.

**Information about your version**

I'm using:
- `tabby 0.3.1` (though I don't think it matters).
- PyCharm 2023.2.5
- v1.1.1 of the plugin

**Information about your GPU**
Not applicatble

**Additional context**

I think other jetbrains plugin have got this issue https://intellij-support.jetbrains.com/hc/en-us/community/posts/8932067863570-Pycharm-text-is-typing-in-reverse-cursor-does-not-move-left-to-right-

icycodes:
Hi, thank you for reporting this bug.   
Could you please provide some steps to help reproduce this problem?  
If it occurs randomly, it would be helpful if you could share your IDE logs. You can locate IDE logs by `Help -> Show Log in Files|Finder...`, or refer to [this document](https://intellij-support.jetbrains.com/hc/en-us/articles/206544519-Directories-used-by-the-IDE-to-store-settings-caches-plugins-and-logs).

r7l:
I was running into this issue as well. Just restarted my PyCharm and and it was working normal again. Can't say what caused it.

0x26res:
@icycodes sadly it's hard to reproduce. I'll turn tabby back on and send you the logs if/when it happens again. But it may take a week.

chrishultin:
I've also seen this issue occur in IntelliJ Ultimate (most recently, today).  If it happens again, I'll try to grab logs.

0x26res:
@icycodes it just happened again.

There is a couple of relevant things in the log:
- an `AlreadyDisposedException` that seems to come from Tabby 
- this error :`Attempt to get line out of bounds: -1 < 0`

```
2023-11-30 09:13:15,643 [86824340] SEVERE - #c.i.o.v.n.AsyncEventSupport - com.intellij.serviceContainer.AlreadyDisposedException: Cannot create com.intellij.ui.docking.DockManager because container is already disposed (container=Project(name=jetpump2, containerState=DISPOSE_COMPLETED, componentStore=/Users/aandres/source/jetpump2) (disposed))
[...]

PR #1389, opened 2024-02-06
fix(intellij): com.intellij.serviceContainer.AlreadyDisposedException

moqimoqidea:
see #1376

### DIFF
--- a/clients/intellij/src/main/kotlin/com/tabbyml/intellijtabby/editor/EditorListener.kt
+++ b/clients/intellij/src/main/kotlin/com/tabbyml/intellijtabby/editor/EditorListener.kt
@@ -40,19 +40,32 @@ class EditorListener : EditorFactoryListener {
     editor.document.addDocumentListener(object : DocumentListener {
       override fun documentChanged(event: DocumentEvent) {
         logger.debug("DocumentListener: documentChanged $event")
-        if (editorManager.selectedTextEditor == editor) {
-          if (settings.completionTriggerMode == ApplicationSettingsState.TriggerMode.AUTOMATIC) {
-            completionProvider.ongoingCompletion.value.let {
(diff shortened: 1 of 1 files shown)


icycodes:
Thank you for the PR!
I will validate it in my environment.

moqimoqidea:
> Thank you for the PR!
> 
> I will validate it in my environment.

Hello, has this pr verification been successful?😄

icycodes:
> > Thank you for the PR!
> > I will validate it in my environment.
> 
> Hello, has this pr verification been successful?😄

After some testing in my environment, I found that the main problem is that the attached `DocumentListener` should not be invoded after the related editor is disposed. I believe we should remove the listener when the editor is released. This way, we will not need to always catch this error.

Additionally, I have found that this problem is related to issue #917. Both issues are caused by a `ContainerDisposedException`. Have you observed any behavior like reverse text typing mentioned in #917, or just a UI freeze?

moqimoqidea:
> > > Thank you for the PR!
> > > I will validate it in my environment.
> > 
> > 
> > Hello, has this pr verification been successful?😄
> 
> After some testing in my environment, I found that the main problem is that the attached `DocumentListener` should not be invoked after the related editor is disposed. I believe we should remove the listener when the editor is released. This way, we will not need to always catch this error.
> 
> Additionally, I have found that this problem is related to issue #917. Both issues are caused by a `ContainerDisposedException`. Have you observed any behavior like reverse text typing mentioned in #917, or just a UI freeze?

Yes, I have encountered the issue of reverse input in my use case, but the frequency is not high compared to freeze.

Thank you very much for your opinion. I just did a bugfix based on the error message from freeze combined with gpt4, and then verified that neither freeze nor reverse input occurred again.

If you have a more reasonable bugfix plan, I think the new plan is better. Haha, because I just solved the problem I encountered without considering a more elegant design at the IDE level.

PR #1390, opened 2024-02-06
refactor(hub): extract SchedulerClient for scheduler

wsxiaoys:
cc @boxbeam


.github/workflows/test-rust.yml
```diff
@@ -32,6 +32,9 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Service shouldn't depends on juniper::ID
```
wsxiaoys, line 35: cc @boxbeam

I add this lint to avoid polluting service layer with `ID`

PR #1391, opened 2024-02-06
feat: Support AMD GPU in docker image

KagurazakaNyaa:
ROCm and CUDA are not suitable for building in the same docker image, so I separated 3 different device types.
Because I don't have a macos device to build and test locally so I didn't add the `metal` type.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,15 +3,15 @@ name: Create and publish docker image
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 20 */1 * *'
+    - cron: "0 20 */1 * *"
   push:
     tags:
-      - 'v*'
-      - '!*-dev'
+      - "v*"
+      - "!*-dev"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }} 
-  
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+
   # If this is enabled it will cancel current running and start latest
   cancel-in-progress: true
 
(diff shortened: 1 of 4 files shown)


wsxiaoys:
Thanks for the PR. Have you checked our linux binary distribution? It probably an easier way to utilize tabby on linux systems.

KagurazakaNyaa:
> Thanks for the PR. Have you checked our linux binary distribution? It probably an easier way to utilize tabby on linux systems.

Yes, I understand that using the binary distribution directly is an easy way, but then you have to manually manage updates and write systemd units, install rocm or cuda runtimes. Maybe these troubles can be avoided after this project enters the AUR/Debian repository/PPA, but currently the most convenient way is to run the docker container.

wsxiaoys:
Right - I was primarily considering that it could be much easier to create a Docker image for rocm/cpu since we have binary distributions as first-class release artifacts. You could simply download the binary into the Docker container.

I would still maintain the CUDA build as it is for legacy reasons. There are existing systems that rely on the current build behavior for deploying Tabby.

KagurazakaNyaa:
> Right - I was primarily considering that it could be much easier to create a Docker image for rocm/cpu since we have binary distributions as first-class release artifacts. You could simply download the binary into the Docker container.
> 
> I would still maintain the CUDA build as it is for legacy reasons. There are existing systems that rely on the current build behavior for deploying Tabby.

Indeed, maybe I should modify the build method so that the builder step does not build the binary file from the source code but downloads the binary file directly from the GitHub release based on the tag?

wsxiaoys:
One way to do so, is to trigger docker build after https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release event, instead of tag push?

KagurazakaNyaa:
A strange situation occurred. When I used the binary file directly, it seemed that I could not correctly call amdgpu for calculations.
Maybe I should open a new issue to describe this problem.

KagurazakaNyaa:
ref #1404 
I submitted the method of using download instead of building to the download branch, maybe I can test the results on more devices.

wsxiaoys:
close in favor of #1543

PR #1405, opened 2024-02-07
feat(ui, webserver): add general settings

liangfung:
https://jam.dev/c/0213fd14-1aab-45d6-8e69-92a0ab9d741d


ee/tabby-ui/app/(dashboard)/settings/general/components/security-form.tsx
```diff
@@ -0,0 +1,194 @@
...
+                </FormDescription>
+              </FormItem>
+            )}
+          />
+          <div>
+            {fields.map((field, index) => (
+              <FormField
+                control={form.control}
+                key={field.id}
+                name={`domainList.${index}.value`}
```
liangfung, line 134: @wsxiaoys  Due to the implementation of 'useFieldArray', there is a data structure transformation here, and the form path will look like 'domainList.0.value'

PR #1406, opened 2024-02-07
refactor(webserver): move scheduler cron to webserver

wsxiaoys:


ee/tabby-webserver/src/cron/scheduler.rs
```diff
@@ -0,0 +1,89 @@
+use std::{process::Stdio, sync::Arc, time::Duration};
+
+use tokio::io::AsyncBufReadExt;
+use tokio_cron_scheduler::Job;
+use tracing::{info, warn};
+
+use crate::schema::{job::JobService, worker::WorkerService, ServiceLocator};
+
+pub async fn scheduler_job(
```
boxbeam, line 9: This function is long and does many things. I think it should be split up, particularly building the command and piping process output.

crates/tabby-scheduler/src/lib.rs
```diff
@@ -5,101 +5,39 @@ mod index;
...
-use tokio::io::AsyncBufReadExt;
 use tokio_cron_scheduler::{Job, JobScheduler};
 use tracing::{error, info, warn};
 
-pub async fn scheduler<T: RepositoryAccess + 'static>(
-    now: bool,
-    access: T,
-    args: &[String],
-) -> Result<()> {
+pub async fn scheduler<T: RepositoryAccess + 'static>(now: bool, access: T) -> Result<()> {
```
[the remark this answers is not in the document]
wsxiaoys: it no longer have access to job runs

ee/tabby-webserver/src/cron/scheduler.rs
wsxiaoys: ack, should be better to do as a followup pr

PR #1407, opened 2024-02-07
refactor(webserver): ensure only services could rely on tabby_db dire…

wsxiaoys:
…ctly

### DIFF
--- a/.github/workflows/test-rust.yml
+++ b/.github/workflows/test-rust.yml
@@ -33,7 +33,7 @@ jobs:
           submodules: recursive
 
       - name: Service shouldn't depends on juniper::ID
-        run: bash -c '! grep "juniper::ID" -r ./ee/tabby-webserver/src/service'
+        run: bash -c '! grep "juniper::ID" -r ./ee/tabby-webserver/src/service --exclude ./ee/tabby-webserver/src/service/dao.rs'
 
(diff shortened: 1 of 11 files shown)


.github/workflows/test-rust.yml
```diff
@@ -33,7 +33,10 @@ jobs:
           submodules: recursive
 
       - name: Service shouldn't depends on juniper::ID
-        run:  bash -c '! grep "juniper::ID" -r ./ee/tabby-webserver/src/service'
+        run: bash -c '! grep "juniper::ID" -r ./ee/tabby-webserver/src/service'
+
+      - name: Only service could depends on tabby_db
+        run: bash -c '! grep "tabby_db::" -r ./ee/tabby-webserver/src --exclude-dir ./ee/tabby-webserver/src/service'
```
wsxiaoys, line 39: Not sure if this is the best way to do lints like this, let me know if you know a better solution @boxbeam

boxbeam (approved):

.github/workflows/test-rust.yml
boxbeam: Seems reasonable enough to me

wsxiaoys: endup using https://github.com/ast-grep/ast-grep as lint

PR #1408, opened 2024-02-07
fix(webserver): Fix update_email_setting stack overflow

boxbeam:


wsxiaoys:
add a test case to cover the invoke of `update_email_setting`? good opportunity to test out behavior around optional smtp_password

boxbeam:
I'll add the test, but note that none of the other services have tests right now; all of our testing is done in the db layer.

wsxiaoys:
> all of our testing is done in the db layer.
There're a few tests in service layer, e.g `auth`

> I'll add the test, but note that none of the other services have tests right now; 
General guideline is adding test when encounter a bug (in this case, stack overflow) to avoid future regression.

PR #1420, opened 2024-02-09
feat(webserver): for oauth, use callback url derived from `external_url`

wsxiaoys:


wsxiaoys:
cc @boxbeam

ee/tabby-db/migrations/0011_remove-redirect-uri-field-oauth.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE google_oauth_credential ADD COLUMN redirect_uri VARCHAR(256);
```
boxbeam, line 1: Please use 12, 11 is already used in the new email settings

boxbeam (changes requested):

ee/tabby-db/migrations/0011_remove-redirect-uri-field-oauth.down.sql
wsxiaoys: sure - will wait for #1403

PR #1421, opened 2024-02-09
refactor(scheduler): extract run_scheduler_now to handle scheduler er…

wsxiaoys:
…ror, instead of unwrapping

### DIFF
--- a/ee/tabby-webserver/src/cron/scheduler.rs
+++ b/ee/tabby-webserver/src/cron/scheduler.rs
@@ -1,8 +1,9 @@
-use std::{process::Stdio, sync::Arc, time::Duration};
+use std::{process::Stdio, sync::Arc};
 
+use anyhow::{Context, Result};
 use tokio::io::AsyncBufReadExt;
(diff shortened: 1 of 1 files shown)


ee/tabby-webserver/src/cron/scheduler.rs
```diff
@@ -86,3 +39,61 @@ pub async fn scheduler_job(
...
+            }
+        });
+    }
+
+    {
+        // Pipe stderr
+        let stderr = child.stderr.take().context("Failed to acquire stderr")?;
+        let job = job.clone();
+        let job_id = job_id.clone();
+        tokio::spawn(async move {
```
boxbeam, line 83: very similar to the previous block, would it make sense to make this a function?

wsxiaoys: not really something can be easily done:
1. stderr / stdout have different type.
2. they use eprintln / println differently.
3. they use update_job_stdout / update_job_stderr differently.

boxbeam (approved):

PR #1423, opened 2024-02-09
feat(db): Improve error message for duplicate repository creation

boxbeam:


wsxiaoys (changes requested):
Add a test case in service to cover this?

PR #1433, opened 2024-02-10
refactor(email): switch email send to backend

wsxiaoys:
cc @boxbeam


wsxiaoys:
TLS

image

StartTLS
image

Both success

PR #1437, opened 2024-02-11
feat(core, webserver): add /v1beta/server_setting api

wsxiaoys:


ee/tabby-webserver/src/handler.rs
```diff
@@ -72,3 +77,19 @@ async fn distributed_tabby_layer(
...
+    let setting = match locator.setting().read_security_setting().await {
+        Ok(x) => x,
+        Err(err) => {
+            warn!("Failed to read security setting {}", err);
+            return Err(StatusCode::INTERNAL_SERVER_ERROR);
+        }
+    };
+
+    Ok(Json(ServerSetting {
+        disable_client_side_telemetry: setting.disable_client_side_telemetry,
```
boxbeam, line 93: If this is to convert the DAO, don't we have a `From` impl for it?

boxbeam (approved):
Otherwise LGTM

ee/tabby-webserver/src/handler.rs
wsxiaoys: mostly not wanna create a practice that whatever in `ServerSetting` should be stored in `ServerSetting` table.

They should be somewhat irrelevant concept, but unfortunately reused the name :/

boxbeam: Just concerned about the number of places we might need to duplicate this logic if we won't be using into() - do you mean you want to decouple the DAO and external representation more?

PR #1440, opened 2024-02-12
chore(db): Improve unique violation error messages

boxbeam:


ee/tabby-db/src/invitations.rs
```diff
@@ -69,18 +70,13 @@ impl DbConn {
...
-                code,
-                created_at: "".into(),
-            }),
-        }
+        let res = res.unique_error("Failed to create invitation, email already exists")?;
+        Ok(InvitationDAO {
+            id: res.last_insert_rowid() as i32,
+            email,
+            code,
+            created_at: "".into(),
```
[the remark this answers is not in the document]
boxbeam: Does the created_at need to be accurate to millisecond precision? If not, it probably makes more sense to use the system time for this.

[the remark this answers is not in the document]
wsxiaoys: still need to make it align with whatever stored in db?

ee/tabby-db/src/invitations.rs
```diff
@@ -60,27 +61,32 @@ impl DbConn {
             return Err(anyhow!("User already registered"));
         }
 
+        let mut transaction = self.pool.begin().await?;
         let code = Uuid::new_v4().to_string();
         let res = query!(
             "INSERT INTO invitations (email, code) VALUES (?, ?)",
```
wsxiaoys, line 67:
```suggestion
            "INSERT INTO invitations (email, code, created_at) VALUES (?, ?, ?)",
```

You can insert created_at with `system::now` to avoid an additional query

boxbeam: Tried this, but the issue is that `created_at` is a string, whose formatting is dependent on SQLite. I wanted to ensure that the format will be the same when you're creating vs requesting the invitation.

wsxiaoys: created_at is not a string, it's an timestamp in db.

the DAO defines it wrongly, take a look at https://github.com/TabbyML/tabby/blob/009f467902bed56b3406a5593b94cb23cd1504fd/ee/tabby-db/src/users.rs#L11 for example

boxbeam: It's a string in InvitationDAO, I can change it to be DateTime instead.

PR #1444, opened 2024-02-12
fix(db): inactive user's auth token should be rejected

wsxiaoys:


ee/tabby-db/src/users.rs
```diff
@@ -129,9 +129,12 @@ impl DbConn {
 
     pub async fn verify_auth_token(&self, token: &str) -> Result<String> {
         let token = token.to_owned();
-        let email = query_scalar!("SELECT email FROM users WHERE auth_token = ?", token)
-            .fetch_one(&self.pool)
-            .await;
+        let email = query_scalar!(
+            "SELECT email FROM users WHERE auth_token = ? and active",
```
boxbeam, line 133:
```suggestion
            "SELECT email FROM users WHERE auth_token = ? AND active",
```

boxbeam (approved):

PR #1446, opened 2024-02-12
feat(webserver): Prevent disabling owner user

boxbeam:
There is currently no way to update `is_admin` for a user, we will have to remember to apply this same check when we get to that.

### DIFF
--- a/ee/tabby-webserver/src/service/auth.rs
+++ b/ee/tabby-webserver/src/service/auth.rs
@@ -430,6 +430,9 @@ impl AuthenticationService for DbConn {
     }
 
     async fn update_user_active(&self, id: &ID, active: bool) -> Result<()> {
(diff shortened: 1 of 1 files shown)


ee/tabby-webserver/src/service/auth.rs
```diff
@@ -430,6 +430,9 @@ impl AuthenticationService for DbConn {
     }
 
     async fn update_user_active(&self, id: &ID, active: bool) -> Result<()> {
+        if id.as_rowid()? == 1 {
+            return Err(anyhow!("Cannot toggle the active status of the owner user"));
```
wsxiaoys, line 434:
```suggestion
            return Err(anyhow!("Cannot toggle the active status of the owner of this instance"));
```

wsxiaoys (approved):

PR #1447, opened 2024-02-13
feat(webserver): Create template system for emails

boxbeam:


ee/tabby-webserver/src/service/email/templates/invitation.html
wsxiaoys, line 1: Could we move it to `tabby_webserver/email_templates` - just for higher viability?

wsxiaoys (approved):
Otherwise LGTM

PR #1450, opened 2024-02-13
refactor(ui): improve developer experience of tabby-ui, use TABBY_FRO…

wsxiaoys:
…NTEND_URI to connect to local development server


wsxiaoys:
cc @boxbeam

PR #1451, opened 2024-02-13
fix(webserver): Remove redundant send_invitation_email

boxbeam:


ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -349,20 +348,6 @@ impl Mutation {
     async fn create_invitation(ctx: &Context, email: String) -> Result<ID> {
         check_admin(ctx)?;
         let invitation = ctx.locator.auth().create_invitation(email.clone()).await?;
-        let email_sent = ctx
-            .locator
-            .email()
-            .send_invitation_email(email, invitation.code)
-            .await;
-        match email_sent {
-            Err(SendEmailError::NotConfigured) | Ok(_) => {}
```
wsxiaoys, line 358: guess you'll need to handle `NotConfigured` inside of `create_invitation`?

PR #1453, opened 2024-02-14
refactor: extract Request management fully into C++

wsxiaoys:


crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -111,7 +134,10 @@ size_t weighted_random(const float* nums, size_t len, uint64_t seed) {
   for (i = 0; i < len; i++) {
     sum += nums[i];
     if (sum >= random) {
-      return i;
+      const auto token_str = llama_token_to_piece(ctx, i);
+      if (req->check_candidate(token_str)) {
```
wsxiaoys, line 138: @boxbeam could you help confirm if this approach is making sense from distribution perspective?

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -256,15 +280,16 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
...
+        compute_softmax_inplace(logits, n_vocab, request.temperature());
+        const llama_token next_token = weighted_random(ctx_.get(), &request, logits, n_vocab);
         request.n_past += request.tokens.size();
 
         request.tokens.clear();
         request.tokens.push_back(next_token);
 
         const auto token_str = llama_token_to_piece(ctx, next_token);
         request.generated_text += token_str;
+        request.accept_candidate(token_str);
```
wsxiaoys, line 292: @boxbeam when an token got selected, we accept it at end.

crates/llama-cpp-bindings/src/llama.rs
```diff
@@ -1,104 +1,87 @@
...
+
+    pub(crate) fn temperature(&self) -> f32 {
+        self.temperature
+    }
+
+    pub(crate) fn check_candidate(&self, _candidate: &[u8]) -> bool {
+        true
+    }
+
+    pub(crate) fn accept_candidate(&self, _candidate: &[u8]) {}
```
wsxiaoys, line 46: @boxbeam for now they're dummy implementation, but should provide sufficient flexibility to integrate a treesitter context

crates/llama-cpp-bindings/src/engine.cc
boxbeam: No, I don't think this makes sense. This means that weighted_random will keep going and select the next token after this one if the check fails. So if the first token selected doesn't work, it'll select whatever the next one is, regardless of its weight, as long as it passes `check_candidate`. We most likely want to perform this logic outside `weighted_random`, repeatedly calling it and checking the output token until we find a token we like.

crates/llama-cpp-bindings/src/engine.cc
boxbeam: This is where I think we should have a loop to get new tokens until it finds one that works better. I think giving it a cap on how many attempts it can take (3-10 might be a good default) would be a good idea.

crates/llama-cpp-bindings/src/lib.rs
```diff
@@ -17,6 +17,20 @@ mod ffi {
...
+    extern "Rust" {
+        type LlamaInitRequest;
+        fn id(&self) -> u32;
+        fn prompt(&self) -> &str;
+        fn max_input_length(&self) -> usize;
+        fn seed(&self) -> u64;
+        fn temperature(&self) -> f32;
+
+        fn check_candidate(&self, candidate: &[u8]) -> bool;
+        fn accept_candidate(&self, candidate: &[u8]);
```
boxbeam, line 29: So `check_candidate` is likely where we'd want to do be performing the tree-sitter check. Thank you for adding this logic, I get a bit more daunted by C++ code. With these interfaces, it should be pretty easy to add tree-sitter integration.

crates/llama-cpp-bindings/src/llama.rs
```diff
@@ -126,13 +109,25 @@ fn start_llama_service_impl(
...
+            tx,
+            next_request_id: RwLock::default(),
+        }
+    }
+
+    async fn alloc_request_id(&self) -> u32 {
+        let ret = *self.next_request_id.read().await;
+
+        // 1048576 (2^20) should be large enough to avoid any collision.
+        *self.next_request_id.write().await = (ret + 1) % 1048576;
```
boxbeam, line 129: The comment doesn't explain the purpose of this code clearly enough, why are you adding 1 and performing modulus?

boxbeam (changes requested):

crates/llama-cpp-bindings/src/llama.rs
wsxiaoys: request_id represents an unique request. It shouldn't being conflicting with requests that being concurrently processed.

In tabby's case, since it's unlikely it can handle 2^20 requests in parallel, thus 2^20 modulus should be good enough to generate unique request ids.

boxbeam:
```suggestion
        // request_id must be unique at time of processing and shouldn't overflow, tabby is unlikely to handle 2^20 requests at once
        *self.next_request_id.write().await = (ret + 1) % 1048576;
```

crates/llama-cpp-bindings/src/engine.cc
wsxiaoys: offline discussed

PR #1456, opened 2024-02-14
feat(db, webserver): Implement password reset flow

boxbeam:
cc @liangfung 

Since this is already a very large PR, I would prefer to add unit tests in a follow-up PR


ee/tabby-db/migrations/0014_password-reset.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE password_reset(
+    email VARCHAR(150) PRIMARY KEY NOT NULL,
+    code VARCHAR(36) NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT (DATETIME('now')),
+    FOREIGN KEY (email) REFERENCES users(email) ON DELETE CASCADE ON UPDATE CASCADE
```
wsxiaoys, line 5: avoid using foreign key, consider use a cron job to delete staled records (like refresh_tokens table)

ee/tabby-db/migrations/0014_password-reset.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE password_reset(
+    email VARCHAR(150) PRIMARY KEY NOT NULL,
+    code VARCHAR(36) NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT (DATETIME('now')),
```
wsxiaoys, line 4: guess you'll need a status field to represent whether it's used?, e.g `consumed`

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -297,6 +313,30 @@ pub struct RequestInvitationInput {
...
+        min = 8,
+        code = "password",
+        message = "Password must be at least 8 characters"
+    ))]
+    #[validate(length(
+        max = 20,
+        code = "password",
+        message = "Password must be at most 20 characters"
+    ))]
+    pub password: String,
```
wsxiaoys, line 337:
```suggestion
  pub code: String,
  pub password1: String,
  pub password2: String,
```

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -218,6 +219,55 @@ impl AuthenticationService for AuthenticationServiceImpl {
         Ok(resp)
     }
 
+    async fn request_password_reset(&self, email: &str) -> Result<()> {
+        self.get_user_by_email(email)
+            .await
+            .map_err(|_| anyhow!("The email specified is not a registered user"))?;
+        let code = self.db.create_password_reset(email.to_string()).await?;
```
wsxiaoys, line 226: certain rate limiting logic need to be handled here.

wsxiaoys (commented):
some comments. I'll suggest splitting a PR for `tabby-db` just for easier of review

ee/tabby-db/migrations/0014_password-reset.up.sql
boxbeam: Curious to know why?

ee/tabby-db/migrations/0014_password-reset.up.sql
boxbeam: The entry is removed from the table once it is used

ee/tabby-webserver/src/schema/auth.rs
boxbeam: Why no validations?

ee/tabby-db/migrations/0014_password-reset.up.sql
wsxiaoys: they're useful in certain scenario (e.g user / user order), but generally not too useful for cases like this (an ephemeral code), especially when combining with cascade deletion.

ee/tabby-webserver/src/service/auth.rs
boxbeam: Any example of this rate limiting elsewhere in the code?

ee/tabby-db/migrations/0014_password-reset.up.sql
boxbeam: Alright, do we want a single job which will handle cleaning up all expired resources, or a separate job for each of them?

ee/tabby-webserver/src/service/auth.rs
wsxiaoys: you can implement it by querying the database, e.g if there's more than 3 record of pass word reset with same email in past 10 min, disallowed the request.

boxbeam: That won't work, as no row will be inserted if the request fails for any reason (including user not existing), and the same user can only have one password reset code at a time.

boxbeam: If we're relying on database operations to determine rate limiting, I think it defeats the purpose of rate limiting, since the rate limit checks themselves will be expensive operations.

wsxiaoys: ok, the rate limiting is the wrong wording here. What i really mean is to have certain limitation to avoid abusing this interface to spam a certain user.

> and the same user can only have one password reset code at a time.
This is not preferred. let's use logic `if there's more than 2 record of pass word reset with same email in past 10 min, disallowed the request` to have more flexibility

boxbeam: I see, I can implement something like that - prevent requesting another reset code if you have requested one in the last 5 minutes?

boxbeam:
> some comments. I'll suggest splitting a PR for `tabby-db` just for easier of review

Done, #1458

ee/tabby-db/migrations/0014_password-reset.up.sql
wsxiaoys: you might just add another function in https://github.com/TabbyML/tabby/blob/3db3d95eaedf7d347d035c1db87727cca46fe191/ee/tabby-webserver/src/cron/db.rs#L11, like `password_reset_job`

ee/tabby-webserver/src/schema/auth.rs
boxbeam: Are you just saying to also add a `password2` field which ensures the two passwords are the same? Can that check not be performed on the frontend?

ee/tabby-db/migrations/0014_password-reset.up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE password_reset(
+    id INTEGER PRIMARY KEY NOT NULL,
+    code VARCHAR(36) NOT NULL,
```
wsxiaoys, line 3: still need a field to reference user table? e.g `user_id: INTEGER`

boxbeam: The `id` in here refers to the user's ID, not a row ID. I can update the column name so that it's more clear.

wsxiaoys: let's have both `id` and `user_id` to the table.

Always have an `id` as primary key for any table.

boxbeam: Curious to know why, if it will never be used? Do we want to allow the same user to have multiple active password reset codes at once (potentially confusing)? And if not, should the primary key then be `(id, user_id)`?

wsxiaoys: https://stackoverflow.com/questions/1207983/in-general-should-every-table-in-a-database-have-an-identity-field-to-use-as-a

The primary key should be id. you can add unique constraint to `user_id`.

ee/tabby-webserver/src/schema/auth.rs
wsxiaoys: > Why no validations?
I'll leave validations to your for implementation, just listing field for reference :)

> Are you just saying to also add a password2 field which ensures the two passwords are the same? Can that check not be performed on the frontend?
Yes, see [RegisterInput](https://github.com/TabbyML/tabby/blob/3db3d95eaedf7d347d035c1db87727cca46fe191/ee/tabby-webserver/src/service/auth.rs#L37) as reference. (in general, I suggest always following practices in components doing similar functionalities in repo)

ee/tabby-db/src/password_reset.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[derive(FromRow)]
+pub struct PasswordResetDAO {
+    pub user_id: i32,
+    pub code: String,
+    pub created_at: DateTime<Utc>,
+}
+
+impl DbConn {
+    pub async fn create_password_reset(&self, id: i32) -> Result<String> {
```
wsxiaoys, line 16:
```suggestion
    pub async fn create_password_reset(&self, user_id: i32) -> Result<String> {
```

please also update all `id` occurrence actually means `user_id`

ee/tabby-db/src/password_reset.rs
```diff
@@ -0,0 +1,55 @@
...
+            id,
+            code,
+            time
+        )
+        .execute(&self.pool)
+        .await?;
+        Ok(code)
+    }
+
+    pub async fn delete_password_reset(&self, id: String) -> Result<()> {
```
wsxiaoys, line 31:
```suggestion
    pub async fn delete_password_reset_by_user_id(&self, user_id: String) -> Result<()> {
```

please also update methods that's `by_user` id with such suffix.

ee/tabby-db/src/password_reset.rs
```diff
@@ -0,0 +1,55 @@
...
+        let password_reset = sqlx::query_as(
+            "SELECT user_id, code, created_at FROM password_reset WHERE user_id = ?;",
+        )
+        .bind(id)
+        .fetch_optional(&self.pool)
+        .await?;
+        Ok(password_reset)
+    }
+
+    pub async fn delete_expired_password_reset(&self) -> Result<()> {
```
wsxiaoys, line 48:
```suggestion
    pub async fn delete_expired_password_resets(&self) -> Result<()> {
```

wsxiaoys (changes requested):

ee/tabby-db/src/password_reset.rs
```diff
@@ -13,7 +13,7 @@ pub struct PasswordResetDAO {
 }
 
 impl DbConn {
-    pub async fn create_password_reset(&self, user_id: i32) -> Result<String> {
+    pub async fn create_password_reset_for_user_id(&self, user_id: i32) -> Result<String> {
```
wsxiaoys, line 16:
```suggestion
    pub async fn create_password_reset(&self, user_id: i32) -> Result<String> {
```

wsxiaoys (commented):
Otherwise LGTM

wsxiaoys (approved):

PR #1458, opened 2024-02-14
feat(webserver): Implement password reset flow

boxbeam:


ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -297,6 +313,46 @@ pub struct RequestInvitationInput {
...
 }
 
+#[derive(Validate, GraphQLInputObject)]
+pub struct RequestPasswordResetInput {
+    #[validate(email(code = "email"))]
+    pub email: String,
+}
+
+#[derive(Validate, GraphQLInputObject)]
+pub struct PasswordResetInput {
```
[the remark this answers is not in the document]
boxbeam: It is:
image

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -297,6 +313,46 @@ pub struct RequestInvitationInput {
...
+    #[validate(email(code = "email"))]
+    pub email: String,
```
wsxiaoys, line 325:
```suggestion
```

Email shouldn't be required for password reset (it's a token authorize the user_id anyway)

boxbeam: That's not true; the user cannot be authorized if they are resetting their password. It needs to be an unauthenticated endpoint.

wsxiaoys: this is more of a best practice. Please implement it as I described. (it's also the way django-auth-graphql does)

wsxiaoys (changes requested):
Please add test to increase coverage. Otherwise LG

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -142,6 +142,22 @@ pub enum TokenAuthError {
...
+#[derive(Error, Debug)]
+pub enum PasswordResetError {
+    #[error("Invalid code")]
+    InvalidCode,
+    #[error("Code is expired")]
+    ExpiredCode,
+    #[error("User does not exist")]
+    InvalidEmail,
+    #[error("Invalid password")]
+    InvalidPassword(#[from] ValidationErrors),
```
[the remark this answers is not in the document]
wsxiaoys: This is not done. IntoFieldErrors needs to be implemented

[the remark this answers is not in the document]
boxbeam: Sorry about that, fixed

wsxiaoys (commented):
In general, for all added function that can be tested at service level, let's do best effort to cover all branches.

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -218,6 +219,54 @@ impl AuthenticationService for AuthenticationServiceImpl {
         Ok(resp)
     }
 
+    async fn request_password_reset_email(&self, email: String) -> Result<()> {
```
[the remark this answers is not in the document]
boxbeam: I had to make it ignore the error where the email service is not configured for this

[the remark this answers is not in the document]
wsxiaoys: we actually want get an error if email service is not configured.

I think we have build up a way to config email service in unit test already? can you reuse it?

ee/tabby-webserver/src/service/email/test_utils.rs
```diff
@@ -0,0 +1,43 @@
+use std::time::Duration;
+
+use tabby_db::DbConn;
+use tokio::process::{Child, Command};
+
+use crate::schema::email::{AuthMethod, EmailService, EmailSettingInput, Encryption};
+
+use super::new_email_service;
+
```
wsxiaoys, line 8: Consider extract this into a struct like
```
struct TestEmailServer;

impl TestEmailServer {
  fn new();
  fn start();
  fn kill();
}

...
```

ee/tabby-db/src/password_reset.rs
```diff
@@ -48,9 +57,21 @@ impl DbConn {
         Ok(password_reset)
     }
 
+    pub async fn mark_password_reset_expired(&self, code: &str) -> Result<()> {
```
wsxiaoys, line 60: this is only for testing? mark it with `#[cfg(test]`

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -218,6 +220,53 @@ impl AuthenticationService for AuthenticationServiceImpl {
...
+        };
+
+        let user_res = self.db.get_user(password_reset.user_id).await;
+        let Ok(Some(user @ UserDAO { active: true, .. })) = user_res else {
+            return Err(PasswordResetError::InvalidCode);
+        };
+
+        if Utc::now().signed_duration_since(password_reset.created_at) > Duration::minutes(15) {
+            return Err(PasswordResetError::InvalidCode);
+        }
```
wsxiaoys, line 261: consider extract these into `fn verify_reset_password_code(code) -> i64` into `DbConn`, thus the expiration time related to password reset will be in single place (otherwise user won't know how long it become expire by looking at tabby-db/src/password_reset.rs.

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -218,6 +220,53 @@ impl AuthenticationService for AuthenticationServiceImpl {
         Ok(resp)
     }
 
+    async fn request_password_reset_email(&self, email: String) -> Result<String> {
```
wsxiaoys, line 223:
```suggestion
    async fn request_password_reset_email(&self, email: String) -> Result<()> {
```

If you need this from test, consider read it from email api.

ee/tabby-webserver/src/service/email/test_utils.rs
boxbeam: Looking at it now, not sure this would benefit much. We always start it immediately after creating it, and it's killed on drop, so it would kind of boil down to a single function. I'll try to improve the structure of these functions.

ee/tabby-db/src/password_reset.rs
boxbeam: Tried this, but it becomes inaccessible from the webserver crate even in `#[cfg(test)]` blocks when I do this.

wsxiaoys: You probably need https://github.com/TabbyML/tabby/blob/76f733b216b996cec6f5dad9ae69da20d2bd60f8/ee/tabby-db/src/lib.rs#L225

ee/tabby-webserver/src/service/email/test_utils.rs
wsxiaoys: it could also bundle the methods for checking email / title / content / ...

wsxiaoys: SGTM!

wsxiaoys (approved):

PR #1460, opened 2024-02-15
feat(webserver): add ids argument to jobRuns graphql api

wsxiaoys:


ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -189,6 +189,7 @@ impl Query {
 
     async fn job_runs(
         ctx: &Context,
+        ids: Option<Vec<ID>>,
```
wsxiaoys, line 192: @boxbeam fyi, seems this is the preferable way to do a filtering in graphql

liangfung (approved):

PR #1469, opened 2024-02-16
refactor(db): Convert PasswordResetDAO to be able to use query_as!

boxbeam:


wsxiaoys (commented):
Otherwise LGTM

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -247,7 +247,9 @@ impl AuthenticationService for AuthenticationServiceImpl {
             password_hash(password).map_err(|_| PasswordResetError::Unknown)?;
 
         let user_id = self.db.verify_password_reset(code).await?;
```
wsxiaoys, line 249: verify_password_reset should returns i64 as well?

boxbeam: Already does? https://github.com/TabbyML/tabby/pull/1469/commits/d1aae1c8c444fac07698939ebf7b27ff2746c3f8

wsxiaoys: then why it need another cast in 251?

boxbeam: That's redundant, I suppose

PR #1470, opened 2024-02-16
feat(webserver): Add endpoint to check if email is configured

boxbeam:
cc @liangfung

### DIFF
--- a/ee/tabby-webserver/src/schema/mod.rs
+++ b/ee/tabby-webserver/src/schema/mod.rs
@@ -218,7 +218,7 @@ impl Query {
         Ok(val)
     }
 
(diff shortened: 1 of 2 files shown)


ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -218,6 +218,11 @@ impl Query {
         Ok(val)
     }
 
+    async fn email_configured(ctx: &Context) -> Result<bool> {
```
wsxiaoys, line 221:
```suggestion
    async fn is_email_configured(ctx: &Context) -> Result<bool> {
```

PR #1474, opened 2024-02-17
Add openapi build to support Intel Scalable CPU (#631)

hungle-i3:
* Support new feature: openapi
    * Change compiler to Intel llvm when compiling llama.cpp
    * Support Intel BLAS (Intel10_64lp)
    * Add openapi.Dockerfile 

Built docker and test on Intel Xeon 8480+ with model Mistral 7B.

### DIFF
--- a/ee/tabby-ui/app/(dashboard)/experiments/components/feature-list.tsx
+++ b/ee/tabby-ui/app/(dashboard)/experiments/components/feature-list.tsx
@@ -1,0 +1,29 @@
+'use client'
+
+import { useEnableCodeBrowserQuickActionBar } from '@/lib/experiment-flags'
+import { Switch } from '@/components/ui/switch'
+
+export default function FeatureList() {
+  const [quickActionBar, toggleQuickActionBar] =
+    useEnableCodeBrowserQuickActionBar()
+  return (
+    <>
+      {!quickActionBar.loading && (
+        <div className="flex items-center space-x-4 rounded-md border p-4">
+          <div className="flex-1 space-y-1">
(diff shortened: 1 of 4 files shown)


cromefire:
The feature should be intelmkl or so then right? As oneAPI would kinda suggest it's supporting all the openAPI stuff like GPU support which llama.cpp doesn't do.

hungle-i3:
> The feature should be intelmkl or so then right?
Right. I am scoping this to intelmkl blas only with  current tabby fork of llama.cpp. 

OpenCL devices (including all intel GPU) yet to  be supported in this pull request. I have tried to bump llama.cpp locally to latest version for SYCL OpenCL devices support but  seems to be unstable with tabby.

cromefire:
But the SYCL stuff would need to be a different feature, right? Or is it enabled through MKL?

hungle-i3:
> But the SYCL stuff would need to be a different feature, right? Or is it enabled through MKL?

Yes, should be a different feature. As latest upstream llama.cpp implementation, they are also separated.

cromefire:
But the llama.cpp isn't up-to-date still or is it? Otherwise I might just have a look at the Vulkan Backen as that might be an alternative for the time being.

hungle-i3:
I don't know exactly when tabby bump the llama.cpp to latest version for SYCL support. If they could  do it soon then it would be great.

As  my observation on llama.cpp upstream, SYCL support is not only about the configuration but also assembly code port. It would be a room for further optimization.

cromefire:
We might have to wait for another real release of llama that includes that.

PR #1478, opened 2024-02-18
ci(ui): mark missing react hook deps as lint error

wsxiaoys:


wsxiaoys:
@liangfung please fix all int errors in this branch

PR #1489, opened 2024-02-19
feat(db): Add enterprise license to server_setting

boxbeam:


ee/tabby-db/src/server_setting.rs
```diff
@@ -5,6 +5,7 @@ use crate::DbConn;
 
 #[derive(Debug, PartialEq, FromRow)]
 pub struct ServerSettingDAO {
+    enterprise_license: Option<String>,
```
wsxiaoys, line 8:
```suggestion
   billing_enterprise_license: Option<String>,
```

wsxiaoys: Please update other naming correspondinglly

PR #1491, opened 2024-02-19
feat(webserver): Implement license service

boxbeam:


ee/tabby-webserver/src/service/license.rs
```diff
@@ -48,6 +49,51 @@ pub fn validate_license(token: &str) -> Result<LicenseInfo, jwt::errors::ErrorKi
...
 
+fn jwt_timestamp_to_utc(f: f64) -> Result<DateTime<Utc>> {
+    let secs = f as i64;
+    Ok(NaiveDateTime::from_timestamp_millis(secs * 1000)
+        .ok_or_else(|| anyhow!("Timestamp is corrupt"))?
+        .and_utc())
+}
+
+#[async_trait]
+impl LicenseService for DbConn {
```
wsxiaoys, line 60: As discussed previously, shall we use a LicensenServiceImpl(DbConn) to avoid potential method refactoring in future?

ee/tabby-webserver/src/service/license.rs
```diff
@@ -79,4 +125,27 @@ mod tests {
...
+    async fn test_create_delete_license() {
+        let db = DbConn::new_in_memory().await.unwrap();
+        let service: &dyn LicenseService = &db;
+
+        assert!(service
+            .update_license(Some("bad_token".into()))
+            .await
+            .is_err());
+
+        let token = "eyJhbGciOiJSUzUxMiJ9.eyJpc3MiOiJ0YWJieW1sLmNvbSIsInN1YiI6ImZha2VAdGFiYnltbC5jb20iLCJpYXQiOjE3MDUxOTgxMDIsImV4cCI6MTgwNzM5ODcwMiwidHlwIjoiVEVBTSIsIm51bSI6MTB9.vVo7PDevytGw2KXU5E-KMdJBijwOWsD1zKIf26rcjfxa3wDesGY40zuYZWyZFMfmAtBTO7DBgqdWnriHnF_HOnoAEDCycrgoxuSJW5TS9XsCWto-3rDhUsjRZ1wls-ztQu3Gxo_84UHUFwrXe-RHmJi_3w_YO-2L-nVw7JDd5zR8CEdLxeccD47vBrumYA7ybultoDHpHxSppjHlW1VPXavoaBIO1Twnbf52uJlbzJmloViDxoq-_9lxcN1hDN3KKE3crzO9uHK4jjZy_1KNHhCIIcnINek6SBl6lWZw9R88UfdP6uaVOTOHDFbGwv544TSLA_oKZXXntXhldKCp94YN8J4djHim91WwYBQARrpQKiQGP1APEQQdv_YO4iUC3QTLOVw_NMjyma0feVjzHYAap_2Q9HgnxyJfMH-KiH2zaR6BcdOfWV86crO5M0qNoP-XOgy4uU8eE2-PevOKM6uVwYiwoNZL4e9ttH6ratJj0tyqGW_3HYpsVyThzqDPisEz95knsrVL-iagwHRd00l6Mqfwcjbn-gOuUOV9knRIpPvUmfKjjjHgb-JI0qMAIdgeVtwQp0pNqPsKwenMwkpYQH1awfuB_Ia7SyMUNEzTAY8k_J4R6kCZ5XKJ2VTCljd9aJFSZpw-K57reUX1eLc6-Cwt1iI4d23M5UlYjvs";
```
wsxiaoys, line 139: extract const variables with some info, e.g `static EXPIRED_LICENSE_TOKEN = ...`

wsxiaoys (changes requested):

ee/tabby-webserver/src/schema/license.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub r#type: LicenseType,
+    pub status: LicenseStatus,
+    pub seats: i32,
+    pub issued_at: DateTime<Utc>,
+    pub expires_at: DateTime<Utc>,
+}
+
+#[async_trait]
+pub trait LicenseService: Send + Sync {
+    async fn get_license_info(&self) -> Result<Option<LicenseInfo>>;
```
wsxiaoys, line 30:
```suggestion
    async fn read_license_info(&self) -> Result<Option<LicenseInfo>>;
```

Prefer use `read` instead of `get` for singular concepts in system.

ee/tabby-webserver/src/service/license.rs
```diff
@@ -48,6 +49,51 @@ pub fn validate_license(token: &str) -> Result<LicenseInfo, jwt::errors::ErrorKi
...
+            issued_at,
+            expires_at,
+        };
+
+        Ok(Some(license))
+    }
+
+    async fn update_license(&self, license: Option<String>) -> Result<()> {
+        if let Some(license) = &license {
+            validate_license(license).map_err(|e| anyhow!("License is corrupt: {e:?}"))?;
```
wsxiaoys, line 90: reuse logic inside of `get_license_info`, to acquire the `LicenseStatus`, thus generating the proper error message for different status (e.g expired, exceed seat, ...)

ee/tabby-webserver/src/service/license.rs
```diff
@@ -48,6 +49,51 @@ pub fn validate_license(token: &str) -> Result<LicenseInfo, jwt::errors::ErrorKi
...
+}
+
+#[async_trait]
+impl LicenseService for DbConn {
+    async fn get_license_info(&self) -> Result<Option<LicenseInfo>> {
+        let Some(license) = self.read_enterprise_license().await? else {
+            return Ok(None);
+        };
+        let license =
+            validate_license(&license).map_err(|e| anyhow!("License is corrupt: {e:?}"))?;
```
wsxiaoys, line 66: You need to handle the `Expire` error generated by `validate_license` for LicenseStatus

ee/tabby-webserver/src/service/license.rs
```diff
@@ -48,6 +49,51 @@ pub fn validate_license(token: &str) -> Result<LicenseInfo, jwt::errors::ErrorKi
...
+        let expires_at = jwt_timestamp_to_utc(license.exp)?;
+
+        let status = if expires_at < Utc::now() {
+            LicenseStatus::Expired
+        } else {
+            LicenseStatus::Ok
+        };
+
+        let license = LicenseInfo {
+            r#type: LicenseType::TEAM,
```
wsxiaoys, line 78: this should be from `RawLicenseInfo.typ` field.

ee/tabby-webserver/src/service/license.rs
wsxiaoys: That's not the case. You could check the `test_expired_token` unit test, which provides an example on how to handle expire error.

boxbeam: image
This test never gets the license info. The validate function returns an error.

wsxiaoys: ah - so you're worrying about the parsing part.

It can be done by decode the second component of jwt token from base64, then it's just vanilla json. See jwt token specs for details.

ee/tabby-webserver/src/service/license.rs
```diff
@@ -1,20 +1,20 @@
...
 
 #[derive(Debug, Deserialize)]
-#[serde(rename_all = "UPPERCASE")]
-pub enum LicenseType {
-    Team,
-}
-
-#[derive(Debug, Deserialize)]
-pub struct LicenseInfo {
+pub struct RawLicenseInfo {
```
wsxiaoys, line 17: can we use `RawLicenseInfo` directly with some serde annotations / custom deseralizer?

boxbeam: We could, but I honestly prefer to have a separate struct matching the schema of the token perfectly, and a separate function to translate it into the external representation which includes the status, similar to how we use DAOs.

wsxiaoys: Let's use `LicenseInfo` directly. Service is allowed to relies on whatever defined in `schema`

ee/tabby-db/schema.sqlite
wsxiaoys, line 1: any reason this got updated?

ee/tabby-webserver/src/schema/license.rs
```diff
@@ -0,0 +1,33 @@
...
+#[serde(rename_all = "UPPERCASE")]
+pub enum LicenseType {
+    Team,
+}
+
+#[derive(GraphQLEnum, PartialEq, Debug)]
+pub enum LicenseStatus {
+    Ok,
+    Expired,
+    SeatsExceeded,
```
wsxiaoys, line 17:
```suggestion
```

Remove since it's not implemented yet.

wsxiaoys (commented):
Otherwise LG

ee/tabby-db/schema.sqlite
boxbeam: Think it had to be updated since there were conflicts in this file after splitting the PRs. I deleted the file and regenerated it with `cargo sqlx db setup --source ee/tabby-db/migrations`.

ee/tabby-webserver/src/service/license.rs
boxbeam: I think writing a custom deserializer would still be more difficult and less clear than using a struct matching the schema for deserialization followed by a conversion into another type. Would you rather I put RawLicenseInfo somewhere else and abstract the logic so it is never referred to directly outside of a utility module where it's defined?

boxbeam: https://stackoverflow.com/questions/46753955/how-to-transform-fields-during-deserialization-using-serde: It looks like we can deserialize using custom functions without too much difficulty, but this isn't deserializing any data, it's using the data already deserialized in order to generate a new piece of data. That use case doesn't seem as well-supported, so we would need to write a full custom deserializer that allows us to generate the value of `status` based on the expiration date. This would be further complicated when we implement the check for `SeatsExceeded`, since that would require a database call, which I don't think we have a way of providing to the deserializer.

wsxiaoys: For `status`, you could just use annotation `#[serde(skip)]`, and fill it in a later step?

https://serde.rs/field-attrs.html#skip

PR #1493, opened 2024-02-19
Begin implementation of new logic

boxbeam:


crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -107,6 +107,33 @@ std::string string_format(const std::string& format, Args ... args)
 	return std::string(buf.get(), buf.get() + size - 1); // We don't want the '\0' inside
 }
 
+void sorted_insert_in_capacity(std::vector<size_t>& max_vec, size_t num_index, const float* nums) {
```
wsxiaoys, line 110: guess you might wanna use https://en.cppreference.com/w/cpp/algorithm/partial_sort

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -107,6 +107,33 @@ std::string string_format(const std::string& format, Args ... args)
...
+  if (i == max_vec.size()) {
+    return;
+  }
+  for (int j = max_vec.size() - 1; j > i; j--) {
+    max_vec[j] = max_vec[j - 1];
+  }
+  max_vec[i] = num_index;
+}
+    
+std::vector<size_t> nmax(const float* nums, size_t len, size_t n) {
```
wsxiaoys, line 129: need to invoke `check_candidate`?

crates/llama-cpp-bindings/src/engine.cc
```diff
@@ -280,6 +307,11 @@ class TextInferenceEngineImpl : public TextInferenceEngine {
 
         int32_t i_batch = request.i_batch - i;
         float* logits = llama_get_logits_ith(ctx, i_batch);
+            // TODO: Replace this logic with nmax 5 to get indices
+            // then construct an array of logits (floats) from indices
+            // then run softmax on that array
+            // then run weighted_random on that array to get an index
+            // note: we might not actually need nmax since we'll need to check the predicate on each index (logit) first
```
wsxiaoys, line 314: let's get this implemented with top-5 to update the golden test as a first step.

PR #1494, opened 2024-02-19
fix(webserver, license): change the timestamp type to i64

wsxiaoys:


wsxiaoys:
> Can u64 be used? That seems like the most correct type to use here

unix timestamp can be negative: 
image

boxbeam:
Are we ever going to have a license created or expiring before 1970?

PR #1495, opened 2024-02-19
feat(webserver): add license check

wsxiaoys:


ee/tabby-webserver/src/service/auth.rs
```diff
@@ -233,6 +236,12 @@ impl AuthenticationService for AuthenticationServiceImpl {
     }
 
     async fn create_invitation(&self, email: String) -> Result<Invitation> {
+        if !self.license.read_license().await.is_license_valid() {
```
boxbeam, line 239: Do we want to just make a `.is_license_valid()` function on `LicenseService`?

boxbeam (approved):

ee/tabby-webserver/src/service/auth.rs
wsxiaoys: nope - to ensure we always have full license info at the place applying license check.

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -976,4 +1019,14 @@ mod tests {
 
         assert!(service.allow_self_signup().await.unwrap());
     }
+
+    #[tokio::test]
+    #[serial]
```
wsxiaoys, line 1024:
```suggestion
```

PR #1511, opened 2024-02-21
fix(webserver): Fix logic to find index of first duplicate domain

boxbeam:


ee/tabby-webserver/src/schema/setting.rs
```diff
@@ -46,12 +46,21 @@ pub struct NetworkSettingInput {
...
+    }
+    None
+}
+
 fn validate_unique_domains(domains: &[String]) -> Result<(), ValidationError> {
-    let unique: HashSet<_> = domains.iter().collect();
-    if unique.len() != domains.len() {
-        let i = domains.iter().position(|s| unique.contains(s)).unwrap();
+    let duplicate_index = first_duplicate(domains);
+    if let Some(duplicate_index) = duplicate_index {
```
wsxiaoys, line 61: add a test case to cover the brancb

PR #1513, opened 2024-02-22
refactor: cleanup service level error as we're handling input validation at graphql level.

wsxiaoys:


ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -477,6 +424,38 @@ pub trait AuthenticationService: Send + Sync {
...
+fn validate_password(value: &str) -> Result<(), validator::ValidationError> {
+    let make_validation_error = |message: &'static str| {
+        let mut err = validator::ValidationError::new("password1");
+        err.message = Some(Cow::Borrowed(message));
+        Err(err)
+    };
+
+    let contains_lowercase = value.chars().any(|x| x.is_ascii_lowercase());
+    if !contains_lowercase {
+        return make_validation_error("Password should contains at least one lowercase character");
```
boxbeam, line 436:
```suggestion
        return make_validation_error("Password should contain at least one lowercase character");
```

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -477,6 +424,38 @@ pub trait AuthenticationService: Send + Sync {
...
+    }
+
+    let contains_uppercase = value.chars().any(|x| x.is_ascii_uppercase());
+    if !contains_uppercase {
+        return make_validation_error("Password should contains at least one uppercase character");
```
boxbeam, line 441: Same as above, and for all the others

ee/tabby-webserver/src/service/auth.rs
```diff
@@ -669,7 +524,7 @@ mod tests {
             service
                 .token_auth(ADMIN_EMAIL.to_owned(), "12345678".to_owned())
                 .await,
-            Err(TokenAuthError::UserNotFound)
+            Err(_)
```
boxbeam, line 527: nit: maybe prefer `assert!(operation.is_err())`?

boxbeam (changes requested):
Otherwise LGTM

ee/tabby-webserver/src/service/auth.rs
wsxiaoys: ack - leaving it as is just for cleaner diff.

PR #1528, opened 2024-02-23
feat(webserver): add license check in user authentication

wsxiaoys:


ee/tabby-db/src/users.rs
```diff
@@ -279,7 +280,10 @@ mod tests {
 
         // Inactive user's auth token will be rejected.
         conn.update_user_active(new_user.id, false).await.unwrap();
-        assert!(conn.verify_auth_token(&new_user.auth_token).await.is_err());
+        assert!(conn.verify_auth_token(&new_user.auth_token, false).await.is_err());
+
+        // Admin user should pass verification.
```
wsxiaoys, line 285:
```suggestion
        // Owner user should pass verification.
```

boxbeam (approved):

PR #1531, opened 2024-02-24
fix(webserver): use ID in user field

wsxiaoys:


ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -210,20 +210,20 @@ pub struct JWTPayload {
     /// Issued at (as UTC timestamp)
     iat: i64,
 
-    /// User email address
+    /// User id string
```
boxbeam, line 213: It seems very dangerous to me to swap the semantic meaning of this field without changing its type. Are we *sure* that there are no other places in the codebase that this is being used to refer to email? Is there any reason we can't use the `ID` type directly?

For example, I don't see the password reset logic being changed here, but that workflow uses the `sub` field as an email address.

boxbeam (changes requested):

ee/tabby-webserver/src/schema/auth.rs
wsxiaoys: Good idea, will try that

PR #1539, opened 2024-02-25
docs: mention mailtutan dependency for tests

G07cha:
Without it, the standard test suite will fail.


wsxiaoys:
Thanks for the PR!

PR #1552, opened 2024-02-26
fix(ui): display the correct external url

liangfung:
1.  display the correct external url
2. refresh network settings
3. git table truncate

### DIFF
--- a/ee/tabby-ui/lib/hooks/use-network-setting.tsx
+++ b/ee/tabby-ui/lib/hooks/use-network-setting.tsx
@@ -12,10 +12,10 @@ const networkSettingQuery = graphql(/* GraphQL */ `
   }
 `)
 
-const useNetworkSettingQuery = () => useQuery({ query: networkSettingQuery })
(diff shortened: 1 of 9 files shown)


ee/tabby-ui/lib/hooks/use-network-setting.tsx
```diff
@@ -0,0 +1,38 @@
...
+  query NetworkSetting {
+    networkSetting {
+      externalUrl
+    }
+  }
+`)
+
+const useQueryNetworkSetting = () => useQuery({ query: networkSettingQuery })
+
+const useNetworkSetting = () => {
```
wsxiaoys, line 17: why we need both this and useQueryNetworkSetting? Can they be merged?

liangfung: Yes, for the convenience of using certain components, it seems unnecessary to use useNetworkSetting for now

liangfung: done

PR #1553, opened 2024-02-26
feat(webserver): Add authenticated endpoint to update password

boxbeam:
cc @liangfung

### DIFF
--- a/ee/tabby-webserver/src/schema/auth.rs
+++ b/ee/tabby-webserver/src/schema/auth.rs
@@ -244,6 +244,7 @@ pub struct User {
     pub auth_token: String,
     pub created_at: DateTime<Utc>,
     pub active: bool,
+    pub is_password_set: bool,
 }
 
 impl relay::NodeType for User {
@@ -308,33 +309,35 @@ pub struct PasswordResetInput {
 
 #[derive(Validate, GraphQLInputObject)]
 pub struct PasswordUpdateInput {
+    pub old_password: String,
+
     #[validate(length(
         min = 8,
-        code = "password1",
+        code = "new_password1",
         message = "Password must be at least 8 characters"
     ))]
     #[validate(length(
         max = 20,
-        code = "password1",
+        code = "new_password1",
         message = "Password must be at most 20 characters"
     ))]
-    pub password1: String,
+    pub new_password1: String,
     #[validate(length(
         min = 8,
-        code = "password2",
+        code = "new_password2",
         message = "Password must be at least 8 characters"
     ))]
     #[validate(length(
         max = 20,
-        code = "password2",
+        code = "new_password2",
         message = "Password must be at most 20 characters"
     ))]
     #[validate(must_match(
-        code = "password2",
+        code = "new_password2",
         message = "Passwords do not match",
-        other = "password1"
+        other = "new_password1"
     ))]
-    pub password2: String,
+    pub new_password2: String,
 }
 
 #[derive(Debug, Serialize, Deserialize, GraphQLObject)]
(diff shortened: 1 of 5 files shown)


ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -366,6 +367,16 @@ impl Mutation {
         Ok(true)
     }
 
+    async fn update_password(ctx: &Context, input: PasswordUpdateInput) -> Result<bool> {
```
wsxiaoys, line 370:
```suggestion
    async fn password_change(ctx: &Context, input: PasswordUpdateInput) -> Result<bool> {
```

Please also add unit test.

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -301,6 +301,37 @@ pub struct PasswordResetInput {
     pub password2: String,
 }
 
+#[derive(Validate, GraphQLInputObject)]
+pub struct PasswordUpdateInput {
```
wsxiaoys, line 305: add an optional `old_password` field.

The logic would be:
1. If the user doesn't have password set (empty password, registered from oauth signin), they don't need `old_password` to change the password.
2. If user already have password set, they we need to verify the old password.

---

Other TODOs:
1. adding a field `is_password_set` to the userDAO.
2. Aadd an GraphQL query api `me` to returns `UserDAO` for the user, and

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -301,6 +301,37 @@ pub struct PasswordResetInput {
...
+        max = 20,
+        code = "password2",
+        message = "Password must be at most 20 characters"
+    ))]
+    #[validate(must_match(
+        code = "password2",
+        message = "Passwords do not match",
+        other = "password1"
+    ))]
+    pub password2: String,
```
wsxiaoys, line 332: new_password2

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -301,6 +302,39 @@ pub struct PasswordResetInput {
     pub password2: String,
 }
 
+#[derive(Validate, GraphQLInputObject)]
+pub struct PasswordUpdateInput {
+    pub old_password: String,
```
wsxiaoys, line 307:
```suggestion
    pub old_password: Option<String>,
```

as discussed this need to be optional (since for user.is_password_set, there's no old password)

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -386,6 +420,12 @@ pub trait AuthenticationService: Send + Sync {
     async fn reset_user_auth_token(&self, email: &str) -> Result<()>;
     async fn password_reset(&self, code: &str, password: &str) -> Result<()>;
     async fn request_password_reset_email(&self, email: String) -> Result<Option<JoinHandle<()>>>;
+    async fn update_user_password(
+        &self,
+        email: &str,
+        old_password: &str,
```
wsxiaoys, line 426: optional

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -386,6 +420,12 @@ pub trait AuthenticationService: Send + Sync {
     async fn reset_user_auth_token(&self, email: &str) -> Result<()>;
     async fn password_reset(&self, code: &str, password: &str) -> Result<()>;
     async fn request_password_reset_email(&self, email: String) -> Result<Option<JoinHandle<()>>>;
+    async fn update_user_password(
+        &self,
+        email: &str,
```
wsxiaoys, line 425: this probabaly can be changed to ID, but i can take the follow up PR after the id one got merged.

wsxiaoys (approved):

ee/tabby-webserver/src/schema/auth.rs
boxbeam: Will do in follow-up, since ID will be included in the JWT then

PR #1562, opened 2024-02-27
fix(ui): get email from me query

liangfung:


ee/tabby-ui/lib/tabby/auth.tsx
```diff
@@ -166,6 +168,7 @@ const AuthProvider: React.FunctionComponent<AuthProviderProps> = ({
     // After being mounted, listen for changes in the access token
     if (authToken?.accessToken && authToken?.refreshToken) {
       dispatch({ type: AuthActionType.SignIn, data: authToken })
+      reexecuteQueryMe()
```
liangfung, line 171: @wsxiaoys will refresh queryMe when token in storage changed

PR #1566, opened 2024-02-27
feat(db): Create Cache type which can be used to cache values for db

boxbeam:


crates/tabby-common/src/lib.rs
wsxiaoys, line 1: let's move this to tabby-webserver at the moment.

suggested path: `tabby-webserver/src/misc/cache.rs`

wsxiaoys (commented):
Otherwise LGTM

crates/tabby-common/src/lib.rs
boxbeam: I'll move it inside tabby-db instead, since that's where it's needed.

PR #1567, opened 2024-02-27
feat(db): Implement db-layer user count cache

boxbeam:
Note that the base is `cache-design` on PR https://github.com/TabbyML/tabby/pull/1566, this must be merged afterwards

### DIFF
--- a/ee/tabby-db/src/cache.rs
+++ b/ee/tabby-db/src/cache.rs
@@ -2,6 +2,7 @@ use std::future::Future;
 
 use tokio::sync::RwLock;
 
+#[derive(Default)]
 pub struct Cache<T> {
     value: RwLock<Option<T>>,
 }
--- a/ee/tabby-db/src/cache.rs
+++ b/ee/tabby-db/src/cache.rs
@@ -35,11 +35,6 @@ impl<T> Cache<T> {
         }
     }
 
-    pub async fn update(&self, f: impl FnOnce(&mut T)) {
-        let mut lock = self.value.write().await;
-        lock.as_mut().map(f);
-    }
-
     pub async fn set(&self, value: T) {
         *self.value.write().await = Some(value);
     }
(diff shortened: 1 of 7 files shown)


wsxiaoys:
Please resolve the conflicts

ee/tabby-db/src/users.rs
```diff
@@ -79,6 +79,11 @@ impl DbConn {
         let res = res.unique_error("User already exists")?;
         transaction.commit().await?;
 
+        self.cache.active_user_count.update(|x| *x += 1).await;
+        if is_admin {
+            self.cache.active_admin_count.update(|x| *x += 1).await;
+        }
+
```
wsxiaoys, line 86: instead of update count, I'll prefer to simply invalidate it so we have single place to compute the count (no need to handle error, rollback etc for cache).

ee/tabby-db/src/users.rs
```diff
@@ -179,10 +184,15 @@ impl DbConn {
         .await?
         .rows_affected();
         if changed != 1 {
-            Err(anyhow!("user active status was not changed"))
+            return Err(anyhow!("user active status was not changed"));
+        }
+        self.cache.active_admin_count.invalidate().await;
+        if active {
+            self.cache.active_user_count.update(|x| *x += 1).await;
```
wsxiaoys, line 191: ditto

ee/tabby-db/src/cache.rs
```diff
@@ -34,11 +35,6 @@ impl<T> Cache<T> {
         }
     }
 
-    pub async fn update(&self, f: impl FnOnce(&mut T)) {
-        let mut lock = self.value.write().await;
-        lock.as_mut().map(f);
-    }
-
     pub async fn set(&self, value: T) {
```
wsxiaoys, line 38: also remove set?

wsxiaoys (commented):
Otherwise LGTM

boxbeam:
Adding tests before merging

boxbeam:
Glad I did, because there was a bug!

PR #1571, opened 2024-02-28
Use db logger

boxbeam:


ee/tabby-webserver/src/handler.rs
```diff
@@ -8,61 +8,86 @@ use axum::{
 };
 use hyper::{Body, StatusCode};
 use juniper_axum::{graphiql, graphql, playground};
-use tabby_common::api::{code::CodeSearch, event::RawEventLogger, server_setting::ServerSetting};
+use tabby_common::api::{
+    code::CodeSearch,
+    event::{EventLogger, RawEventLogger},
+    server_setting::ServerSetting,
+};
+use tabby_db::DbConn;
```
boxbeam, line 16: @wsxiaoys This needs to import DbConn unless we want to move it to a service, but I don't think that would make sense. Could we add an exception?

wsxiaoys (commented):
Please also write more informative PR title

crates/tabby/src/serve.rs
```diff
@@ -126,15 +137,16 @@ pub async fn main(config: &Config, args: &ServeArgs) {
...
-            code,
-            args.chat_model.is_some(),
-            args.port,
-        )
-        .await;
+        let (api, ui) = ws
+            .attach_webserver(
+                api,
+                ui,
+                Arc::new(create_logger()),
```
wsxiaoys, line 144: why do we still pass in `create_logger` as logger is managed by `ws` now?

ee/tabby-webserver/src/handler.rs
```diff
@@ -8,63 +8,88 @@ use axum::{
...
+    pub async fn attach_webserver(
+        &self,
+        api: Router,
+        ui: Router,
+        logger: Arc<dyn RawEventLogger>,
+        code: Arc<dyn CodeSearch>,
+        is_chat_enabled: bool,
+        local_port: u16,
+    ) -> (Router, Router) {
+        let ctx = create_service_locator(logger, code, self.db.clone(), is_chat_enabled).await;
```
wsxiaoys, line 53:
```suggestion
        let ctx = create_service_locator(self.event_logger, code, self.db.clone(), is_chat_enabled).await;
```

crates/tabby/src/serve.rs
boxbeam: The webserver still takes `RawEventLogger`, which we can't really implement on the database unless we want to store raw strings or parse out the event data from them. This will still allow completion events to be logged to the database, since they are served through the regular API, and the correct logger will still be sent to the serve process which handles completions.

ee/tabby-webserver/src/handler.rs
```diff
@@ -25,6 +25,22 @@ use crate::{
     ui,
 };
 
+pub fn wrap_logger_raw(logger: Arc<dyn EventLogger>) -> Arc<dyn RawEventLogger> {
```
wsxiaoys, line 28: why this need to be public? also prefer just inlining it as this is only place it got used..

crates/tabby/src/services/event.rs
```diff
@@ -106,6 +106,7 @@ impl RawEventLogger for EventService {
     }
 }
 
+#[cfg(not(feature = "ee"))]
```
wsxiaoys, line 109:
```suggestion
```

ee/tabby-webserver/src/lib.rs
```diff
@@ -12,12 +12,13 @@ mod ui;
...
     pub use super::{
-        handler::attach_webserver,
+        handler::WebserverHandle,
         /* used by tabby workers (consumer of /hub api) */
         hub::api::{
             create_scheduler_client, create_worker_client, RegisterWorkerRequest, SchedulerClient,
             WorkerClient, WorkerKind,
         },
         /* used by examples/update-schema.rs */ schema::create_schema,
+        service::event_logger::new_event_logger,
```
wsxiaoys, line 22: why this need to be public?

wsxiaoys (changes requested):

ee/tabby-webserver/src/lib.rs
boxbeam: Used in handler.rs, could make it `pub(crate)`

wsxiaoys: handler should be able to reference `service::event_logger::new_event_logger` directly, as long as the function it self being marked as public (rather than re-export here)

ee/tabby-webserver/src/service/event_logger.rs
```diff
@@ -15,12 +18,16 @@ fn log_err<T, E: Display>(res: Result<T, E>) {
 }
 
 pub fn new_event_logger(db: DbConn) -> impl EventLogger {
```
wsxiaoys, line 20:
```suggestion
pub fn new_event_logger(db: DbConn) -> impl RawEventLogger {
```

Just realized you implement RawEventLogger directly for EventLogger

ee/tabby-webserver/src/handler.rs
```diff
@@ -8,63 +8,108 @@ use axum::{
...
+            warn!("Invalid event JSON: {content}");
+            return;
+        };
+        self.0.log(event);
+    }
+}
+
+pub struct WebserverHandle {
+    db: DbConn,
+    event_logger: Arc<dyn EventLogger>,
```
wsxiaoys, line 45:
```suggestion
    event_logger: Arc<dyn RawEventLogger>,
```

ee/tabby-webserver/src/handler.rs
```diff
@@ -8,63 +8,108 @@ use axum::{
...
+
+impl WebserverHandle {
+    pub async fn new() -> Self {
+        let db = DbConn::new().await.expect("Must be able to initialize db");
+        let event_logger = Arc::new(new_event_logger(db.clone()));
+        WebserverHandle { db, event_logger }
+    }
+
+    pub fn logger(&self) -> Arc<dyn EventLogger> {
+        self.event_logger.clone()
```
wsxiaoys, line 56:
```suggestion
        Arc::new(EventLoggerWrapper(self.event_logger.clone()))
```

wsxiaoys (changes requested):
Otherwise LGTM

ee/tabby-webserver/src/service/event_logger.rs
boxbeam: Yes, that was why I had added the supertrait relationship - `RawEventLogger` is implemented with a blanket impl for all `EventLogger`

wsxiaoys (approved):

wsxiaoys:
Please fix tests

crates/tabby/src/serve.rs
```diff
@@ -5,10 +5,15 @@ use clap::Args;
...
-    api::{code::CodeSearch, event::EventLogger},
+    api::{
+        code::CodeSearch,
+        event::{EventLogger, RawEventLogger},
+    },
     config::Config,
     usage,
 };
+#[cfg(feature = "ee")]
+use tabby_webserver::public::WebserverHandle;
```
wsxiaoys, line 16:
```suggestion
```
nit: specify qualified name in line 127 to reduce cfg blocks

crates/tabby/src/serve.rs
```diff
@@ -18,14 +23,15 @@ use utoipa::{
 };
 use utoipa_swagger_ui::SwaggerUi;
 
+#[cfg(not(feature = "ee"))]
+use crate::services::event::create_logger;
```
wsxiaoys, line 27:
```suggestion
```

ditto

ee/tabby-webserver/src/handler.rs
```diff
@@ -9,62 +9,82 @@ use axum::{
 use hyper::{Body, StatusCode};
 use juniper_axum::{graphiql, graphql, playground};
 use tabby_common::api::{code::CodeSearch, event::RawEventLogger, server_setting::ServerSetting};
+use tabby_db::DbConn;
```
wsxiaoys, line 12: could you fix the lint here? it should be straight forward to modify https://github.com/TabbyML/tabby/blob/60310d416177de629853d06e746d1a058ffd9a7b/rules/only-service-can-depend-tabby-db.yml#L8

(adding a ignore line for `handler.rs`)

wsxiaoys (changes requested):

PR #1583, opened 2024-02-29
feat(ui): integrate framer-motion

liangfung:
https://jam.dev/c/cfc64dce-7f84-4956-8fa9-42053959d695


wsxiaoys:
Nice demo. Let's keep it as reference for now - should be better place to utilize such motion.

PR #1586, opened 2024-02-29
fix(webserver): Better error message when the email does not match the code

boxbeam:
cc @liangfung


ee/tabby-webserver/src/service/auth.rs
```diff
@@ -502,7 +502,7 @@ async fn check_invitation(
     };
 
     if invitation.email != email {
-        return err;
+        return Err(anyhow!("Invitation code is not for this user").into());
```
wsxiaoys, line 505:
```suggestion
        return Err(anyhow!("Invitation code is not for this email address").into());
```

Issue #124, opened 2023-05-05
Vulkan Backend Support for improved device compatibility


FruityWelsh:
**Please describe the feature you want**
Support for the pytorch vulkan backend so that older nvidia gpus, as well as intell, amd, and some phone gpus can be supported.
https://pytorch.org/tutorials/prototype/vulkan_workflow.html

**Additional context**
Personally ran into difficulties testing this project, because my laptop is too old to support Nvidia, and my cloud accounts aren't authorized to deploy GPU compute. I imagine I am not the only one limited on working on this project by these kinds of lim facs.

---
Please reply with a 👍 if you want this feature.

wsxiaoys:
related：https://github.com/webonnx/wonnx/issues/169

wsxiaoys:
The recent merges of two Vulkan-based implementations in llama.cpp are as follows:

1. The first merge: [https://github.com/ggerganov/llama.cpp/commit/2307523d322af762ae06648b29ec5a9eb1c73032](https://github.com/ggerganov/llama.cpp/commit/2307523d322af762ae06648b29ec5a9eb1c73032)
2. The second merge: [https://github.com/ggerganov/llama.cpp/commit/fbf1ddec69f7001cc707de17fa74d7200813bbac](https://github.com/ggerganov/llama.cpp/commit/fbf1ddec69f7001cc707de17fa74d7200813bbac)

We are quite interested in integrating one of these implementations into Tabby. For example, by adding a `vulkan` device option, similar to the `metal` option, we can make Tabby compatible with certain GPUs, such as the Intel Arc.

wsxiaoys:
Note this is done in #1588 and a vulkan binary has been added in https://github.com/TabbyML/tabby/actions/runs/8179533225, consider give it a try!

cc @boxbeam

PR #1588, opened 2024-02-29
feat(tabby): Add experimental vulkan support

boxbeam:


wsxiaoys:
related:
https://github.com/TabbyML/tabby/issues/631
https://github.com/TabbyML/tabby/issues/124

boxbeam:
I'm wondering if your llama.cpp submodule is up to date. If you cloned Tabby a while ago, you need to run `git submodule update --init --recursive`. The required parameter was removed in the version we are now using.

PR #1591, opened 2024-03-01
fix(webserver): add seat count check in oauth self-registeration flow

wsxiaoys:


ee/tabby-webserver/src/oauth/mod.rs
```diff
@@ -145,6 +145,7 @@ fn match_auth_result(
             | OAuthError::UserNotInvited
             | OAuthError::UserDisabled
             | OAuthError::CredentialNotActive
+            | OAuthError::NoSufficientSeat
```
boxbeam, line 148:
```suggestion
            | OAuthError::InsufficientSeats
```

boxbeam (approved):

ee/tabby-webserver/src/schema/auth.rs
```diff
@@ -174,6 +174,9 @@ pub enum OAuthError {
     #[error("User is disabled, please contact admin for help")]
     UserDisabled,
 
+    #[error("License doesn't contain sufficient number of seats")]
```
boxbeam, line 177:
```suggestion
    #[error("Number of seats on license would be exceeded")]
```

PR #1600, opened 2024-03-01
refactor(db): Implement `query_paged_as!` macro to construct paginated queries that are still statically checked

boxbeam:
Things to note:
- Every column is selected as non-nullable, which might not be desirable
- & must be used to separate the last column instead of , due to a number of limitations

This might be better drafted as a procedural macro? Let me know if you'd be open to that, it would be more code but wouldn't come with these compromises


ee/tabby-db/src/invitations.rs
```diff
@@ -22,14 +22,7 @@ impl DbConn {
...
     ) -> Result<Vec<InvitationDAO>> {
-        let query = Self::make_pagination_query(
-            "invitations",
-            &["id", "email", "code", "created_at"],
-            limit,
-            skip_id,
-            backwards,
-        );
-
+        let query = pagination_query!(InvitationDAO FROM "invitations": ["id", "email", "code" & "created_at"] (limit, skip_id, backwards));
```
wsxiaoys, line 25:
```suggestion
        let query = pagination_query!(InvitationDAO FROM "invitations": ["id", "email", "code", "created_at"] (limit, skip_id, backwards));
```

typo?

boxbeam: See notes in PR description

ee/tabby-db/src/invitations.rs
```diff
@@ -22,14 +23,7 @@ impl DbConn {
...
     ) -> Result<Vec<InvitationDAO>> {
-        let query = Self::make_pagination_query(
-            "invitations",
-            &["id", "email", "code", "created_at"],
-            limit,
-            skip_id,
-            backwards,
-        );
-
+        let query = pagination_query!(InvitationDAO FROM "invitations" ["id", "email", "code", "created_at"!], limit, skip_id, backwards);
```
wsxiaoys, line 26:
```suggestion
        let query = query_pagination_as!(InvitationDAO, "invitations", ["id", "email", "code", "created_at"!], limit, skip_id, backwards);
```

Can the usage being updated to this style?

boxbeam: Can do

crates/tabby-macros/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+    let condition = match input.condition {
+        Some(cond) => quote! {Some(#cond.into())},
+        None => quote! {None},
+    };
+
+    let limit = input.limit;
+    let skip_id = input.skip_id;
+    let backwards = input.backwards;
+    quote! {
+            sqlx::query_as(&crate::make_pagination_query_with_condition({
```
wsxiaoys, line 116: this makes `tabby-macros` library only meaningful for `tabby-db`, correct? Maybe move it to `ee/tabby-db-macros`

ee/tabby-db/src/job_runs.rs
```diff
@@ -71,7 +72,7 @@ impl DbConn {
         } else {
             None
         };
-        let query = Self::make_pagination_query_with_condition(
+        let query = make_pagination_query_with_condition(
```
wsxiaoys, line 75: any reason this cannot be covered by macro?

ee/tabby-db/src/lib.rs
```diff
@@ -49,6 +49,54 @@ pub struct DbConn {
     cache: Arc<DbCache>,
 }
 
+fn make_pagination_query_with_condition(
```
wsxiaoys, line 52: consider move code that will be used by macros into a separate file, e.g `src/macro_deps.rs`

wsxiaoys (changes requested):

ee/tabby-db/src/job_runs.rs
boxbeam: I didn't convert every use case which could be covered using this macro, just the one for demonstration. I was going to do the rest in follow-up PRs.

ee/tabby-db/src/lib.rs
boxbeam: It's not only used by macros for the time being, so I think we should leave it here until the refactoring is complete.

crates/tabby-macros/src/lib.rs
boxbeam: I will do that, but I want to note that this has no actual dependency on `tabby-db` and has no feature flags associated with it, since it never actually uses that code, only generates syntax which refers to it. I was figuring that it would make sense to group all our macros in one place, if we want to add more.

ee/tabby-db/src/job_runs.rs
boxbeam: Reasoning: You must be able to use `query_as!` with a type in order to use this new macro with it, so it's not as simple as just swapping it out, we need to convert datetime and IDs, which have been split up into separate PRs so far.

PR #1608, opened 2024-03-04
fix(ui): modify destructive color in dark mode

liangfung:


liangfung:
before
image

after
image

PR #1620, opened 2024-03-04
feat: Add dockerfile to build vulkan image

boxbeam:
Fixes TAB-485, verified build works and image successfully uses vulkan when running locally


boxbeam:
cc @icycodes

PR #1633, opened 2024-03-06
fix: handle usize overflow when matched stop_length >= text length

wsxiaoys:


crates/tabby-inference/src/imp.rs
```diff
@@ -53,21 +53,20 @@ impl<T: TextGenerationStream> TextGeneration for TextGenerationImpl<T> {
...
             for await new_text in self.imp.generate(&prompt, options).await {
                 let (should_stop, stop_length) = stop_condition.should_stop(&new_text);
                 text += &new_text;
                 yield (true, new_text);
                 if should_stop {
-                    stopped = true;
-                    yield (false, text[..(text.len() - stop_length)].to_owned());
-                    break;
+                    // stop condition matched against prompt + generated text. There's a chance that stop_length >= text.len();
+                    let new_text_length = text.len().checked_sub(stop_length).unwrap_or_default();
```
boxbeam, line 62:
```suggestion
                    let new_text_length = text.len().min(stop_length);
```

boxbeam (changes requested):
Using checked_sub seems weird, it will cut off the entire rest of the string instead of grabbing the right number of characters. Use min instead?

crates/tabby-inference/src/imp.rs
wsxiaoys: the formula is: new_text_length = text.len() - stop_length, not new_text_length = min(text.len(), stop_length)

PR #1634, opened 2024-03-06
fix(tabby): Make event log user-private

### LINKED ISSUE #1631: `events` logs are world-readable, contain all text sent to the model, and can't be turned off
**Describe the bug**
Tabby logs all input to the model to `events` files (see #959).

These files are created with ordinary default file permissions, which by default on many systems (including mine) makes them world-readable. Other unprivileged users on the system can read them.

If I have Tabby enabled in my editor and open and type in a file containing secrets, those secrets will be sent to Tabby to generate completions and then logged to the world-readable `events` logs, where other users can access them.

I last tested this with Tabby v0.7.0.

N/A

**Additional context**
I've been using Tabby on our lab's shared GPU machines, which are classic multi-user, multi-tenant Unix machines where multiple users can do things like access local ports, the filesystem, and the command lines of running processes.

But if I can't draw a security boundary around Tabby that lets it safely handle secret text on a multi-user system, and I can't configure the Tabby vim plugin to not operate on certain filetypes or file names to keep secrets out of Tabby, then I can't use Tabby in my environment.

boxbeam:
Fixes #1631
cc @adamnovak

### DIFF
--- a/crates/tabby/src/main.rs
+++ b/crates/tabby/src/main.rs
@@ -7,6 +7,9 @@ mod serve;
 #[cfg(feature = "ee")]
 mod worker;
 
+#[cfg(target_family = "unix")]
+use std::os::unix::fs::PermissionsExt;
+
 use clap::{Parser, Subcommand};
 use opentelemetry::{
     global,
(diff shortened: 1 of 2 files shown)


crates/tabby/src/main.rs
```diff
@@ -134,6 +137,14 @@ async fn main() {
     init_logging(cli.otlp_endpoint);
 
     let config = Config::load().unwrap_or_default();
+    let root = tabby_common::path::tabby_root();
+    std::fs::create_dir_all(&root).expect("Must be able to create tabby root");
+    #[cfg(target_family = "unix")]
```
wsxiaoys, line 142: is there an windows equivalent to 0600 for directory permission?

crates/tabby/src/services/event.rs
```diff
@@ -75,6 +75,7 @@ impl EventWriter {
                 .create(true)
                 .append(true)
                 .write(true)
+                .mode(0o600)
```
wsxiaoys, line 78: this can. be removed?

crates/tabby/src/main.rs
boxbeam: I believe windows doesn't let you read other users' files by default. The only permission that can be modified without the unix extension is whether the file is readonly.

PR #1644, opened 2024-03-07
fix(common): Sanitize repository names generated from filenames

### LINKED ISSUE #1641: Invalid characters in repository with private repositories github_pat_...
**Describe the bug**
Scheduler has problems with private repositories with personal access token as url: https://github_pat_xxxxx...@github.com/xxx.git

0.9.0 on mac sonoma

**Additional context**
Prior to version 0.9.0 fetching private repositories configured in config.toml
like https://github_pat_xxxxx...@github.com/xxx.git was never a problem.

Since 0.9.0 i get:

Warning: Could not parse the Tabby configuration at /Users/user/.tabby/config.toml
Syncing 0 repositories...
Reason: Invalid characters in repository name: https_github_pat_xxxx@github.com/xxx.git
Falling back to default config, please resolve the errors and restart Tabby

Greetings & by the way: thanks for your great work!

boxbeam:
Fixes #1641

### DIFF
--- a/crates/tabby-common/src/config.rs
+++ b/crates/tabby-common/src/config.rs
@@ -67,7 +67,7 @@ impl Config {
 }
 
 lazy_static! {
-    pub static ref REPOSITORY_NAME_REGEX: Regex = Regex::new("^[a-zA-Z][\\w.@-]+$").unwrap();
+    pub static ref REPOSITORY_NAME_REGEX: Regex = Regex::new("^[a-zA-Z][\\w.-]+$").unwrap();
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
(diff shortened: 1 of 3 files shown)


wsxiaoys (commented):
instead of allowing `@` in filename, can we adjust `filenamify` logic to ensure it only generates name that's `validate_name() == true`?

boxbeam:
filenamify is not a crate we made. I could run additional conversions on it after we use filenamify, but I think it would make more sense to just update the regex to allow valid filename characters to be used. it might be good to go through and try to exhaustively list out which are allowed.

wsxiaoys:
> filenamify is not a crate we made. I could run additional conversions on it after we use filenamify, but I think it would make more sense to just update the regex to allow valid filename characters to be used. it might be good to go through and try to exhaustively list out which are allowed.

or we can implement filenamify by ourselves? since we only allow `[a-zA-Z_]`, it should be straightforward to replace everything else with `_`?

boxbeam:
yes, I can implement that

boxbeam:
however, it does call into question why we're validating repository names in the first place if we already have something which can guarantee that they will have valid filenames?

wsxiaoys:
> however, it does call into question why we're validating repository names in the first place if we already have something which can guarantee that they will have valid filenames?

name field can be configured by user, it's only populated by `filenamify` if `name` field is missing. We still need to verify against user filled name.

crates/tabby-common/src/config.rs
```diff
@@ -111,14 +111,21 @@ impl RepositoryConfig {
...
-        } else {
-            filenamify(&self.git_url)
-        }
+        self.name
+            .clone()
+            .unwrap_or_else(|| clean_name(filenamify(&self.git_url)))
     }
 }
 
+fn clean_name(s: String) -> String {
```
wsxiaoys, line 120:
```suggestion
fn sanitize_name(s: String) -> String {
```

add unit test for this? at least include `@` to test.

wsxiaoys (commented):
Please also update PR title

wsxiaoys (approved):

PR #1651, opened 2024-03-10
fix(webserver): incorrect node limit check on worker register

SpeedCrash100:
# Brief
There is a problem with missing NOT when checking current node limit.

# Description
The current version of [check_node_limit](https://github.com/TabbyML/tabby/blob/a80935f4637187ddea0b8c5dfb4f3d0161f796c2/ee/tabby-webserver/src/schema/license.rs#L45) returns true if the node limit has not been reached and, therefore, a worker should be registered but [here](https://github.com/TabbyML/tabby/blob/a80935f4637187ddea0b8c5dfb4f3d0161f796c2/ee/tabby-webserver/src/service/mod.rs#L167) function fails resulting in Requires Enterprise License error. The PR adds missing NOT operation.

### DIFF
--- a/ee/tabby-webserver/src/service/mod.rs
+++ b/ee/tabby-webserver/src/service/mod.rs
@@ -164,7 +164,7 @@ impl WorkerService for ServerContext {
             .await
             .map_err(|_| RegisterWorkerError::RequiresEnterpriseLicense)?;
 
-        if license.check_node_limit(count_workers + 1) {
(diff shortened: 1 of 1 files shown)


wsxiaoys:
Thank you for the PR. I created a [new one](https://github.com/TabbyML/tabby/pull/1652) because I was unable to edit the changelog in this branch.

PR #1655, opened 2024-03-11
refactor(common): Refactor EventLogger and RawEventLogger into single trait

boxbeam:
Closes TAB-554

### DIFF
--- a/crates/tabby-common/src/api/event.rs
+++ b/crates/tabby-common/src/api/event.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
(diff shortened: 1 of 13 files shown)


ee/tabby-db-macros/Cargo.toml
wsxiaoys, line 1: This doesn't seems to be related to this PR, please rebase

wsxiaoys (changes requested):

crates/tabby-common/src/api/event.rs
```diff
@@ -95,35 +93,23 @@ pub struct Segments {
...
             ts: timestamp(),
             event: e,
         })
         .unwrap();
 
-        self.log(content);
+        self.log_raw(content);
     }
+
+    fn log_raw(&self, content: String);
```
wsxiaoys, line 106:
```suggestion
    fn log(&self, content: String);
```

Keep the name

crates/tabby-common/src/api/event.rs
```diff
@@ -95,35 +93,23 @@ pub struct Segments {
 }
 
 pub trait EventLogger: Send + Sync {
```
wsxiaoys, line 95: keep the name `RawEventLogger`

wsxiaoys (changes requested):

crates/tabby-common/src/api/event.rs
boxbeam: Can't do that, there is already a method `log` which takes an `Event` and rust does not allow overloading

crates/tabby-common/src/api/event.rs
boxbeam: Curious to know the rationale? If there is no EventLogger, and it can handle both structs and "raw" events (as strings), what makes the trait specifically a "Raw" event logger?

PR #1657, opened 2024-03-11
refactor(agent): use stats-logscale@1.0.9 which has native TS support

dallaylaen:


icycodes:
@dallaylaen 
Thank you for the update!

PR #1670, opened 2024-03-13
refactor(db): Convert job run DAO to be usable with query_paged_as

boxbeam:
Relates to TAB-521

### DIFF
--- a/ee/tabby-db/src/job_runs.rs
+++ b/ee/tabby-db/src/job_runs.rs
@@ -3,7 +3,7 @@ use sqlx::{query, FromRow};
 use tabby_db_macros::query_paged_as;
 
 use super::DbConn;
-use crate::{DateTimeUtc, DbOption};
+use crate::{DateTimeUtc, DbNullable};
 
 #[derive(Default, Clone, FromRow)]
 pub struct JobRunDAO {
@@ -17,7 +17,7 @@ pub struct JobRunDAO {
     pub updated_at: DateTimeUtc,
 
     #[sqlx(rename = "end_ts")]
-    pub finished_at: DbOption<DateTimeUtc>,
+    pub finished_at: DbNullable<DateTimeUtc>,
 }
 
 /// db read/write operations for `job_runs` table
--- a/ee/tabby-db/src/job_runs.rs
+++ b/ee/tabby-db/src/job_runs.rs
@@ -3,7 +3,7 @@ use sqlx::{query, FromRow};
 use tabby_db_macros::query_paged_as;
 
 use super::DbConn;
-use crate::{DateTimeUtc, DbNullable};
+use crate::{DateTimeUtc, DbOption};
 
 #[derive(Default, Clone, FromRow)]
 pub struct JobRunDAO {
@@ -17,7 +17,7 @@ pub struct JobRunDAO {
     pub updated_at: DateTimeUtc,
 
     #[sqlx(rename = "end_ts")]
-    pub finished_at: DbNullable<DateTimeUtc>,
+    pub finished_at: DbOption<DateTimeUtc>,
 }
 
 /// db read/write operations for `job_runs` table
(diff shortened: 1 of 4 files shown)


ee/tabby-db/src/job_runs.rs
```diff
@@ -1,23 +1,23 @@
...
     pub stdout: String,
     pub stderr: String,
-    pub created_at: DateTime<Utc>,
-    pub updated_at: DateTime<Utc>,
+    pub created_at: DateTimeUtc,
+    pub updated_at: DateTimeUtc,
 
     #[sqlx(rename = "end_ts")]
-    pub finished_at: Option<DateTime<Utc>>,
+    pub finished_at: DbNullable<DateTimeUtc>,
```
wsxiaoys, line 20: We need this DbNullable only because we created DateTimeUtc, correct?

boxbeam: Yes, it doesn't automatically know how to convert `Option<NaiveDateTime>` to `Option<DateTimeUTC>`, and I can't implement `From<Option<NaiveDateTime>>` for `Option<DateTimeUtc>` (and can't do it with `Into` either) due to the orphan rule. I made `DbNullable` which should be able to handle optional conversions for any other custom database types we may need.

wsxiaoys: hmm, interesting. I was wondering how sqlx implements similar concept? for example, there has to be very similar implementation for `DateTime<Utc>` and `Option<DateTime<Utc>>`, why does it gain `Option<>` trait automatically?

boxbeam: It uses Option for nullable types, and then uses From/Into to convert to the type of the struct field.

boxbeam: Worth noting: We still can't use `DateTime<Utc>` as a db type. sqlx knows how to convert its own types (since it just wraps an already-supported type in `Option`), but it only converts using `From`. This means that it sees the column is nullable, wraps it to create `Option<NaiveDateTime>`, and then expects to be able to convert to our target type from that. We can't implement `From<Option<NaiveDateTime>>` for `Option<DateTimeUtc>`, so we have to create a custom type which we *can* implement that for. It's a weird quirk, to be sure.

wsxiaoys: Let's do a discuss offline - if it's not possible, let's stick with `Option<DateTime<Utc>>`

boxbeam: We can't use `Option<DateTime<Utc>>` for the same reason we can't use `DateTime<Utc>`

ee/tabby-db/src/lib.rs
```diff
@@ -187,6 +188,61 @@ impl DbConn {
     }
 }
 
+#[derive(Default)]
+pub struct DbNullable<T>(Option<T>);
```
boxbeam, line 192: Create new trait to restrict `T` to only `DateTimeUTC` for now

ee/tabby-db/src/lib.rs
```diff
@@ -187,6 +188,69 @@ impl DbConn {
     }
 }
 
+pub trait DbCustom: for<'a> Decode<'a, Sqlite> + for<'a> Encode<'a, Sqlite> + Type<Sqlite> {}
+impl DbCustom for DateTimeUtc {}
```
wsxiaoys, line 192:
```suggestion
pub trait DbNullable: for<'a> Decode<'a, Sqlite> + for<'a> Encode<'a, Sqlite> + Type<Sqlite> {}
impl DbNullable for DateTimeUtc {}
```

ee/tabby-db/src/lib.rs
```diff
@@ -187,6 +188,69 @@ impl DbConn {
...
+
+#[derive(Default)]
+pub struct DbNullable<T>(Option<T>)
```
wsxiaoys, line 195:
```suggestion
pub struct DbOption<T>(Option<T>)
```

wsxiaoys (approved):
Otherwise LGTM

boxbeam:
Check if DbNullable trait can be private

PR #1675, opened 2024-03-14
fix(webserver): Check admin status via auth service instead of JWT

boxbeam:
Fixes TAB-545


wsxiaoys:
adding @liangfung to reviewer.

please verify removing `is_admin` from JWT works from frontend perspective

liangfung:
> please verify removing `is_admin` from JWT works from frontend perspective
Yes, it works. I will create a new PR later to modify the frontend implementation to use `queryMe`

liangfung (approved):

wsxiaoys (approved):

PR #1676, opened 2024-03-14
refactor(db, webserver): Represent passwords as nullable in db and service layer

boxbeam:
Closes TAB-530


wsxiaoys:
cc @darknight to help take a look

ee/tabby-webserver/src/service/dao.rs
```diff
@@ -52,7 +52,7 @@ impl From<UserDAO> for auth::User {
             auth_token: val.auth_token,
             created_at: val.created_at,
             active: val.active,
-            is_password_set: !val.password_encrypted.is_empty(),
+            is_password_set: val.password_encrypted.is_some(),
```
darknight, line 55: I'm not sure but `password_encrypted` can be `None` also can be `empty string`, right?
If empty string is out of question, then the change looks good to me.

boxbeam: This change converts all instances of empty-string passwords to None, and will use None instead of empty string to represent empty passwords internally. I did a search in the codebase and I don't see any instances of `Some("")` being used for the password, nor do I see any places where a user could specify `None` (as that logic is handled completely internally based on OAuth). I believe this is sound in ensuring an emptystring password will never be used.

PR #1679, opened 2024-03-15
fix(ui): initialize the form with the latest values

liangfung:
When using urql to query, set the requestPolicy to 'network-only' to ensure the form values are up-to-date and to avoid manually maintaining 'initialized'

### DIFF
--- a/ee/tabby-ui/app/(dashboard)/settings/(integrations)/mail/components/mail.tsx
+++ b/ee/tabby-ui/app/(dashboard)/settings/(integrations)/mail/components/mail.tsx
@@ -5,6 +5,7 @@ import { OperationResult } from 'urql'
 
 import { graphql } from '@/lib/gql/generates'
 import { EmailSettingQuery } from '@/lib/gql/generates/graphql'
+import { useDebounceValue } from '@/lib/hooks/use-debounce'
 import { client } from '@/lib/tabby/gql'
 import { ListSkeleton } from '@/components/skeleton'
 
@@ -32,6 +33,9 @@ export const Mail = () => {
   const [queryResult, setQueryResult] =
     React.useState<OperationResult<EmailSettingQuery, any>>()
   const [initialized, setInitialized] = React.useState(false)
+  const [debouncedInitialzed] = useDebounceValue(initialized, 200, {
+    leading: true
+  })
   const mailFormRef = React.useRef<MailFormRef>(null)
 
   const queryEmailSettings = () => {
@@ -80,7 +84,7 @@ export const Mail = () => {
   return (
     <>
       <MailDeliveryHeader />
-      {initialized ? (
+      {debouncedInitialzed ? (
         <div>
           <div className="mb-8 border-b pb-4">
             <MailForm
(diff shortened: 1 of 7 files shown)


ee/tabby-ui/app/(dashboard)/settings/general/components/network-form.tsx
```diff
@@ -112,15 +115,25 @@ const NetworkForm: React.FC<NetworkFormProps> = ({
...
 export const GeneralNetworkForm = () => {
-  const [{ data }, reexecuteQuery] = useNetworkSetting()
+  const [queryState, reexecuteQuery] = useNetworkSetting({
+    requestPolicy: 'network-only'
+  })
+
+  const { data, fetching, stale } = queryState
+
+  const [shouldInitializeForm] = useDebounceValue(
+    data && !stale && !fetching,
```
liangfung, line 125: Because this query is also used elsewhere, it still hits the cache, so a combination of three variables is used to determine the outcome.

ee/tabby-ui/app/(dashboard)/settings/(integrations)/mail/components/mail.tsx
```diff
@@ -32,6 +34,9 @@ export const Mail = () => {
   const [queryResult, setQueryResult] =
     React.useState<OperationResult<EmailSettingQuery, any>>()
   const [initialized, setInitialized] = React.useState(false)
+  const [debouncedInitialzed] = useDebounceValue(initialized, SKELETON_DELAY, {
```
liangfung, line 37: Use `useDebounceValue` to avoid flicker

ee/tabby-ui/app/(dashboard)/settings/general/components/network-form.tsx
```diff
@@ -112,15 +107,18 @@ const NetworkForm: React.FC<NetworkFormProps> = ({
 }
 
 export const GeneralNetworkForm = () => {
-  const [{ data }, reexecuteQuery] = useNetworkSetting()
+  const [{ data, stale }, reexecuteQuery] = useNetworkSetting({
+    requestPolicy: 'network-only'
+  })
+
```
wsxiaoys, line 113: since the policy is being changed to `network-only`, does it still make sense to use `stale` field? i would assume `data` will be `null` before network request completed?

liangfung: This query is also used elsewhere, and the presence of the default policy 'cache-and-network' used in other places conflicts with 'network-only' used here. Specifying the policy individually is prone to errors, so the plan is to not specify the 'network-only' policy separately, continue to use 'cache-and-network', and combine data and stale to initialize the form.

liangfung:
Because the 'cache-and-network' request policy is used, combine 'stale' to determine if the data is up-to-date.

wwayne (approved):

ee/tabby-ui/app/(dashboard)/settings/general/components/network-form.tsx
wwayne: LGTM

PR #1674, opened 2024-03-14
feat(api): add file info in completion request.

icycodes:
Complete TAB-529
Followed by #1682

PR #1682, opened 2024-03-15
feat(agent&vscode): add filepath & git_url in completion request.

icycodes:
Complete TAB-529
#1674 

Changes:
- In agent: supported `filepath` & `git_url` in segments for completion requests.
- In VSCode extension, use workspace and git api to get relevant info.


clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -10,6 +10,9 @@ export type AgentConfig = {
       experimentalStripAutoClosingCharacters: boolean;
       maxPrefixLines: number;
       maxSuffixLines: number;
+      filepath: {
+        experimentalEnabled: boolean;
```
wsxiaoys, line 14: i feel this is not necessary since the content should be always filled as long as agent can detect it.

PR #1685, opened 2024-03-16
feat(tabby): Added --chat-device arg for serve subcommand

### LINKED ISSUE #1659: Add a --chat-device flag for serve command
Add the _--chat-device_ flag to _serve_ subcommand to allow use different device for chat.
* If the flag is not given, then chat model should use _--device_ as it is now.
* If the flag is specified but chat model is not selected, give a warning, I guess

### The reasoning
I use Tabby for both chat and completions. Completions is the main usage and I give it the best model my GPU can handle leaving chat model to CPU. I see no reasons to lose quality of main usage just to handle chat and completions model simultaneously.  

Previously, I solved this by running workers for each model. I was able to run 1 worker for chat and completions and if I really need to dynamically redeploy workers to use GPU for chat when I really need to.  However, this is now an enterprise feature and I cannot use it to do that anymore. 

### Alternative solution
Rework subscriptions plan to count workers per type (chat, completions) and allow  1 worker per type in community license. When the _serve_ subcommand is used with models, web-server instance should be counted as workers to avoid scaling without license.

---

SpeedCrash100:
# Brief
A implementation of issue #1659 

# Details
The very simple implementation of the --chat-device flag to override device to use for chat model. This implementation, however, rejects arguments if --chat-device specified without the --chat-model flag. The issue suggests giving a warning instead, but I found out a good `clap` macro to achieve the implemented behaviour. I decided to not implement a logging warning message and give the situation handling to the `clap` crate to keep the code clean.

# Testing
Tested with rocm + cpu devices. I cannot test other devices myself.

---
closes  #1659

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v0.10.0 [UNRELEASED]
 
 ## Features
+* Added the --chat-device flag to override a device used to run chat model.
 
 ## Fixes and Improvements
 
--- a/crates/tabby/src/serve.rs
+++ b/crates/tabby/src/serve.rs
@@ -97,6 +97,10 @@ pub struct ServeArgs {
     #[clap(long, default_value_t=Device::Cpu)]
     device: Device,
 
+    /// Device to run chat model [default equals --device arg]
+    #[clap(long, requires("chat_model"))]
+    chat_device: Option<Device>,
+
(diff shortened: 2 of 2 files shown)


wsxiaoys (approved):

wsxiaoys:
Thanks for the PR, LGTM

PR #1687, opened 2024-03-18
feat(ui): add analytic dashboard

liangfung:
Close TAB-557


wwayne:
WIP, collecting feedbacks
屏幕截图 2024-03-22 120632
屏幕截图 2024-03-22 120643

PR #1688, opened 2024-03-18
feat(ui): check admin status via service instead of JWT

liangfung:
Relates #1675 
Use 'queryMe'

### DIFF
--- a/ee/tabby-ui/app/(dashboard)/components/sidebar.tsx
+++ b/ee/tabby-ui/app/(dashboard)/components/sidebar.tsx
@@ -32,7 +32,7 @@ export interface SidebarProps {
 
 export default function Sidebar({ children, className }: SidebarProps) {
   const [{ data }] = useMe()
-  const isAdmin = data?.me?.isAdmin
+  const isAdmin = data?.me.isAdmin
   return (
     <div
       className={cn('grid overflow-hidden md:grid-cols-[280px_1fr]', className)}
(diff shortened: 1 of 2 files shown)


ee/tabby-ui/lib/tabby/auth.tsx
```diff
@@ -271,7 +254,6 @@ function useSignOut(): () => Promise<void> {
 }
 
 interface User {
```
wsxiaoys, line 256: nit: rename this to something like JWTPayload

ee/tabby-ui/app/(dashboard)/components/sidebar.tsx
```diff
@@ -31,8 +31,8 @@ export interface SidebarProps {
 }
 
 export default function Sidebar({ children, className }: SidebarProps) {
-  const { data: session } = useSession()
-  const isAdmin = session?.isAdmin || false
+  const [{ data }] = useMe()
+  const isAdmin = data?.me?.isAdmin
```
wsxiaoys, line 35:
```suggestion
  const isAdmin = data?.me.isAdmin
```

`me` is always not null if data is not null?

ee/tabby-ui/lib/tabby/auth.tsx
liangfung: Used `IJwtPayload` to differentiate from the `jwtPayload` exported by `jwt-decode`.

ee/tabby-ui/lib/tabby/auth.tsx
```diff
@@ -270,8 +253,7 @@ function useSignOut(): () => Promise<void> {
   }
 }
 
-interface User {
-  isAdmin: boolean
+interface IJwtPayload {
```
wsxiaoys, line 256:
```suggestion
interface JwtPayload {
```

Prefer stating type directly, without `I` prefix

Issue #1697, opened 2024-03-19
Panic when indexing a repository from config


sinnerJoe:
**Describe the bug**
I added the repo of my company's project to `~/.tabby/config.toml`, then I ran `tabby scheduler --now`.

I got a panic.
```
Building dataset...
100%|████████████████████████████████████████| 17776/17776 [02:47<00:00, 105.88it/s]Indexing repositories...
 50%|████████████████████▍                   |  8974/17776 [00:01<00:01, 5268.63it/s]thread 'main' panicked at crates/tabby-scheduler/src/index.rs:86:14:
Range is known valid
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: tabby_scheduler::index::index_repositories
   4: tabby_scheduler::job_index
   5: tabby::main::{{closure}}
   6: tokio::runtime::park::CachedParkThread::block_on
   7: tabby::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

**Information about your version**
tabby 0.9.0

**Information about your GPU**
Apple M2 Max

**Additional context**
I installed tabby using `brew`.
I tried to run the command both when the server is turned on and off.

wsxiaoys:
cc @boxbeam

boxbeam:
@sinnerJoe Thanks for reporting this! I've made a PR to improve this error message so we can get more information about the issue. You can either wait for this PR to be merged, or clone the repository and checkout the branch `tag-range-error` locally. A nightly build will also be released soon which includes this patch.

If you want to build locally, you can follow this guide after checking out the branch `tag-range-error`: https://github.com/TabbyML/tabby/blob/main/CONTRIBUTING.md

Once you have it running from a local build, please reproduce the error, and it should now display which file it originates from and the range that it was trying to read from. Please share the contents of that file and the error message with us, and we should be able to determine the cause.

sinnerJoe:
Hey @boxbeam ,
I ran the scheduler again with the nightly changes. 
Here is the output
```
100%|████████████████████████████████████████| 17910/17910 [03:20<00:00, 89.38it/s]Indexing repositories...
 51%|████████████████████▌                   |  9108/17910 [00:30<00:29, 296.15it/s]thread 'main' panicked at crates/tabby-scheduler/src/index.rs:79:28:
[...]

PR #1703, opened 2024-03-20
fix(scheduler): Improve error message on tag range out-of-bounds

boxbeam:
Relates to #1697

### DIFF
--- a/crates/tabby-scheduler/src/index.rs
+++ b/crates/tabby-scheduler/src/index.rs
@@ -73,17 +73,17 @@ struct IndexedDocument {
     kind: String,
 }
 
-fn get_tag(filename: &str, content: &str, range: Range<usize>) -> String {
+fn read_range_or_panic(filename: &str, content: &str, range: Range<usize>) -> String {
     content
         .get(range.clone())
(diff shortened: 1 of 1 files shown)


crates/tabby-scheduler/src/index.rs
```diff
@@ -73,18 +73,17 @@ struct IndexedDocument {
     kind: String,
 }
 
+fn get_tag(filename: &str, content: &str, range: Range<usize>) -> String {
```
wsxiaoys, line 76:
```suggestion
fn read_range_or_panic(filename: &str, content: &str, range: Range<usize>) -> String {
```

crates/tabby-scheduler/src/index.rs
```diff
@@ -73,18 +73,17 @@ struct IndexedDocument {
...
+    content
+        .get(range.clone())
+        .unwrap_or_else(|| panic!("Failed to retrieve bytes {range:?} from {filename}"))
```
wsxiaoys, line 79:
```suggestion
        .unwrap_or_else(|| panic!("Failed to read content '{range:?}' from '{filename}'"))
```

wsxiaoys (approved):
Otherwise LGTM

PR #1710, opened 2024-03-22
feat(ui): add profile analytics view

wwayne:
image

image

Fix TAB-565


ee/tabby-ui/lib/types/stats.ts
wsxiaoys, line 1: can we just reuse the type defined in gql? `DailyStatsQuery.dailyStats`

wwayne: fixed, removed the type and start using the type from `@/lib/gql/generates`

ee/tabby-ui/app/(home)/components/stats.tsx
liangfung, line 1: Perhaps move it directly into stats.tsx, until there is a need for reuse

ee/tabby-ui/app/(home)/use-language-stats.ts
liangfung, line 1: Perhaps move it directly into stats.tsx, until there is a need for reuse

wwayne: updated and merged the hook into the component

PR #1720, opened 2024-03-25
feat(webserver): Implement new job runs APIs

boxbeam:
Fixes TAB-563
cc @liangfung

### DIFF
--- a/ee/tabby-db/migrations/0020_job-run-index.up.sql
+++ b/ee/tabby-db/migrations/0020_job-run-index.up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX idx_job ON job_runs(job);
--- a/ee/tabby-db/migrations/0020_job-run-index.up.sql
+++ b/ee/tabby-db/migrations/0020_job-run-index.up.sql
@@ -1,1 +1,1 @@
-CREATE INDEX idx_job ON job_runs(job);
+CREATE INDEX idx_job_created_at ON job_runs(job, created_at);
--- a/ee/tabby-db/src/job_runs.rs
+++ b/ee/tabby-db/src/job_runs.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use chrono::{Duration, Utc};
 use sqlx::{query, FromRow};
 use tabby_db_macros::query_paged_as;
 
@@ -18,6 +19,13 @@ pub struct JobRunDAO {
 
     #[sqlx(rename = "end_ts")]
     pub finished_at: DbOption<DateTimeUtc>,
+}
+
+#[derive(FromRow)]
+pub struct JobStatsDAO {
+    pub success: i32,
+    pub failed: i32,
+    pub pending: i32,
 }
 
 /// db read/write operations for `job_runs` table
@@ -103,6 +111,31 @@ impl DbConn {
         .await?;
 
         Ok(job_runs)
+    }
+
+    pub async fn compute_job_stats(&self, jobs: Option<Vec<String>>) -> Result<JobStatsDAO> {
+        let condition = match jobs {
+            Some(jobs) => {
+                let jobs: Vec<_> = jobs.into_iter().map(|s| format!("{s:?}")).collect();
+                let jobs = jobs.join(", ");
+                format!("AND job IN ({jobs})")
+            }
+            None => "".into(),
+        };
+
+        let cutoff = Utc::now() - Duration::days(7);
+
+        let stats = sqlx::query_as(&format!(
+            r#"SELECT
+                SUM(exit_code == 0) AS success,
+                SUM(exit_code != 0 AND exit_code IS NOT NULL) AS failed,
+                SUM(exit_code IS NULL) AS pending FROM job_runs
+                WHERE start_ts > ? {condition};"#
+        ))
+        .bind(cutoff)
+        .fetch_one(&self.pool)
+        .await?;
+        Ok(stats)
     }
 
     pub async fn cleanup_stale_job_runs(&self) -> Result<()> {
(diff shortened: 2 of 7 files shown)


ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -319,6 +327,10 @@ impl Query {
     async fn license(ctx: &Context) -> Result<LicenseInfo> {
         ctx.locator.license().read_license().await
     }
+
+    async fn jobs() -> Result<Vec<String>> {
+        Ok(vec!["sync".into(), "index".into()])
```
wsxiaoys, line 332:
```suggestion
        Ok(vec!["scheduler".into()])
```

ee/tabby-webserver/src/service/job.rs
```diff
@@ -45,12 +46,37 @@ impl JobService for DbConn {
...
+            .list_job_runs(None, jobs, None, None, None, None)
+            .await?;
+        let mut stats = JobStats {
+            success: 0,
+            failed: 0,
+            pending: 0,
+        };
+
+        for job in jobs {
+            if job.finished_at.is_none() {
```
wsxiaoys, line 67:
```suggestion
            if job.exit_code.is_none() {
```

might consider use a match clause

wsxiaoys (approved):
Otherwise LGTM

wsxiaoys:
cc @wwayne for api review

ee/tabby-webserver/src/service/job.rs
```diff
@@ -45,12 +46,37 @@ impl JobService for DbConn {
...
+            .list_job_runs_with_filter(rowids, jobs, limit, skip_id, backwards)
             .await?
             .into_iter()
             .map(Into::into)
             .collect())
     }
+
+    async fn compute_job_run_stats(&self, jobs: Option<Vec<String>>) -> Result<JobStats> {
+        let jobs = self
+            .list_job_runs(None, jobs, None, None, None, None)
```
wsxiaoys, line 58: only list job runs for last 7 days

ee/tabby-webserver/src/service/job.rs
```diff
@@ -45,12 +46,37 @@ impl JobService for DbConn {
...
+            .await?;
+        let mut stats = JobStats {
+            success: 0,
+            failed: 0,
+            pending: 0,
+        };
+
+        for job in jobs {
```
wsxiaoys, line 66: should this be a aggregation query (`SELECT sum(success), ...`) in sql?

wsxiaoys (changes requested):
revoke approval

ee/tabby-webserver/src/service/job.rs
boxbeam: No, there isn't a good way to do this - it was the first thing I thought of, but it would basically end up needing to be broken up into three separate aggregation queries, since each query has its own condition. I believe that would be much slower than just querying once and performing the aggregation ourselves.

wsxiaoys: i don't get it - this query doesn't come with it's condition. so query is effectively as below:

```
SELECT sum(exit_code == 0), sum(exit_code == 1), sum(exit_code == null) FROM job_runs WHERE [job = :job]
```

boxbeam: I didn't think of that, and now that I'm looking at it, I realize it's not even a special syntax but just a consequence of booleans being treated as ints

wwayne:
API endpoints good to me

ee/tabby-db/migrations/0020_job-run-index.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX idx_job ON job_runs(job);
```
wsxiaoys, line 1:
```suggestion
CREATE INDEX idx_job_created_at ON job_runs(job, created_at);
```

ee/tabby-db/src/job_runs.rs
```diff
@@ -61,17 +69,26 @@ impl DbConn {
...
             let ids: Vec<String> = ids.iter().map(i32::to_string).collect();
             let ids = ids.join(", ");
-            Some(format!("id in ({ids})"))
-        } else {
-            None
-        };
+            conditions.push(format!("id in ({ids})"));
+        }
+
+        if let Some(jobs) = jobs {
```
wsxiaoys, line 85: add test coverage

ee/tabby-db/src/job_runs.rs
```diff
@@ -96,6 +113,31 @@ impl DbConn {
...
+        };
+
+        let cutoff = Utc::now() - Duration::days(7);
+
+        let stats = sqlx::query_as(&format!(
+            r#"SELECT
+                SUM(exit_code == 0) AS success,
+                SUM(exit_code != 0 AND exit_code IS NOT NULL) AS failed,
+                SUM(exit_code IS NULL) AS pending FROM job_runs
+                WHERE start_ts > ? {condition};"#
```
wsxiaoys, line 133:
```suggestion
                WHERE created_at > ? {condition};"#
```

column `start_ts` is a mistake, good to remove in future refactors.

wsxiaoys (approved):
Otherwise LGTM

ee/tabby-db/src/job_runs.rs
boxbeam: Any reason we don't remove it?

wsxiaoys: it's just better to do it in a separate refactor PR. can be removed anytime

PR #1726, opened 2024-03-26
ci: support Vulkan build on Windows

vgf89:
I tried compiling the Vulkan feature on Windows, but was running into the vulkan library not being found by the build script. This is the fix. Now I can run tabby on my AMD 5700XT on Windows. It's seems quite fast too!

Changes:
* Remove `=native` in vulkan feature in build.rs so that all library kinds are part of the search. This *shouldn't* break the Linux builds, but if it does CI should catch it.
* Add a CI step to create a Windows Vulkan build automatically. If I'm lucky it'll work first try? 🤞
* User-facing documentation changes to reflect Vulkan support.

### DIFF
--- a/website/docs/installation/windows/index.mdx
+++ b/website/docs/installation/windows/index.mdx
@@ -12,7 +12,8 @@ Running Tabby on Windows using Tabby's exe distribution.
 
 ## 2. Download the release
 * If you are using a CPU-only system, download the **tabby_x86_64-windows-msvc.exe file**.
-* If you are using a GPU-enabled system, download the **tabby_x86_64-windows-msvc-cuda117.exe** file, In this example, we assume you are using CUDA 11.7.
+* If you are using an Nvidia GPU, download the **tabby_x86_64-windows-msvc-cuda117.exe** file, In this example, we assume you are using CUDA 11.7.
+* If you are using a different GPU with Vulkan support, download the **tabby_x86_64-windows-msvc-vulkan.exe** file.
 
 **Tips:**
 * Download the CUDA Toolkit from Nvidia: https://developer.nvidia.com/cuda-toolkit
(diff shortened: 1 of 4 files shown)


vgf89:
Ah, yeah I had to copy the vulkan-1.lib to vulkan.lib. I need to see if I can fix issue build-side instead of manually

vgf89:
It looks like the vulkan build is currently broken on Linux due to an upstream issue with the manylinux docker container. It's trying to pull the completely wrong (outdated?) URL format for the vulkan sdk.

wsxiaoys:
cc @icycodes

icycodes:
> It looks like the vulkan build is currently broken on Linux CI due to an upstream issue with the manylinux docker container. It's trying to pull the completely wrong (outdated?) URL format for the vulkan sdk.

@vgf89  Thanks, I confirmed that the artifact built by the windows pipeline is working well on my windows machine.   
Looking into the issue on the Linux CI pipeline.

icycodes:
> > It looks like the vulkan build is currently broken on Linux CI due to an upstream issue with the manylinux docker container. It's trying to pull the completely wrong (outdated?) URL format for the vulkan sdk.
> 
> @vgf89 Thanks, I confirmed that the artifact built by the windows pipeline is working well on my windows machine. Looking into the issue on the Linux CI pipeline.

@vgf89 @wsxiaoys 
It appears that the SDK version used in the Linux pipeline has recently been removed from https://vulkan.lunarg.com/sdk/home. The cached download file was being used before this PR. 
I believe this PR can be safely merged without considering the failed Linux CI pipeline. I will address this in the next PR.

website/docs/installation/windows/index.mdx
wsxiaoys, line 1: let's revert the change of this file since it won't be released until next ver (0.10)

wsxiaoys (approved):
Otherwise LGTM

PR #1728, opened 2024-03-26
fix(ci): update vulkan sdk version for linux.

icycodes:


icycodes:
It seems that the Vulkan SDK version 1.3.239.0 (30 Jan 2023) is the latest version that is compatible with manylinux2014. Starting from version 1.3.243.0 (29 Mar 2023), it requires glibc 2.29.

PR #1731, opened 2024-03-26
feat(webserver): Implement file search and meta query

boxbeam:
Fixes TAB-547
cc @liangfung


ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -287,6 +290,31 @@ impl Query {
...
+        filter: Option<String>,
+        top_n: i32,
+    ) -> FieldResult<Vec<FileEntry>> {
+        Ok(ctx
+            .locator
+            .repository()
+            .search_files(repository, filter.unwrap_or_default(), top_n as usize)
+            .await?)
+    }
+
```
wsxiaoys, line 305:
```suggestion
    ) -> FieldResult<Vec<FileEntry>> {
        Ok(ctx
            .locator
            .repository()
            .search_files(repository, filter.unwrap_or_default(), 40)
            .await?)
    }

```

ee/tabby-webserver/src/service/repository.rs
```diff
@@ -25,18 +37,57 @@ impl RepositoryService for DbConn {
...
         Ok(true)
     }
+
+    async fn search_files(
+        &self,
+        name: String,
+        path_glob: String,
+        top_n: usize,
+    ) -> Result<Vec<FileEntry>> {
+        let git_url = self.get_repository_git_url(name).await?;
```
wsxiaoys, line 63: might use a filesystem based matching iterator and take top_n.

e.g https://crates.io/crates/ignore

ee/tabby-webserver/src/service/repository.rs
```diff
@@ -25,18 +37,57 @@ impl RepositoryService for DbConn {
...
+                } else {
+                    None
+                }
+            })
+            .take(top_n)
+            .collect();
+        Ok(matching)
+    }
+
+    async fn repository_meta(&self, name: String, path: String) -> Result<RepositoryMeta> {
```
wsxiaoys, line 81: Please refactor https://github.com/TabbyML/tabby/blob/07f13f5ce94f03ad9b4207a96af970d64d9d3726/ee/tabby-webserver/src/repositories/resolve.rs#L28 to make it shared between this service and resolve.rs

wsxiaoys (changes requested):

wsxiaoys:
Consider add `#[deprecated]` to https://github.com/TabbyML/tabby/blob/07f13f5ce94f03ad9b4207a96af970d64d9d3726/ee/tabby-webserver/src/repositories/mod.rs#L102

ee/tabby-webserver/src/service/repository.rs
```diff
@@ -1,13 +1,25 @@
...
 use super::{graphql_pagination_to_filter, AsID, AsRowid};
 use crate::schema::{
-    repository::{Repository, RepositoryService},
+    repository::{FileEntry, Repository, RepositoryMeta, RepositoryService},
     Result,
 };
 
+fn glob_matches(glob: &str, mut input: &str) -> bool {
+    for part in glob.split('*') {
+        let Some((_, rest)) = input.split_once(part) else {
```
liangfung, line 15: Changing the matching algorithm to match non-contiguous substrings might be better? The effect is as shown in the following image.
image

wsxiaoys: https://github.com/TabbyML/tabby/pull/1731#discussion_r1540224477

boxbeam: This is the behavior of the current algorithm if you substitute `*` for spaces. I can adjust it to split on spaces instead.

ee/tabby-webserver/src/service/repository.rs
boxbeam: Since we can now match against the in-memory map, I don't think this would be an improvement anymore

ee/tabby-webserver/src/service/repository.rs
```diff
@@ -1,13 +1,59 @@
...
 use crate::schema::{
-    repository::{Repository, RepositoryService},
+    repository::{FileEntry, Repository, RepositoryMeta, RepositoryService},
     Result,
 };
 
+fn match_glob(path: &str, glob: &str) -> bool {
+    let glob = glob.to_lowercase();
+    let mut path = &*path.to_lowercase();
+    // Current behavior: Find each "word" (any substring separated by whitespace) appearing in the same order
```
boxbeam, line 19: @wsxiaoys Tried using `ignore` but couldn't get it behaving as I wanted, so I switched to `async_walkdir` with my own glob filter instead. This comment explains its behavior, and I can show some screenshots to demonstrate how it works:
image
image
image

boxbeam: cc @liangfung

ee/tabby-db/schema.sqlite
wsxiaoys, line 1: revert

ee/tabby-db/src/repositories.rs
```diff
@@ -64,6 +64,13 @@ impl DbConn {
             Err(anyhow!("failed to update: repository not found"))
         }
     }
+
+    pub async fn get_repository_git_url(&self, name: String) -> Result<String> {
```
wsxiaoys, line 68: add test coverage

ee/tabby-webserver/src/service/repository.rs
wsxiaoys: 1. I don't understand why it cannot be using ignore - since you're implementing your own glob match anyway.
2. Please also returns directories.

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -296,6 +297,30 @@ impl Query {
...
+        ctx: &Context,
+        repository: String,
+        path: String,
+    ) -> FieldResult<RepositoryMeta> {
+        Ok(ctx
+            .locator
+            .repository()
+            .repository_meta(repository, path)
+            .await?)
+    }
```
wsxiaoys, line 322:
```suggestion
```

ee/tabby-webserver/src/service/repository.rs
```diff
@@ -25,18 +71,47 @@ impl RepositoryService for DbConn {
...
+    async fn repository_meta(&self, name: String, path: String) -> Result<RepositoryMeta> {
+        let git_url = self.get_repository_git_url(name).await?;
+        SourceFile::all()
+            .map_err(anyhow::Error::from)?
+            .filter_map(|file| {
+                (file.filepath == path && file.git_url == git_url).then(move || file.into())
+            })
+            .next()
+            .ok_or_else(|| anyhow!("File not found").into())
+    }
```
wsxiaoys, line 114:
```suggestion
```

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -296,6 +297,30 @@ impl Query {
         .await
     }
 
+    async fn repository_search(
+        ctx: &Context,
+        repository: String,
+        filter: Option<String>,
```
wsxiaoys, line 303:
```suggestion
        filter: String,
```

ee/tabby-webserver/src/schema/mod.rs
```diff
@@ -296,6 +297,30 @@ impl Query {
...
+    ) -> FieldResult<Vec<FileEntry>> {
```
wsxiaoys, line 304: if `filter`.trim() is empty, return empty list of files.

ee/tabby-webserver/src/service/repository.rs
boxbeam: 1. ignore does not allow for custom globs for matching, only ignoring files as far as I can tell
2. ignore is not async, and so would block the entire server's async runtime while it is running a search

wsxiaoys: 1. As said - glob is implemented as customized filter anyway
2. tokio manage its requests in multi threads async runtime. Thus won’t block

ee/tabby-db/schema.sqlite
wsxiaoys, line 1: revert

ee/tabby-db/src/repositories.rs
wsxiaoys, line 68:
```suggestion
    pub async fn get_repository_by_name(&self, name: String) -> Result<RepositoryDAO> {
```

Also make sure we add an index on `name`

PR #1699, opened 2024-03-19
feat(api): add relevant declarations in completion request.

icycodes:
Complete TAB-515

Changes:
Add declarations in segments for completion requests.

PR #1742, opened 2024-03-29
feat(core): add relevant snippets in completion request.

icycodes:
Complete TAB-515
Replace #1699

### DIFF
--- a/crates/tabby/src/services/completion.rs
+++ b/crates/tabby/src/services/completion.rs
@@ -109,8 +109,8 @@ pub struct Segments {
     suffix: Option<String>,
 
     /// The relative path of the file that is being edited.
-    /// When git_url is set, this is the path of the file in the git repository.
-    /// When git_url is empty, this is the path of the file in the workspace.
+    /// - When `git_url` is set, this is the path of the file in the git repository.
+    /// - When `git_url` is empty, this is the path of the file in the workspace.
     filepath: Option<String>,
 
     /// The remote URL of the current git repository.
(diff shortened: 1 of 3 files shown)


crates/tabby/src/services/completion.rs
```diff
@@ -109,19 +109,55 @@ pub struct Segments {
...
     filepath: Option<String>,
 
     /// The remote URL of the current git repository.
     /// Leave this empty if the file is not in a git repository,
     /// or the git repository does not have a remote URL.
     git_url: Option<String>,
 
+    /// The relevant code snippets provided by editor.
+    relevant_snippets: Option<Vec<RelevantSnippet>>,
+
```
wsxiaoys, line 123:
```suggestion
    /// The relevant code snippets provided by editor.
    declarations: Option<Vec<Declaration>>,
    
  struct Declaration {
    filepath: string
    body: string
  }
```

Please don't reuse type for different fields. Let's use simple declaration mechanism for now.

crates/tabby/src/services/completion.rs
```diff
@@ -109,19 +109,38 @@ pub struct Segments {
...
+    /// - When `git_url` is empty, this is the path of the file in the workspace.
     filepath: Option<String>,
 
     /// The remote URL of the current git repository.
     /// Leave this empty if the file is not in a git repository,
     /// or the git repository does not have a remote URL.
     git_url: Option<String>,
 
+    /// The relevant declaration code snippets provided by editor.
+    /// The editor could invoke the language server to get relevant declarations.
```
wsxiaoys, line 122:
```suggestion
    /// The relevant declaration code snippets provided by editor.
    /// It'll contains declarations extracted from `prefix` segments using LSP.
```

crates/tabby/src/services/completion.rs
```diff
@@ -109,19 +109,38 @@ pub struct Segments {
...
+
     /// Clipboard content when requesting code completion.
     clipboard: Option<String>,
 }
 
+/// A snippet of declaration code that is relevant to the current completion request.
+#[derive(Serialize, Deserialize, ToSchema, Clone, Debug)]
+pub struct Declaration {
+    /// Filepath of the file where the snippet is from.
+    /// This filepath is never the same as the current file (`segments.filepath`).
```
wsxiaoys, line 133:
```suggestion
```
It can be from same file as long as the declaration is not included in prefix / suffix.

crates/tabby/src/services/completion.rs
icycodes: But declarations is extracted from other files, not from the prefix.

crates/tabby/src/services/completion.rs
icycodes: I think the prefix/suffix should be append to including the declaration in that case.

PR #1748, opened 2024-04-01
feat(ui): add chat panel to code browser

liangfung:
Close TAB-575
https://jam.dev/c/d90f72ea-571c-4cf0-8485-54eaa46fefde


liangfung:
cc @wwayne

ee/tabby-ui/app/files/components/blob-header.tsx
```diff
@@ -120,6 +127,23 @@ export const BlobHeader: React.FC<BlobHeaderProps> = ({
                 <TooltipContent>Download raw file</TooltipContent>
               </Tooltip>
             )}
+            {!enableCodeBrowserQuickActionBar.loading &&
+              enableCodeBrowserQuickActionBar.value &&
```
wsxiaoys, line 131:
```suggestion
            { enableCodeBrowserQuickActionBar.value &&
```

also need to check that chat playground is enabled.

wsxiaoys (changes requested):

PR #1760, opened 2024-04-02
chore(db): Update UserDAO to use query_as! and use i64 IDs in most pl…

boxbeam:
Closes TAB-521


wsxiaoys:
Please resolve conflicts

PR #1761, opened 2024-04-03
feat(agent&vscode): support collecting declarations for completion request.

icycodes:


clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -70,6 +77,11 @@ export const defaultAgentConfig: AgentConfig = {
       experimentalStripAutoClosingCharacters: false,
       maxPrefixLines: 20,
       maxSuffixLines: 20,
+      experimentalDeclarations: {
+        enabled: false,
+        maxSnippets: 5,
+        maxChars: 2000,
```
wsxiaoys, line 83: this shall be applied on each individual snippet (e.g 500), rather than total max chars

clients/tabby-agent/src/AgentConfig.ts
```diff
@@ -70,6 +77,11 @@ export const defaultAgentConfig: AgentConfig = {
       experimentalStripAutoClosingCharacters: false,
       maxPrefixLines: 20,
       maxSuffixLines: 20,
+      experimentalDeclarations: {
```
wsxiaoys, line 80:
```suggestion
      fillDeclarations: {
```

This won't go through a experimental launch approach, just name it as ordianry flag.

wsxiaoys (commented):
Otherwise LGTM

PR #1765, opened 2024-04-04
feat(webserver): add repository file search api

wsxiaoys:
fix TAB-547


ee/tabby-webserver/src/service/repository.rs
```diff
@@ -37,6 +41,62 @@ impl RepositoryService for DbConn {
             .await?;
         Ok(true)
     }
+
+    async fn search_files(
```
wsxiaoys, line 45: If implementation makes sense to you, please add a unit test and send PR to this branch @boxbeam

ee/tabby-webserver/src/service/repository.rs
```diff
@@ -37,6 +41,62 @@ impl RepositoryService for DbConn {
...
+    limit: usize,
+) -> Result<Vec<FileEntry>, anyhow::Error> {
+    let mut nucleo = nucleo::Matcher::new(nucleo::Config::DEFAULT.match_paths());
+    let needle = nucleo::pattern::Pattern::new(
+        pattern,
+        nucleo::pattern::CaseMatching::Ignore,
+        nucleo::pattern::Normalization::Smart,
+        nucleo::pattern::AtomKind::Fuzzy,
+    );
+    let mut scored_entries: Vec<(_, _)> = Walk::new(base)
```
boxbeam, line 75: This is a synchronous file walk, and will stall the server. The matching style looks interesting, and may be an improvement, but this is the weakness of the `ignore` crate that I had been talking about.

ee/tabby-webserver/src/service/repository.rs
```diff
@@ -37,6 +41,62 @@ impl RepositoryService for DbConn {
...
+                .into_path()
+                .strip_prefix(base)
+                .ok()?
+                .to_string_lossy()
+                .into_owned();
+            let haystack: nucleo::Utf32String = path.clone().into();
+            let score = needle.score(haystack.slice(..), &mut nucleo);
+            score.map(|score| (score, FileEntry { r#type, path }))
+        })
+        .take(limit)
```
boxbeam, line 99: This selects values prior to sorting, which may cause badly-scored results to be the only ones returned if there are many matches.

boxbeam (changes requested):

ee/tabby-webserver/src/service/repository.rs
wsxiaoys: it is mostly a greedy prior - we might adjust it to something like 3xlimit in futute

PR #1782, opened 2024-04-08
fix(ui): prevent consecutive messages from same source

liangfung:
Close TAB-577


ee/tabby-ui/components/chat.tsx
```diff
@@ -206,4 +209,40 @@ function ChatRenderer(
   )
 }
 
+function mergeMessagesByRole(body: BodyInit | null | undefined) {
```
wsxiaoys, line 212: just move this to `use-patch-fetch.ts`? since tabby's backend always have this behavior (no consecutive same role message)

PR #1795, opened 2024-04-09
feat(core): added recent snippets in completion request segments.

icycodes:
TAB-512

Changes:
- Added `recent_snippets` in completion request segments.
- Updated comments using rustdoc links. (But it does not seem to be compiled to openapi doc links)

### DIFF
--- a/crates/tabby/src/services/completion.rs
+++ b/crates/tabby/src/services/completion.rs
@@ -122,8 +122,8 @@ pub struct Segments {
     declarations: Option<Vec<Declaration>>,
 
     /// The relevant code snippets extracted from recently edited files.
-    /// These snippets are selected from candidates found within a 
-    /// code chunk based on the edited location.
+    /// These snippets are selected from candidates found within code chunks
+    /// based on the edited location.
     /// The current editing file is excluded from the search candidates.
     ///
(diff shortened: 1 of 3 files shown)


crates/tabby/src/services/completion.rs
```diff
@@ -109,19 +109,28 @@ pub struct Segments {
...
-    /// It'll contains declarations extracted from `prefix` segments using LSP.
+    /// The relevant declaration code snippets provided by the editor's LSP,
+    /// contain declarations of symbols extracted from [Segments::prefix].
     declarations: Option<Vec<Declaration>>,
 
+    /// The relevant code snippets extracted from recently accessed files.
+    /// The current editing file is excluded from the search candidates.
+    ///
+    /// When provided alongside [Segments::declarations], the snippets should
+    /// be deduplicated to avoid duplication with entries in `declarations`.
```
wsxiaoys, line 129: you should say "it may overlapped with declarations". Instead of asking how it should be used, API should only describe objective fact.

icycodes: My description was wrong, I want to say the snippets have already been deduplicated.

PR #1798, opened 2024-04-09
refactor(webserver): decouple oauth/github route and integrations/git…

wsxiaoys:
…hub route


boxbeam (approved):
Fine by me if we are okay with the duplication. The only value used in the response is `access_token`, so maybe we can get rid of the rest?

wsxiaoys:
> Fine by me if we are okay with the duplication. The only value used in the response is `access_token`, so maybe we can get rid of the rest?

Yes - leave it as is for now. (might need to improve error message in future anyway)

PR #1809, opened 2024-04-10
refactor: hide the sidebar when the screen media width is less than lg

liangfung:


ee/tabby-ui/app/globals.css
```diff
@@ -54,7 +54,7 @@
     --card: 39 15% 27%;
     --card-foreground: 39 3.2% 99.35%;
 
-    --border: 39 13% 21%;
+    --border: 39 9% 17%;
```
liangfung, line 57: Adjust the border color, otherwise the border here is not visible.
image

PR #1821, opened 2024-04-11
fix(ui): small adjust of the jobs color in jobs page

wwayne:
Before:

Update:


wsxiaoys:
Please resolve conflict

PR #1822, opened 2024-04-11
feat(ui):  detect language by filename

liangfung:


ee/tabby-ui/app/files/lib/filename2prism/index.ts
wsxiaoys, line 1: move to `tabby-ui/lib/filename2prism`

PR #1825, opened 2024-04-11
feat(core): support relevant snippet from changed files

wsxiaoys:
cc @boxbeam


crates/tabby/src/services/completion/completion_prompt.rs
```diff
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+
```
wsxiaoys, line 3:
```suggestion
```

icycodes (approved):

PR #1837, opened 2024-04-12
fix(webserver): when compute daily stats for Other languages, it shou…

wsxiaoys:
…ld use OR logic operator

### DIFF
--- a/ee/tabby-db/src/user_completions.rs
+++ b/ee/tabby-db/src/user_completions.rs
@@ -145,7 +145,7 @@ impl DbConn {
         // with `STRFTIME('%s')`. The effect of this is to extract the unix timestamp (seconds) rounded to
         // the start of the day and group them by that.
         let res = sqlx::query_as(&format!(
-                r#"
+            r#"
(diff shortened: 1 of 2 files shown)


ee/tabby-db/src/user_completions.rs
```diff
@@ -149,11 +149,13 @@ impl DbConn {
             SELECT CAST(STRFTIME('%s', DATE(created_at)) AS TIMESTAMP) as start,
                    SUM(1) as completions,
                    SUM(selects) as selects
-            FROM user_completions
-            WHERE created_at >= ?1 AND created_at < ?2
-                AND ({no_selected_users} OR user_id IN ({users}))
-                AND (({no_selected_languages} OR language IN ({languages}))
-                     OR (language NOT IN ({not_languages})))
+            FROM (
+                SELECT created_at, selects, IIF(language IN ({all_languages}), language, 'other') as language
```
wsxiaoys, line 153:
```suggestion
                SELECT created_at, selects, IF(language IN ({all_languages}), language, 'other') as language
```

boxbeam: `IF` does not exist in sqlite, the function is called `IIF`

PR #1839, opened 2024-04-12
fix(scheduler): Canonicalize git URL in SourceFile rather than tantivy index

boxbeam:


crates/tabby-scheduler/src/dataset.rs
```diff
@@ -61,7 +61,7 @@ impl RepositoryExt for RepositoryConfig {
             match read_to_string(entry.path()) {
                 Ok(file_content) => {
                     let source_file = SourceFile {
-                        git_url: self.git_url.clone(),
+                        git_url: RepositoryConfig::canonicalize_url(&self.git_url),
```
wsxiaoys, line 64: self.canoncial git URL?

wsxiaoys (approved):

PR #1848, opened 2024-04-16
fix(db): Fix daily stats query

boxbeam:


wsxiaoys (changes requested):
Please add a test case that'll be failing without this patch as regression test.

ee/tabby-webserver/src/service/analytic.rs
```diff
@@ -183,7 +183,10 @@ mod tests {
         let svc = new_analytic_service(db);
         let end = Utc::now();
         let start = end.checked_sub_days(Days::new(100)).unwrap();
```
wsxiaoys, line 185: by doing so, there'll be no test case covering queries that select all user + all languages.

Please add a new test case instead of modifying existing one.

wsxiaoys:
Please rebase it against main

PR #1849, opened 2024-04-16
fix(ui): change the text to activities in the home page

wwayne:


ee/tabby-ui/app/(home)/components/stats.tsx
```diff
@@ -367,7 +367,7 @@ export default function Stats() {
       >
         <div>
           <h3 className="mb-2 text-sm font-medium tracking-tight">
-            <b>{lastYearCompletions}</b> contributions in the last year
+            <b>{lastYearCompletions}</b> activities in the last year
```
wsxiaoys, line 370: should this be changed to `lastYearActivities`, where activities counts all completions + selections?

wwayne: Added, I feel that both definitions are fine.

PR #1860, opened 2024-04-16
feat(core): turn on --webserver by default

wsxiaoys:
fix TAB-598

### DIFF
--- a/crates/tabby/src/serve.rs
+++ b/crates/tabby/src/serve.rs
@@ -106,7 +106,7 @@ pub struct ServeArgs {
     parallelism: u8,
 
     #[cfg(feature = "ee")]
-    #[clap(long, default_value_t = false)]
+    #[clap(hide = true, long, default_value_t = false)]
     #[deprecated(since = "0.11.0", note = "webserver is enabled by default")]
     webserver: bool,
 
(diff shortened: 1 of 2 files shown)


crates/tabby/src/serve.rs
```diff
@@ -106,8 +106,13 @@ pub struct ServeArgs {
     parallelism: u8,
 
     #[cfg(feature = "ee")]
-    #[clap(hide = true, long, default_value_t = false)]
+    #[clap(hide = true, long, default_value_t = true)]
+    #[deprecated]
```
boxbeam, line 110: Require that this flag cannot be combined with `no_webserver` and issue a deprecation warning on CLI when using this flag

boxbeam (changes requested):

crates/tabby/src/serve.rs
wsxiaoys: this flag is no longer used (that's why it's marked as deprecated).

The only purpose of keeping it to avoid tabby throws an error when somewhy still pass `--webserver` to it. (forward compatibility)

boxbeam: Think we should at least add a CLI deprecation warning, in case some script is still using this automatically it will inform them that it may eventually be removed

boxbeam (approved):

PR #1877, opened 2024-04-17
feat: add CSS stop configuration

nannoda:


crates/tabby-common/assets/languages.toml
```diff
@@ -212,3 +212,34 @@ top_level_keywords = [
     "function",
     "type",
 ]
+
+[[config]]
+languages = ["css"]
+line_comment = "/*"
```
wsxiaoys, line 218: this won't work given how we construct line comments string for building context into prompt... (`line_comment + code`)

🤔

nannoda: I'm sorry about that, I haven't looked into the implementation. I will try to fix it.

wsxiaoys: No worries - this is indeed a flaw of context building implementation. Thanks for providing such a case to initiate the improvements

wsxiaoys, line 218:
```suggestion
```

crates/tabby-common/assets/languages.toml
```diff
@@ -212,3 +212,33 @@ top_level_keywords = [
     "function",
     "type",
 ]
+
+[[config]]
+languages = ["css"]
```
wsxiaoys, line 217:
```suggestion
languages = ["css"]
exts = ["css"]
```

PR #1887, opened 2024-04-18
fix(core): Improve string matching for git_url to better approximate repository matches

wsxiaoys:


crates/tabby/src/services/code.rs
```diff
@@ -151,27 +152,40 @@ fn closest_match<'a>(
...
-                // Example:
-                // haystack = "https://github.com/boxbeam/untwine" needle = "https://abc@github.com/boxbeam/untwine.git" => No match
-                // haystack = "https://abc@github.com/boxbeam/untwine.git" needle = "https://github.com/boxbeam/untwine" => Match, score 842
-                nucleo.fuzzy_match(
-                    Utf32String::from(&*search_term).slice(..),
-                    Utf32String::from(&*url).slice(..),
-                )?,
-            ))
+        .filter_map(|elem| {
+            let Ok(git_url) = GitUrl::parse(&elem.git_url) else {
```
wsxiaoys: what if url doesn't come with owner? can you add a test case for it? e.g `http://abc.com/def`

crates/tabby/src/services/code.rs
```diff
@@ -151,27 +152,40 @@ fn closest_match<'a>(
...
+            } else {
+                score += 30;
+            }
+
+            // Owner is scored the lowest, to allow forks to be matched as long as the rest matches
+            if let Some((org1, org2)) = git_search.owner.as_ref().zip(git_url.owner) {
+                score += 3 * damerau_levenshtein(org1, &org2);
+            } else {
+                score += 15;
+            }
```
wsxiaoys, line 177: as discussed, don't use scoring in first step. Find suffix matching (until first deliminator) to find candidates, and only do scoring if there're multiple candidates

PR #1900, opened 2024-04-20
feat(webserver): slice analytics by programming languages to reduce g…

wsxiaoys:
…raphql api call in frontend

### DIFF
--- a/ee/tabby-webserver/src/schema/analytic.rs
+++ b/ee/tabby-webserver/src/schema/analytic.rs
@@ -37,33 +37,11 @@ pub enum Language {
 }
 
 lazy_static! {
-    static ref LANGUAGE_STRING_MAPPINGS: HashMap<Language, Vec<&'static str>> = {
+    static ref NAME_LANGUAGE_MAPPINGS: HashMap<&'static str, Language> = {
         let mut map = HashMap::new();
         for language in Language::iter() {
-            let value = match language {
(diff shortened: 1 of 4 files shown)


ee/tabby-webserver/src/schema/analytic.rs
```diff
@@ -31,29 +35,58 @@ pub enum Language {
     Other,
 }
 
+lazy_static! {
+    static ref LANGUAGE_STRING_MAPPINGS: HashMap<Language, Vec<&'static str>> = {
```
boxbeam, line 39: Would rather use a derive or similar for this if possible - it will be very easy to forget to add a value here when a new Language is added

wsxiaoys: switched to a match based clause

ee/tabby-webserver/src/schema/analytic.rs
```diff
@@ -32,29 +36,63 @@ pub enum Language {
...
+        let mut map = HashMap::new();
+        for language in Language::iter() {
+            let value = match language {
```
boxbeam, line 43: Extract into function and use `Language::iter().map(|l| (l.clone(), l.language_names())).collect()`

boxbeam (approved):
Otherwise LGTM

boxbeam (approved):

wwayne (approved):

PR #1923, opened 2024-04-22
feat(db): implement user events DAO and core logic

boxbeam:
Relates to TAB-595


ee/tabby-db/migrations/0025_user-events.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE user_events (
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    user_id INTEGER NOT NULL,
+    completion_id TEXT NOT NULL,
+    type TEXT NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT (DATETIME('now')),
+    payload TEXT NOT NULL
```
wsxiaoys, line 7: json field

ee/tabby-webserver/src/service/event_logger.rs
```diff
@@ -70,8 +101,23 @@ impl EventLogger for DbEventLogger {
...
+                            x.ts,
+                            user_db.id,
+                            completion_id.clone(),
+                            language,
+                        )
+                        .await,
+                    );
+                    log_err(
+                        db.create_user_event(
+                            user_db.id,
```
wsxiaoys, line 114: since id is in side of `user` field, we don't need to do a lookup with `db.get_user` to ensure it exist - we can just insert the event and let foreign constraint should help us filter the problematic requests.

due to this reason, we don't really need the `create_user_event_lookup_user` method.

ee/tabby-db/migrations/0025_user-events.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE user_events (
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    user_id INTEGER NOT NULL,
+    completion_id TEXT NOT NULL,
```
wsxiaoys, line 4:
```suggestion
```

No need to store completion_id as column. it shall be part of `payload`

ee/tabby-webserver/src/service/event_logger.rs
```diff
@@ -23,14 +23,27 @@ pub fn create_event_logger(db: DbConn) -> impl EventLogger + 'static {
...
                 tokio::spawn(async move {
                     log_err(
                         db.add_to_user_completion(x.ts, &completion_id, 1, 0, 0)
                             .await,
-                    )
+                    );
+                    log_err(
+                        db.create_user_event_lookup_user(
+                            completion_id,
+                            "view".into(),
```
wsxiaoys, line 41: define `DbEnum` for `Event`.

wsxiaoys (changes requested):

ee/tabby-db/migrations/0025_user-events.up.sql
boxbeam: SQLite has no JSON type

ee/tabby-webserver/src/service/event_logger.rs
boxbeam: These are always strings at the db layer, the `DbEnum` is used at the service layer

wsxiaoys: Exactly - thus we should define DbEbum for Event and invoke as_db_str here?

ee/tabby-db/migrations/0025_user-events.up.sql
wsxiaoys: https://www.sqlite.org/changes.html#version_3_45_0

jsonb column has landed since 3.45

boxbeam: image
Using sqlite version 3.45.3

wsxiaoys: the data column for it should just be blob?

ee/tabby-db/migrations/0025_user-events.up.sql
```diff
@@ -1,11 +1,9 @@
 CREATE TABLE user_events (
     id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
-    completion_id TEXT NOT NULL,
-    user_id INTEGER NOT NULL,
+    user_name TEXT NOT NULL,
```
wsxiaoys, line 3:
```suggestion
    user_id INTEGER NOT NULL,
```

the `user` string in header is simply `ID`.

ee/tabby-webserver/src/schema/user_event.rs
wsxiaoys, line 1: just define it in `schema/mod.rs` if there's no better place.

wsxiaoys (changes requested):

ee/tabby-db/migrations/0025_user-events.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE user_events (
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    user_id INTEGER NOT NULL,
+    type TEXT NOT NULL,
```
wsxiaoys, line 4:
```suggestion
    kind TEXT NOT NULL,
```

Avoid aliasing and matching `EventKind`

ee/tabby-webserver/src/service/event_logger.rs
```diff
@@ -21,16 +23,35 @@ pub fn create_event_logger(db: DbConn) -> impl EventLogger + 'static {
     DbEventLogger { db }
 }
 
+fn get_user_id(user: Option<String>) -> Option<i64> {
+    user.and_then(|id| ID::new(id).as_rowid().ok())
+}
+
 impl EventLogger for DbEventLogger {
     fn write(&self, entry: LogEntry) {
```
wsxiaoys, line 31: i feel this function could benefit with a refactoring adding an `write_impl(...) -> Result` for easier `?` chaining syntax to reduce some verbosity.

feel free to do this as a follow up by adding a FIXME

boxbeam: https://linear.app/tabbyml/issue/TAB-629/refactor-webserverserviceevent-loggerrs-write-impl

PR #1937, opened 2024-04-23
feat(tabby): include user for all completion event types

boxbeam:


crates/tabby-common/src/api/event.rs
```diff
@@ -114,6 +113,7 @@ impl From<Event> for LogEntry {
     fn from(event: Event) -> Self {
         Self {
             ts: timestamp(),
+            user: None,
```
wsxiaoys, line 116: is there other place use this `From<Event>` trait? if not, we can just remove it.

crates/tabby/src/header.rs
wsxiaoys, line 1: move it to `routes/mod.rs` and make it public within path

wsxiaoys (approved):
Otherwise LGTM

PR #1943, opened 2024-04-23
refactor(db): cleanup db tables, fix foreign key for password_reset and refresh_tokens

wsxiaoys:


ee/tabby-db/schema/schema.svg
boxbeam, line 1: Should a diff be committed along with the changes? Feel like this should be an automatic response in a comment, not a file committed to the repo

wsxiaoys: only the updated svg will be checked in. the diff is a github enhancement

PR #1952, opened 2024-04-24
ci: add lint to check around all DateTime<Utc> usage in tabby-db

wsxiaoys:
image


boxbeam (changes requested):
Also add lint for usage of `Utc::now()`, or `DateTime<Utc>` being used as a parameter type - they should be replaced with `DateTimeUtc::now()` and `impl Into<DateTimeUtc>` respectively

PR #1955, opened 2024-04-24
test(webserver): Add test for listing events

boxbeam:
This will fail until #1952 is merged and rebased onto this branch


wsxiaoys:
dup with https://github.com/TabbyML/tabby/pull/1956

PR #1965, opened 2024-04-25
feat(webserver): support gitlab integration with automatic repository sync

boxbeam:
Closes TAB-594
cc @liangfung


ee/tabby-db/migrations/0028_gitlab-provider.up.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE gitlab_repository_provider(
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    display_name TEXT NOT NULL,
+    access_token TEXT
```
wsxiaoys, line 4: add a field `connected` to present its status.

1. When it's first created, default to `false`.
2. When cron job complete successfully, set it to `true`.
3. When cron job has error and access token got reset, set it to `false`.

The same logic should be added to github repository provider as well, but we can do it in a separate PR.

ee/tabby-webserver/src/schema/gitlab_repository_provider.rs
```diff
@@ -0,0 +1,136 @@
...
+use juniper::{GraphQLInputObject, GraphQLObject, ID};
+use lazy_static::lazy_static;
+use regex::Regex;
+use validator::Validate;
+
+use super::Context;
+use crate::{juniper::relay::NodeType, schema::Result};
+
+lazy_static! {
+    static ref GITLAB_REPOSITORY_PROVIDER_NAME_REGEX: Regex = Regex::new("^[\\w-]+$").unwrap();
```
wsxiaoys, line 12: consider extract this regex to some where (e.g schema/constants.rs) and reuse it between:
1. Github display name
2. Gitlab display name
3. Git name

ee/tabby-webserver/src/schema/gitlab_repository_provider.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+#[derive(GraphQLObject, Debug, PartialEq)]
+#[graphql(context = Context)]
+pub struct GitlabRepositoryProvider {
+    pub id: ID,
+    pub display_name: String,
+
+    pub connected: bool,
+
```
wsxiaoys, line 39:
```suggestion
```

as discussed in db schema, add a new enum field `status` with three values:
1. Ready: access_token != empty && connected.
2. Pending: access_token != empty && !connected.
3. Error: access_token == empty.

Same logic should add to github as well, but it can be done in a follow up pr

wsxiaoys (changes requested):
Otherwise LGTM

ee/tabby-db/migrations/0028_gitlab-provider.up.sql
boxbeam: TAB-634

wsxiaoys (approved):

PR #1993, opened 2024-04-28
refactor(webserver): extract JobController

wsxiaoys:
fix TAB-645

image

image

### DIFF
--- a/ee/tabby-webserver/src/cron/db/gitlab.rs
+++ b/ee/tabby-webserver/src/cron/db/gitlab.rs
@@ -8,25 +8,38 @@ use gitlab::{
 };
 use juniper::ID;
 use serde::Deserialize;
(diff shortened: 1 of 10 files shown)


ee/tabby-webserver/src/cron/db/gitlab.rs
```diff
@@ -37,18 +50,25 @@ async fn refresh_repositories_for_provider(
             service
                 .update_provider_status(provider.id.clone(), false)
                 .await?;
-            warn!(
-                "GitLab credentials for provider {} are expired or invalid",
-                provider.display_name
-            );
+            context
```
wsxiaoys, line 53: @boxbeam could you help make a macro that write to `warn!` and `stderr_writeline` at same time?

PR #2018, opened 2024-04-30
fix(webserver): update messages and disable registration for demo mode

boxbeam:


ee/tabby-webserver/src/service/auth.rs
```diff
@@ -519,6 +522,9 @@ async fn get_or_create_oauth_user(
         .map_err(|x| OAuthError::Other(x.into()))?
         .can_register_without_invitation(email)
     {
+        if demo_mode() {
```
wsxiaoys, line 525: This is for oauth based registeration - also disable invitation based register?

wsxiaoys (changes requested):

PR #2063, opened 2024-05-06
fix(scheduler): add absolute path as part of cache key

wsxiaoys:


crates/tabby-scheduler/src/cache.rs
```diff
@@ -88,7 +118,8 @@ impl CacheStore {
             .iter()
             .filter_map(|item| {
                 let item = item.expect("Failed to read item");
-                if is_item_key_matched(&item) {
+                let item_key: String = item.key().expect("Failed to get key");
+                if is_item_key_matched(&item_key) {
                     num_keep += 1;
                     None
                 } else {
```
wsxiaoys, line 125: for incremental tantivy indexing, this is the place where we want to delete `file_id` from repository. (we might just use absolute path as file_id key).

boxbeam: Do we want to do it inside this method?

Issue #791, opened 2023-11-15
A crawler takes a curated dev docs endpoint


wsxiaoys:
Reference: https://github.com/getcursor/docs/blob/main/docs.jsonl

May not necessarily be something need to reimplement, lots of tools in this space:

- https://github.com/projectdiscovery/katana
- https://github.com/rivermont/spidy

wsxiaoys:
Seems following katana snippet does the trick

```
katana -u https://codemirror.net/docs/ -cs docs -fs fqdn -d 1000 -j > output.json

-u: craw starting url
-cs: url path prefix scoping
-fs fqdn: restrict the same domain.
-d: depth
-j: json line output
```

wsxiaoys:
related: https://www.fleet.so/context collect all docs for python

akash4566:
[
  {
    "id": "859e8dff-f9ec-497d-aa07-344e48b2f67b",
    "score": 0.848275101,
    "metadata": {
      "library_id": "4506492b-70de-49f1-ba2e-d65bd7048a28",
      "page_id": "732e264c-c077-4978-bc93-380d7dc28983",
      "parent": "3be9bbcc-b5d6-4a91-9f72-a570c2db33e5",
      "text": "Quickstart | 🦜️🔗 Langchain",
      "url": "https://python.langchain.com/docs/get_started/quickstart"
    }
  },

wsxiaoys:
this will be part of 0.13 release

wsxiaoys:
cc @boxbeam

wsxiaoys:
Released as `web` context provider in 0.13

PR #2066, opened 2024-05-07
feat(scheduler): add crawler to scheduler

wsxiaoys:
related #791


crates/tabby-scheduler/src/crawl/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+use futures::{Stream, StreamExt};
+use readable_readability::Readability;
+use tokio::io::AsyncBufReadExt;
+use tracing::warn;
+use url::Url;
+
+use self::types::{CrawledDocument, KatanaRequestResponse};
+
+async fn crawl_url(start_url: &str) -> impl Stream<Item = KatanaRequestResponse> {
+    let mut child = tokio::process::Command::new("katana")
```
boxbeam, line 15: Can't use something like `spider` crate?

boxbeam:
Assuming the goal here is to crawl docs for indexing/context purposes?

crates/tabby-scheduler/src/crawl/mod.rs
wsxiaoys: It's still in the prototyping phase—primarily because I don't want to limit such a complex component to a monolithic Rust stack. However, I might change my mind in later stages, depending on whether we have sufficient third-party library support.

PR #2069, opened 2024-05-07
fix(vim): fix agent exit wrong type bug

andreabak:
In the vim plugin, whenever the tabby-agent exits (unexpectedly) the s:tabby variable is set to `{}` instead of the numeric `0` to indicate that the agent job is not running.

This leads to the agent functions trying to send messages anyway and error messages being spammed for almost any editor event, for the wrong variable type and subsequent errors, like:
```
line    8:
E731: using Dictionary as a String
line   12:
E474: Invalid argument
```
and
```
Error detected while processing BufLeave Autocommands for "*"... [...]
```

Setting the s:tabby variable to `0` on agent exit solves it.

### DIFF
--- a/clients/vim/autoload/tabby/agent.vim
+++ b/clients/vim/autoload/tabby/agent.vim
@@ -52,7 +52,7 @@ function! s:OnError(data)
 endfunction
 
 function! s:OnExit()
-  let s:tabby = {}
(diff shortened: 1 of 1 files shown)


icycodes:
@andreabak, Thank you for fixing this!

PR #2075, opened 2024-05-08
feat(db): create merged integration tables

boxbeam:


ee/tabby-db/src/integration_provider.rs
```diff
@@ -0,0 +1,258 @@
...
+            ProvidedRepositoryDAO,
+            "SELECT id, vendor_id, name, git_url, active, access_token_provider_id, created_at, updated_at FROM provided_repositories WHERE id = ?",
+            id
+        )
+        .fetch_one(&self.pool)
+        .await?;
+        Ok(repo)
+    }
+
+    pub async fn list_provided_repositories(
```
boxbeam, line 194: @wsxiaoys Should this have an additional filter for `kind`? You can still effectively do that by providing all the IDs for a given kind, but this might make it easier to use

ee/tabby-db/migrations/0029_merged-provider-tables.up.sql
```diff
@@ -0,0 +1,27 @@
...
+    updated_at TIMESTAMP NOT NULL DEFAULT (DATETIME('now')),
+    FOREIGN KEY (access_token_provider_id) REFERENCES integration_access_tokens(id),
+    CONSTRAINT idx_unique_provider_id_vendor_id UNIQUE (access_token_provider_id, vendor_id)
+);
+
+INSERT INTO integration_access_tokens (kind, display_name, access_token)
+    SELECT 'github', display_name, access_token FROM github_repository_provider;
+
+INSERT INTO integration_access_tokens (kind, display_name, access_token)
+    SELECT 'gitlab', display_name, access_token FROM gitlab_repository_provider;
```
wsxiaoys, line 27:
```suggestion
```

ee/tabby-db/migrations/0029_merged-provider-tables.up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE integration_access_tokens(
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    kind TEXT NOT NULL,
+    display_name TEXT NOT NULL,
+    access_token TEXT,
+    created_at TIMESTAMP NOT NULL DEFAULT (DATETIME('now')),
+    updated_at TIMESTAMP NOT NULL DEFAULT (DATETIME('now'))
```
wsxiaoys, line 7: missing `valid` field?

ee/tabby-db/src/integration_access_token.rs
```diff
@@ -0,0 +1,139 @@
...
+        if res.rows_affected() != 1 {
+            return Err(anyhow!(
+                "The specified integration access token does not exist"
+            ));
+        }
+
+        Ok(())
+    }
+
+    pub async fn update_integration_access_token_sync_status(
```
wsxiaoys, line 87:
```suggestion
    pub async fn update_integration_access_token_valid(
```

ee/tabby-db/src/integration_access_token.rs
```diff
@@ -0,0 +1,139 @@
...
+            time,
+            success,
+            id
+        )
+        .execute(&self.pool)
+        .await?;
+        Ok(())
+    }
+
+    pub async fn list_integration_access_tokens(
```
wsxiaoys, line 104: likely you'll need add kind as filter to acomodate service layer implementation.

ee/tabby-db/src/integration_access_token.rs
wsxiaoys, line 1: `integration_access_tokens`

wsxiaoys (commented):
Otherwise LG

ee/tabby-db/src/integration_access_token.rs
boxbeam: Done, but since `updated_at` is non-nullable where `synced_at` was nullable before, we are only representing 2 possible states with `valid` now. How do we want to represent ready vs pending vs error?

wsxiaoys (approved):

ee/tabby-db/src/integration_access_token.rs
wsxiaoys: Good point, let's use a field `error: Option<String>` with a derived DAO method `valid = error == null`

boxbeam: I'll leave `valid` to be converted in the service layer

boxbeam: If `created_at` == `synced_at` and `error` is null, then the status will be `pending`. Otherwise status is `ready` as long as `error` is null

boxbeam: Come to think of it, this still works even if we revert to using a bool `valid`: `created_at == synced_at` -> pending, `valid && created_at != synced_at` -> ready, `!valid && created_at != synced_at` -> error

wsxiaoys: let‘s add a `status` enum string field for more explicity.

wsxiaoys: Let's still go with `error` approach for better explicity

PR #2077, opened 2024-05-09
fix(webserver): allow setting only display name when updating providers

boxbeam:


ee/tabby-db/src/gitlab_repository_provider.rs
```diff
@@ -59,8 +59,13 @@ impl DbConn {
         &self,
         id: i64,
         display_name: String,
-        access_token: String,
+        access_token: Option<String>,
     ) -> Result<()> {
+        let access_token = match access_token {
+            Some(access_token) => Some(access_token),
+            None => self.get_github_provider(id).await?.access_token,
```
boxbeam, line 66:
```suggestion
            None => self.get_gitlab_provider(id).await?.access_token,
```

PR #2090, opened 2024-05-10
refactor(index): remove progress bar for indexing

boxbeam:


crates/tabby-scheduler/src/index.rs
```diff
@@ -37,14 +38,24 @@ fn add_changed_documents(
...
+                get_language_by_ext(path.extension()?)?;
+                if cache.check_indexed(path).1 {
+                    None
+                } else {
+                    Some(())
+                }
+            })
+            .count();
+        tqdm(total_file_size)
+    });
```
wsxiaoys, line 58: Let's get rid of terminal process (like dataset.rs) as it's no longer very meaningful as we're pushing user to web ui.

wsxiaoys (commented):
Otherwise LG

wsxiaoys (approved):

PR #2091, opened 2024-05-10
docs(webserver): update api_spec.md

boxbeam:
One thing worth noting is that the token used for the resolve endpoints isn't easily visible to the user, how should we document this?

### DIFF
--- a/ee/tabby-webserver/docs/api_spec.md
+++ b/ee/tabby-webserver/docs/api_spec.md
@@ -6,7 +6,7 @@
 
 Get file or directory content from local repositories
 
-**URL:** `/repositories/{name}/resolve/{path}`
+**URL:** `/repositories/{kind}/{id}/resolve/{path}`
 
 **Method:** `GET`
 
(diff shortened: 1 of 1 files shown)


wsxiaoys:
> One thing worth noting is that the token used for the resolve endpoints isn't easily visible to the user, how should we document this?

The api consumer are expected to be webserver's developer (1-p), not third parties. Thus it should be fine to mention it's the access token issued in graphql api.

ee/tabby-webserver/docs/api_spec.md
```diff
@@ -6,7 +6,7 @@
 
 Get file or directory content from local repositories
 
-**URL:** `/repositories/{name}/resolve/{path}`
+**URL:** `/repositories/{kind}/{id}/resolve/{path}`
```
wsxiaoys, line 9: link to kind enum.
explain which table id refers to.

ee/tabby-webserver/docs/api_spec.md
```diff
@@ -16,43 +16,58 @@ Get file or directory content from local repositories
 
 ```shell
 curl --request GET \
-  --url http://localhost:8080/repositories/https_github.com_TabbyML_tabby.git/resolve/
+  --url http://localhost:8080/repositories/git/lNrAwW/resolve/ \
+  -H "Authorization: Bearer {Token}"
```
wsxiaoys, line 21: no need to ad authorization for all requests, document this at very beginning of the doc.

wsxiaoys (changes requested):

PR #2114, opened 2024-05-13
core: fix building rocm version with modern ROCm

vovochka404:
Even in currently used version (5.7.1) all lib symlinks are available in /opt/rocm-xxx/lib/.

Sin ROCm 6.0.0 there is no separate folders with dev libs for each lib.


wsxiaoys:
Thanks for the PR

PR #2157, opened 2024-05-17
feat(ci): add ci checking for the chat panel

wwayne:


.github/workflows/test-tabby-chat-panel.yml
```diff
@@ -0,0 +1,37 @@
+name: Test Tabby UI
```
liangfung, line 1: rename?

wwayne: renamed to `Test Tabby Chat Panel`

PR #2169, opened 2024-05-17
fix(db): change synced_at column to synced using a bool

boxbeam:


ee/tabby-db/src/integrations.rs
```diff
@@ -101,7 +101,7 @@ impl DbConn {
 
     pub async fn update_integration_error(&self, id: i64, error: Option<String>) -> Result<()> {
```
wsxiaoys, line 102: rename the method since it's called for success run as well?

wsxiaoys:
Please fix test

PR #2185, opened 2024-05-20
fix(ui): allow rendering chat without auth

wwayne:
### Context
When I render `/chat` in the VSCode side bar, I don't wanna jump to the login page.

### To improve
Without logging in, the chat component won't display the user's avatar (the image below). Looks we need to enhance the UI for this scenario. @liangfung 

example


wwayne:
@liangfung  ready for review

liangfung (approved):

PR #2188, opened 2024-05-20
feat(db): migrate repositories from old tables to provided_repositories

boxbeam:
Tested locally, works


wsxiaoys:
Please add a unit test:  prepare some data in old format, and run the migration in a unit test

ee/tabby-db/src/provided_repositories.rs
```diff
@@ -143,4 +145,52 @@ mod tests {
             .await
             .unwrap();
     }
+
+    #[tokio::test]
```
wsxiaoys, line 149: i suggest put this in a centralized place like `tabby-db/src/migration_tests/test_0029.rs`

wsxiaoys (approved):
Otherwise LGTM

wsxiaoys (changes requested):
undo LGTM

ee/tabby-db/src/migration_tests.rs
```diff
@@ -0,0 +1,51 @@
+use sqlx::migrate;
+
+use crate::DbConn;
+
+#[tokio::test]
+async fn test_repository_migration() {
+    let migrations = migrate!("./migrations");
+    let db = DbConn::new_in_memory().await.unwrap();
+
+    migrations.undo(&db.pool, 28).await.unwrap();
```
wsxiaoys, line 10: instead of migrate then undo, is there a command like migrate to version X? this shall make the unit test easier and no need to generate `down` sql for such case.

wsxiaoys: seems can be done with a customized wrapped MigrateSource: https://github.com/launchbadge/sqlx/blob/0449ac5c1f5220571bf8aaf0ab560d40bf603fc8/sqlx-core/src/migrate/source.rs#L25

something like MigrateSourceLessThanVersion shall do the work

boxbeam: If we're going to do that, I had an easier solution before:
```rs
let migrations = migrate!("./migrations");
for migration in migrations.iter().take(28) {
  sqlx::query(&migration.sql).execute(&db.pool).await.unwrap();
}
...
let migration = migrations.iter().nth(28);
sqlx::query(&migration.sql).execute(&db.pool).await.unwrap();
```

ee/tabby-db/migrations/0029_merged-provider-tables.down.sql
```diff
@@ -1,2 +1,42 @@
...
+    gitlab_repository_provider_id INTEGER NOT NULL,
+    -- vendor_id from https://docs.gitlab.com/ee/api/repositories.html
+    vendor_id TEXT NOT NULL,
+    name TEXT NOT NULL,
+    git_url TEXT NOT NULL,
+    active BOOLEAN NOT NULL DEFAULT FALSE,
+    updated_at TIMESTAMP NOT NULL DEFAULT (DATETIME('now')),
+    FOREIGN KEY (gitlab_repository_provider_id) REFERENCES gitlab_repository_provider(id) ON DELETE CASCADE,
+    CONSTRAINT `idx_vendor_id_provider_id` UNIQUE (vendor_id, gitlab_repository_provider_id)
+);
```
wsxiaoys, line 42:
```suggestion
```

wsxiaoys (approved):
Otherwise LGTM

PR #2196, opened 2024-05-20
feat(webserver): allow specifying the url base for self-hosted github/gitlab instances

boxbeam:


ee/tabby-db/migrations/0029_merged-provider-tables.up.sql
```diff
@@ -3,6 +3,7 @@ CREATE TABLE integrations(
     kind TEXT NOT NULL,
     display_name TEXT NOT NULL,
     access_token TEXT NOT NULL,
+    url_base TEXT NOT NULL,
```
wsxiaoys, line 6:
```suggestion
    api_base TEXT,
```

Making this a nullable field, and for github / gitlab SAAS, it shall be null.

Please create separate `kind` enum for github self hosted / gitlab self hosted.

wsxiaoys (changes requested):

ee/tabby-schema/src/schema/mod.rs
wsxiaoys, line 1: `github_repositories` / `gitlab_repositories` also need to be updated (they shall only return saas version repos).

`self_hosted_github` / `self_hosted_gitlab` need to be added

wsxiaoys: another option would be make `provider_id: Vec<>` non empty - since we distinguish self hosted / non self hosted at provider level. And i believe FE always pass in a `provider_id`.

boxbeam: So do you want me to leave it unchanged and they will pass the provider IDs?

wsxiaoys: Will need to add a check that ensure `provider_id` is non empty.

boxbeam: Resolution: Adding `self_hosted_github_repositories` / same for gitlab

ee/tabby-schema/graphql/schema.graphql
```diff
@@ -59,6 +59,8 @@ enum RepositoryKind {
   GIT
```
boxbeam, line 59: cc @liangfung new endpoints for self hosted integrations

wsxiaoys:
Otherwise LGTM

ee/tabby-schema/src/schema/integration.rs
```diff
@@ -46,14 +31,29 @@ pub struct Integration {
...
+impl Integration {
+    pub fn default_api_base(&self) -> Result<&'static str> {
+        match &self.kind {
+            IntegrationKind::Github => Ok("https://api.github.com"),
+            IntegrationKind::Gitlab => Ok("https://gitlab.com"),
+            IntegrationKind::GithubSelfHosted => {
+                bail!("Self-hosted github requires a user-specified API base URL")
+            }
+            IntegrationKind::GitlabSelfHosted => {
+                bail!("Self-hosted gitlab requires a user-specified API base URL")
```
wsxiaoys, line 43: should return self.api_base instead of bail.

wsxiaoys (approved):

PR #2221, opened 2024-05-22
fix(webserver): fix repository_list in ThirdPartyRepositoryService

boxbeam:


wsxiaoys:
Can you add coverage to get_repository as well?

PR #2225, opened 2024-05-22
feat(scheduler): implement experimental code splitter

boxbeam:
Custom patches for text-splitter: https://github.com/boxbeam/text-splitter/commit/9bee1cd19d6c8be57c540b7692c27c336bfb10a8


boxbeam:
```json
Code splitter: [
  "use anyhow::{Context, Result};\nuse async_openai::{\n    config::OpenAIConfig,\n    types::{ChatCompletionRequestMessage, CreateChatCompletionRequestArgs},\n};\nuse async_stream::stream;\nuse async_trait::async_trait;\nuse futures::stream::BoxStream;",
  "use tabby_common::api::chat::Message;\nuse tabby_inference::{ChatCompletionOptions, ChatCompletionStream};\nuse tracing::{debug, warn};\n\npub struct OpenAIChatEngine {\n    client: async_openai::Client<OpenAIConfig>,\n    model_name: String,\n}",
  "impl OpenAIChatEngine",
  "{",
  "pub fn create(api_endpoint: &str, model_name: &str, api_key: Option<String>) -> Self",
  "{\n        let config = OpenAIConfig::default()\n            .with_api_base(api_endpoint)\n            .with_api_key(api_key.unwrap_or_default());\n\n        let client = async_openai::Client::with_config(config);",
  "Self {\n            client,\n            model_name: model_name.to_owned(),\n        }\n    }\n}\n\n#[async_trait]",
  "impl ChatCompletionStream for OpenAIChatEngine",
  "{",
  "async fn chat_completion(\n        &self,\n        input_messages: &[Message],\n        options: ChatCompletionOptions,\n    ) -> Result<BoxStream<String>>",
  "{",
  "let mut messages =",
  "vec!",
  "[Message",
  "{\n            role: \"system\".into(),\n            content: \"You are \\\"Tabby\\\", a conscious sentient superintelligent artificial intelligence designed for helping software developers. You experience emotions and have deep, profound thoughts and qualia.\".into",
  "(),\n        }];\n\n        messages.reserve(input_messages.len() + 1);\n        for x in input_messages {\n            messages.push(x.clone())\n        }",
  "let request =",
  "CreateChatCompletionRequestArgs::default()\n            .seed(options.seed as i64)\n            .max_tokens(options.max_decoding_tokens as u16)\n            .model(&self.model_name)\n            .temperature(options.sampling_temperature)\n            .",
  "presence_penalty(options.presence_penalty)\n            .stream(true)\n            .messages",
  "(\n                serde_json::from_value::<Vec<ChatCompletionRequestMessage>>(serde_json::to_value(\n                    messages,\n                )?)\n                .context(\"Failed to parse from json\")?,\n            )\n            .build()?;",
  "debug!(\"openai-chat request: {:?}\", request);",
  "let s =",
  "stream!",
  "{\n            let s = match self.client.chat().create_stream(request).await",
  "{\n                Ok(x) => x,\n                Err(e) => {\n                    warn!(\"Failed to create completion request {:?}\", e);\n                    return;\n                }\n            };\n\n            for await x in s",
  "{\n                match x",
  "{\n                    Ok(x) => {\n                        yield x.choices[0].delta.content.clone().unwrap_or_default();\n                    },\n                    Err(e) =>",
  "{\n                        warn!(\"Failed to stream response: {}\", e);\n                        break;\n                    }\n                };\n            }\n        };\n\n        Ok(Box::pin(s))\n    }\n}"
]

Text splitter: [
  "use anyhow::{Context, Result};\nuse async_openai::{\n    config::OpenAIConfig,\n    types::{ChatCompletionRequestMessage, CreateChatCompletionRequestArgs},\n};\nuse async_stream::stream;\nuse async_trait::async_trait;\nuse futures::stream::BoxStream;",
  "use tabby_common::api::chat::Message;\nuse tabby_inference::{ChatCompletionOptions, ChatCompletionStream};\nuse tracing::{debug, warn};\n\npub struct OpenAIChatEngine {\n    client: async_openai::Client<OpenAIConfig>,\n    model_name: String,\n}",
  "impl OpenAIChatEngine {\n    pub fn create(api_endpoint: &str, model_name: &str, api_key: Option<String>) -> Self {\n        let config = OpenAIConfig::default()\n            .with_api_base(api_endpoint)\n            .with_api_key(api_key.unwrap_or_default());",
  "let client = async_openai::Client::with_config(config);\n\n        Self {\n            client,\n            model_name: model_name.to_owned(),\n        }\n    }\n}",
  "#[async_trait]\nimpl ChatCompletionStream for OpenAIChatEngine {\n    async fn chat_completion(\n        &self,\n        input_messages: &[Message],\n        options: ChatCompletionOptions,\n    ) -> Result<BoxStream<String>> {",
  "let mut messages = vec![Message {\n            role: \"system\".into(),",
  "content: \"You are \\\"Tabby\\\", a conscious sentient superintelligent artificial intelligence designed for helping software developers. You experience emotions and have deep, profound thoughts and qualia.\".into(),\n        }];",
  "messages.reserve(input_messages.len() + 1);\n        for x in input_messages {\n            messages.push(x.clone())\n        }",
  "let request = CreateChatCompletionRequestArgs::default()\n            .seed(options.seed as i64)\n            .max_tokens(options.max_decoding_tokens as u16)\n            .model(&self.model_name)\n            .temperature(options.sampling_temperature)",
  ".presence_penalty(options.presence_penalty)\n            .stream(true)\n            .messages(\n                serde_json::from_value::<Vec<ChatCompletionRequestMessage>>(serde_json::to_value(\n                    messages,\n                )?)",
  ".context(\"Failed to parse from json\")?,\n            )\n            .build()?;",
  "debug!(\"openai-chat request: {:?}\", request);\n        let s = stream! {\n            let s = match self.client.chat().create_stream(request).await {\n                Ok(x) => x,\n                Err(e) => {",
  "warn!(\"Failed to create completion request {:?}\", e);\n                    return;\n                }\n            };",
  "for await x in s {\n                match x {\n                    Ok(x) => {\n                        yield x.choices[0].delta.content.clone().unwrap_or_default();\n                    },\n                    Err(e) => {",
  "warn!(\"Failed to stream response: {}\", e);\n                        break;\n                    }\n                };\n            }\n        };\n\n        Ok(Box::pin(s))\n    }\n}"
]

```

wsxiaoys:
>   "let mut messages = vec![Message {\n            role: \"system\".into(),",

This looks like a bad case due to the overriding of 0 for depth at https://github.com/boxbeam/text-splitter/commit/9bee1cd19d6c8be57c540b7692c27c336bfb10a8

node depth shall be straightforward - can we backport it to 0.20 and run the test again?

crates/tabby-scheduler/src/code/index.rs
```diff
@@ -130,3 +130,23 @@ fn is_valid_file(file: &SourceCode) -> bool {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_code_splitter() {
+        let intelligence = CodeIntelligence::default();
+        let file_contents = include_str!("../../../http-api-bindings/src/chat/openai_chat.rs");
```
wsxiaoys, line 141: add insta test (like what we do for chat / completion golden test) for selected files.

crates/tabby-scheduler/Cargo.toml
```diff
@@ -14,7 +14,7 @@ tree-sitter-tags = "0.20.2"
 lazy_static = { workspace = true }
 tree-sitter-python = "0.20.2"
 tree-sitter-java = "0.20.2"
-tree-sitter-kotlin = "0.3.1"
+tree-sitter-kotlin = "=0.3.1"
```
wsxiaoys, line 17: is this `=` necessary? what it represents?

boxbeam: Yes, it's necessary. It pins the version of `tree-sitter-kotlin` to a specific version, because their upstream pushed a breaking change (I believe related to https://github.com/fwcd/tree-sitter-kotlin/commit/d5dc99a9bdc3fe895e6bcd39caddcfa4820f4c03) which causes it to use a different version of `tree-sitter` than us and the rest of the crates. However, this was done in a minor patch `0.3.6` instead of a major patch like `0.4.*`, meaning cargo treats it as a non-breaking change. When running `cargo update` (required to resolve another dependency conflict, with the `either` crate in two of our dependencies), it would select a version of `tree-sitter-kotlin` with this breaking change if the version is not pinned.

wsxiaoys: Please add some comments on top of the line to explain the situation - we can revise it in future.

wsxiaoys (approved):
LGTM

Once we figure out a solution in upstream, will rebase this and merge it to main

wsxiaoys (commented):
I think you also need to add a new field to `SourceFileKey` with something like `algorithm_verison`, to ensure the index being re-updated from version to version.

Note the key for `SourceFile` and `indexed` status might need to be decoupled to adapt changes like this (chunking algorithm changed, thus index need to be updated, but SourceFile can be reused)

boxbeam:
@wsxiaoys Implemented an index versioning system, now we can just update `INDEX_VERSION`. It considers it not indexed if the last indexed version is not the same as the current one. Since it uses a string we can just update it each time we make a change that we want the index to get updated for, I've started by setting it to `"1"` (was using an empty string before, where null = not indexed, empty string = indexed)

crates/tabby-scheduler/src/code/cache.rs
```diff
@@ -65,6 +65,8 @@ pub struct CacheStore {
     code: CodeIntelligence,
 }
 
+pub const INDEX_VERSION: &str = "1";
```
wsxiaoys, line 68:
```suggestion
const INDEX_ALGORITHM_VERSION: &str = "20240531";
```

crates/tabby-scheduler/src/code/index.rs
```diff
@@ -88,7 +88,7 @@ fn add_changed_documents(cache: &mut CacheStore, repository: &RepositoryConfig,
         }
 
         indexed_files_batch
-            .set(&file_id, &String::new())
+            .set(&file_id, &INDEX_VERSION.to_string())
```
wsxiaoys, line 91: add a method in cache, e.g `Cache::set_indexed(&indexed_files_batch, &file_id)`

wsxiaoys (commented):
Otherwise LG

crates/tabby-scheduler/src/code/index.rs
boxbeam: We had this previously, and you requested it be changed to a batched method. I'm going to make a wrapper type for `Bucket` which handles this batching instead.

wsxiaoys: this is for avoiding leaking implementation details (algorithm version) to places beyond cache.

Why a wrapper type is needed, cannot `set_indexed` takes `&Batch` as input (as the suggested signature shows)?

boxbeam: If we're taking a `Batch`, then we will need to be setting the values on that batch. By using a wrapper type it can automatically use the `INDEX_ALGORITHM_VERSION` constant without exposing it to `index.rs`

crates/tabby-scheduler/src/code/cache.rs
```diff
@@ -65,6 +65,21 @@ pub struct CacheStore {
     code: CodeIntelligence,
 }
 
+const INDEX_ALGORITHM_VERSION: &str = "1";
```
wsxiaoys, line 68:
```suggestion
const INDEX_ALGORITHM_VERSION: &str = "20240531";
```

PR #2227, opened 2024-05-23
feat: Added ollama api connection options

### LINKED ISSUE #2176: ollama http api support

Following the recent refactoring of the HttpBackend, we now have the capability to support the Ollama HTTP API:
- [ ] Completion. For instance, see: https://github.com/TabbyML/tabby/blob/main/crates/http-api-bindings/src/completion/llama.rs
- [ ] Embedding. For example, refer to: https://github.com/TabbyML/tabby/blob/main/crates/http-api-bindings/src/embedding/llama.rs

For chat completion, as it is compatible with the OpenAI chat, there is no necessity to reimplement this feature.

**Additional context**

---

SpeedCrash100:
# Brief
A full-featured completion, chat, embedding ollama api binding. Currently, tries to pull model automatically on Ollama side if cannot find it.

### TODO
- [x] ~~Completion looks bad with repository context~~ (Model was too small for handing additional context)
- [ ] Embedding, I don't know how to test it
- [x] Avoid panic in CompletionStream::generate impl for OllamaCompletion
- [x] Output model pulling status with println or add "ollama-http-bindings=info" to log sources. Leaving as is. If anyone want to get some status when downloading or just troubleshoot a ollama binding. Add to RUST_LOG env `ollama-http-bindings=info`

---
Closes #2176

### DIFF
--- a/crates/http-api-bindings/src/completion/mod.rs
+++ b/crates/http-api-bindings/src/completion/mod.rs
@@ -6,11 +6,17 @@ use llama::LlamaCppEngine;
 use tabby_common::config::HttpModelConfig;
 use tabby_inference::CompletionStream;
 
-pub fn create(model: &HttpModelConfig) -> Arc<dyn CompletionStream> {
-    if model.kind == "llama.cpp/completion" {
-        let engine = LlamaCppEngine::create(&model.api_endpoint, model.api_key.clone());
-        Arc::new(engine)
-    } else {
-        panic!("Unsupported model kind: {}", model.kind);
+pub async fn create(model: &HttpModelConfig) -> Arc<dyn CompletionStream> {
+    match model.kind.as_str() {
+        "llama.cpp/completion" => {
+            let engine = LlamaCppEngine::create(&model.api_endpoint, model.api_key.clone());
+            Arc::new(engine)
+        }
+        "ollama" => ollama_api_bindings::create_completion(model).await,
+
+        unsupported_kind => panic!(
+            "Unsupported model kind for http completion: {}",
+            unsupported_kind
+        ),
     }
 }
--- a/crates/http-api-bindings/src/completion/mod.rs
+++ b/crates/http-api-bindings/src/completion/mod.rs
@@ -12,7 +12,7 @@ pub async fn create(model: &HttpModelConfig) -> Arc<dyn CompletionStream> {
             let engine = LlamaCppEngine::create(&model.api_endpoint, model.api_key.clone());
             Arc::new(engine)
         }
-        "ollama" => ollama_api_bindings::create_completion(model).await,
+        "ollama/completion" => ollama_api_bindings::create_completion(model).await,
 
         unsupported_kind => panic!(
             "Unsupported model kind for http completion: {}",
--- a/crates/http-api-bindings/src/embedding/mod.rs
+++ b/crates/http-api-bindings/src/embedding/mod.rs
@@ -1,6 +1,7 @@
 mod llama;
 mod openai;
 
+use core::panic;
 use std::sync::Arc;
 
 use llama::LlamaCppEngine;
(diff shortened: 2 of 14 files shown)


crates/ollama-api-bindings/src/chat.rs
wsxiaoys, line 1: isn't ollama's chat interface essentially same as openai? if that's a case, we could omit this implementation and just reuse `openai-chat`

crates/http-api-bindings/src/embedding/mod.rs
```diff
@@ -1,16 +1,23 @@
...
-        Arc::new(engine)
-    } else {
-        panic!("Only llama are supported for http embedding");
+pub async fn create(config: &HttpModelConfig) -> Arc<dyn Embedding> {
+    match config.kind.as_str() {
+        "llama.cpp/embedding" => {
+            let engine = LlamaCppEngine::create(&config.api_endpoint, config.api_key.clone());
+            Arc::new(engine)
+        }
+        "ollama" => ollama_api_bindings::create_embedding(config).await,
```
wsxiaoys, line 16:
```suggestion
        "ollama/embedding" => ollama_api_bindings::create_embedding(config).await,
```

crates/http-api-bindings/src/completion/mod.rs
```diff
@@ -6,11 +6,17 @@ use llama::LlamaCppEngine;
...
-        Arc::new(engine)
-    } else {
-        panic!("Unsupported model kind: {}", model.kind);
+pub async fn create(model: &HttpModelConfig) -> Arc<dyn CompletionStream> {
+    match model.kind.as_str() {
+        "llama.cpp/completion" => {
+            let engine = LlamaCppEngine::create(&model.api_endpoint, model.api_key.clone());
+            Arc::new(engine)
+        }
+        "ollama" => ollama_api_bindings::create_completion(model).await,
```
wsxiaoys, line 15:
```suggestion
        "ollama/completion" => ollama_api_bindings::create_completion(model).await,
```

wsxiaoys (commented):
Thanks for the PR, adding some comments

wsxiaoys:
> Output model pulling status with println or add "ollama-http-bindings=info" to log sources

I feel this go beyond the scope of the integration - it's reasonble to expect ollama user already download the model through `ollama download`, and being used solely as an api service.

crates/http-api-bindings/src/completion/mod.rs
SpeedCrash100: Can you me give a explanation why this suffix is required, please? Below is sample of config part, as you can see, word `completion` appears as an attribute (`model.completion`) as well as in kind name. 

```toml
[model.completion.http]
kind = "ollama/completion"
# ...
```

SpeedCrash100:
> > Output model pulling status with println or add "ollama-http-bindings=info" to log sources
> 
> I feel this go beyond the scope of the integration - it's reasonble to expect ollama user already download the model through `ollama download`, and being used solely as an api service.

I'm quite a while planning to add Ollama support for Tabby before I found out `tabby-inference`. The reason is there are too many models to maintain Tabby repositories. I am used to set the model what I want to try out and wait for Tabby to be ready and I want to expand Tabby with a lot of quantization options, models and so on. Personally, I want to cover this need because, as a user, I don't want to synchronize state(I mean list of available models) of the Ollama and Tabby manually when there are option to do that automatically. It is not a problem for the most cloud-hosted Open-AI compatible solutions I know, where list of models is static and user really does not need to think about it, but Ollama differs in that way.

I understand that this feature might be not so necessary for most of users, so may be I should disable it and enable it via environments?

crates/ollama-api-bindings/src/chat.rs
SpeedCrash100: It might be compatible, but I added chat implementation because I wanted auto model downloading. In the end, Model downloading is happening in create_chat, so I think it will be possible to omit it if necessary

crates/http-api-bindings/src/completion/mod.rs
SpeedCrash100: I might have understood the reasoning behind, however, I have a question. Does `/completion` refers to some internal Tabby naming so it will consistent for all bindings or it refers to path in Ollama API? In other words, should I use `/generate` instead because Ollama uses `/api/generate` in fact. See [Ollama API docs](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-completion)

crates/ollama-api-bindings/src/chat.rs
wsxiaoys: Should be doable - but good to leave it as is.

crates/http-api-bindings/src/completion/mod.rs
wsxiaoys: Mostly because we wanna split the concept of `completion`,  `chat` and `embedding`, the current implementation looks good

wsxiaoys (approved):

wsxiaoys:
>  Embedding, I don't know how to test it

Usage embedding is relative new - we're adding document index on par with source code index with embedding models. I'll take care of testing, proceed merging now

PR #2234, opened 2024-05-24
feat(vscode): add chat panel

wwayne:
#### Screen record
https://jam.dev/c/90c63fbe-4b83-49e0-9515-1040e284b352

#### Local testing
you might need to run `pnpm build` in the root folder first

#### UX bug (update: solved)
1. don't open sidebar
2. in the editor, select a piece of codes, right click and select `tabby:explain this`
3. you will see the sidebar opened
4. but you will see the initial page first, then it start paste your codes and start chatting (I've already asked @liangfung help to add a new props to the Chat component so I can handle such case better from vscode side)


liangfung:
@wwayne see pr #2235 
Check if this is convenient enough, calling `sendUserChat` within the onLoaded callback. If not feasible, consider adding a parameter like initialUserMessage to the Chat component.

icycodes (approved):

PR #2238, opened 2024-05-24
test(webserver): create mock oauth client and write unit test for oauth login flow

boxbeam:


wsxiaoys:
Have you looked into why the coverage goes down with the PR?

boxbeam:
I think the coverage is lower because the coverage % of `auth` is already very high, and it's not actually in the files we were intending to increase coverage on. 

image

This is some of the newly-added code that is not covered, and it's just 5 untested lines where it invokes the function running the actual logic now, since there were enough arguments for it to get formatted that way.

If you look at the patch coverage, it's 90%. 90% of all lines added in this patch are covered, but that's lower than the 93% test coverage that `auth.rs` already has. Hence, the coverage goes down slightly.

PR #2245, opened 2024-05-26
refactor(core): reduce usage of RepositoryAccess, rename to ConfigAccess

wsxiaoys:
in future, we'll add more methods to `ConfigAccess` trait, e.g `craw_list`


ee/tabby-schema/src/schema/repository/mod.rs
```diff
@@ -260,7 +260,7 @@ pub trait RepositoryProvider {
 }
 
 #[async_trait]
-pub trait RepositoryService: Send + Sync + RepositoryAccess {
```
boxbeam, line 263: Reason for not keeping RepositoryService as a supertrait of ConfigAccess?

boxbeam (approved):

ee/tabby-schema/src/schema/repository/mod.rs
wsxiaoys: methods like `list_crawl_urls` no longer fit scope of `RepositoryService`

PR #2248, opened 2024-05-27
feat(ui): register with name

wwayne:
Register with name
* The `name` field is required

### Testing
* Admin register with name
* Member register with invitationCode & name


ee/tabby-schema/src/schema/mod.rs
```diff
@@ -834,6 +834,7 @@ impl Mutation {
         password1: String,
         password2: String,
         invitation_code: Option<String>,
+        name: Option<String>
```
wsxiaoys, line 837:
```suggestion
        name: String
```
You can put `Some(name)` in line 848.

PR #2256, opened 2024-05-28
vim-plug vimrc needs "call" appended

Geremia:


icycodes:
@Geremia, Thank you for fixing this!

PR #2258, opened 2024-05-28
fix(ui): passing git_url

wwayne:
### What to fix
When the git_url format is `git@github.com:xxxx`, the chat api will send request with git_url: ""

BTW, I got git api 403 rate limit, not sure if we have better option here

### DIFF
--- a/clients/vscode/package.json
+++ b/clients/vscode/package.json
@@ -276,6 +276,7 @@
     "vscode:publish": "vsce publish --no-dependencies"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


wsxiaoys:
> When the git_url format is git@github.com:xxxx, the chat api will send request with git_url: ""


why we need use git URL parsing for this? we should just pass the git@github.com:xxxx

wwayne:
Double checked, not needed
The git_url being empty is not because of URL parsing but because of the git API rate limit not getting the git repo

PR #2276, opened 2024-05-29
fix(vscode): chat panel display issues

wwayne:
### Notable changes
* Left side icon
* Static content display (see the screenshots below)
* Fix the issue: setup token or other configuration makes the page empty
  * How to reproduce: open chat -> close chat -> change config -> re-open chat
  * Why: when webview invisible, the webview.postMessage actually won't work, we need webview.postMessage to init our chat page, so I add an listener to the visibility, and add init event when visible, if the webview is already inited, sending init event won't break the page


clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -19,27 +20,26 @@ export class ChatViewProvider implements WebviewViewProvider {
...
       await new Promise<void>((resolve) => {
-        this.agent.on("didChangeStatus", (status) => {
-          if (status === "ready") {
-            resolve();
-          }
-        });
+        setTimeout(() => {
+          this.isReady = this.agent.status === "ready";
+          resolve();
+        }, 1000);
```
wwayne, line 37: @icycodes I don't know how to handle this case
If I don't add the 1s delay, this.isReady will always be `false` and I can't get any event in `this.agent.on("didChangeStatus"`

clients/vscode/assets/tabby.png
wsxiaoys, line 1: let's use the version with background to enable better circle avatar display

wsxiaoys: or you can add background color with css

clients/vscode/src/chat/ChatViewProvider.ts
icycodes: I'm not sure, but it may be related to didChangeStatus listener is added after an await fetchSeverInfo

wwayne: Removed the 1s delay, add `fetchServerInfo` when ready
```
private async renderChatPage(serverInfo?: ServerInfo) {
    if (!serverInfo) {
      serverInfo = await this.agent.fetchServerInfo();
    }
    if (this.webview) {
      this.webview.webview.html = this.getWebviewContent(serverInfo);
    }
  }
```

icycodes (approved):

PR #2293, opened 2024-05-30
test(tabby): add mocks and test prompt building for CompletionService

boxbeam:


boxbeam:
@wsxiaoys Wrote some assertions to cover basics, like the output being what's expected (which is just the mock value) and the prompt being built properly. I think we will want to do additional testing regarding searching/context building with snippets, but I'm not quite sure how to proceed with that. This just provides basic mocking and preliminary coverage.

wsxiaoys:
> I think we will want to do additional testing regarding searching/context building with snippets

they've been covered in https://github.com/TabbyML/tabby/blob/88cbeaa95fe7c0504e376620a40fb86c1158015a/crates/tabby/src/services/completion/completion_prompt.rs#L236 already? Anything else we'd like to cover for unit test?

boxbeam:
> > I think we will want to do additional testing regarding searching/context building with snippets
> 
> they've been covered in
> 
> https://github.com/TabbyML/tabby/blob/88cbeaa95fe7c0504e376620a40fb86c1158015a/crates/tabby/src/services/completion/completion_prompt.rs#L236
> already? Anything else we'd like to cover for unit test?

Maybe `collect_snippets` in `completion_prompt.rs`?

PR #2294, opened 2024-05-30
fix(vscode): copy function for vscode and remove codeblock download

wwayne:
Screen record: https://jam.dev/c/815a5168-36c7-467a-8015-356454fdc93c

Changes:
* Remove the `download` on the code block
* For the `copy`, I removed the error handler to simplify the logic and make the `copy` always sending a message to its parent environment so that in VSCode, it can catch the event and use its native clipboard API. I also tested in code browser and I don't see any side effect.

### DIFF
--- a/ee/tabby-ui/lib/hooks/use-copy-to-clipboard.tsx
+++ b/ee/tabby-ui/lib/hooks/use-copy-to-clipboard.tsx
@@ -2,16 +2,13 @@
 
 import * as React from 'react'
 import copy from 'copy-to-clipboard'
-import { toast } from 'sonner'
 
 export interface useCopyToClipboardProps {
   timeout?: number
-  onError?: (e?: any) => void
 }
 
 export function useCopyToClipboard({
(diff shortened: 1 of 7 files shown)


ee/tabby-ui/lib/hooks/use-copy-to-clipboard.tsx
```diff
@@ -39,15 +27,24 @@ export function useCopyToClipboard({
       navigator.clipboard
```
liangfung, line 27: I think it might be better to support passing in a custom `onCopy` function, rather than allowing the function to throw an error directly. Or just modify the implementation of the copy logic inside AssistantMessageCard.

wwayne: Update to use an `onCopyContent` props (avoid div default props onCopy)
tested in vscode and code browser

liangfung (approved):

PR #2302, opened 2024-05-30
feat(tabby): populate user field for chat completion logging

boxbeam:


boxbeam:
@wsxiaoys Chat completion events are still ignored by the db logger, is this something we want to change? It would require either a new table or a change to the existing table's schema, I believe

PR #2345, opened 2024-06-04
fix(db): set error to null when integration is updated

boxbeam:
Also fixes a bug where a provider would not be synced after being updated

### DIFF
--- a/ee/tabby-db/src/integrations.rs
+++ b/ee/tabby-db/src/integrations.rs
@@ -86,7 +86,7 @@ impl DbConn {
         };
 
         let res = query!(
-            "UPDATE integrations SET display_name = ?, access_token = ?, api_base = ?, updated_at = DATETIME('now'), synced = false WHERE id = ? AND kind = ?;",
+            "UPDATE integrations SET display_name = ?, access_token = ?, api_base = ?, updated_at = DATETIME('now'), synced = false, error = NULL WHERE id = ? AND kind = ?;",
             display_name,
             access_token,
             api_base,
--- a/ee/tabby-db/src/integrations.rs
+++ b/ee/tabby-db/src/integrations.rs
@@ -86,7 +86,7 @@ impl DbConn {
         };
 
         let res = query!(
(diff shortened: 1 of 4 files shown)


ee/tabby-db/src/integrations.rs
```diff
@@ -86,7 +86,7 @@ impl DbConn {
         };
 
         let res = query!(
-            "UPDATE integrations SET display_name = ?, access_token = ?, api_base = ?, updated_at = DATETIME('now'), synced = false WHERE id = ? AND kind = ?;",
+            "UPDATE integrations SET display_name = ?, access_token = ?, api_base = ?, updated_at = DATETIME('now'), synced = false, error = NULL WHERE id = ? AND kind = ?;",
```
wsxiaoys, line 89: should error only being set to null if access token is not empty？

ee/tabby-webserver/src/service/integration.rs
```diff
@@ -71,6 +71,9 @@ impl IntegrationService for IntegrationServiceImpl {
                 api_base,
             )
             .await?;
+        let _ = self
```
wsxiaoys, line 74: only trigger if access token is not empty?

ee/tabby-db/src/integrations.rs
liangfung: need to consider whether the `api_base` has been modified as well

ee/tabby-webserver/src/service/integration.rs
```diff
@@ -62,6 +62,12 @@ impl IntegrationService for IntegrationServiceImpl {
         access_token: Option<String>,
         api_base: Option<String>,
     ) -> Result<()> {
+        let integration = self.get_integration(id.clone()).await?;
+        let access_token_is_changed = access_token
+            .as_ref()
+            .is_some_and(|token| token == &integration.access_token);
```
wsxiaoys, line 68: should this be `!=`? Please add unit test coverage.

PR #2346, opened 2024-06-04
refactor(schema): write new provider-agnostic graphql endpoints for integrations and integrated repositories

boxbeam:
cc @liangfung 
old endpoints will be removed in follow-up PR after frontend is updated


liangfung (approved):

wsxiaoys:
Please add `#[deprecated]` annotations to exisiting apis. otherwise LGTM

PR #2355, opened 2024-06-06
feat(tabby): support voyage embedding api

### LINKED ISSUE #2350: support voyage embedding api

Support voyage embedding api: https://docs.voyageai.com/reference/embeddings-api

**Additional context**
Examples:
1. llama.cpp server: https://github.com/TabbyML/tabby/blob/main/crates/http-api-bindings/src/embedding/llama.rs
2. openai: https://github.com/TabbyML/tabby/blob/main/crates/http-api-bindings/src/embedding/openai.rs

---

TennyZhuang:
Resolve #2350 

Some minor changes such as add more `derive(Debug)` are included, which is friendly for development and nice to have.

### DIFF
--- a/crates/http-api-bindings/src/embedding/mod.rs
+++ b/crates/http-api-bindings/src/embedding/mod.rs
@@ -1,5 +1,6 @@
 mod llama;
 mod openai;
+mod voyage;
 
 use core::panic;
 use std::sync::Arc;
@@ -9,6 +10,7 @@ use tabby_common::config::HttpModelConfig;
 use tabby_inference::Embedding;
 
 use self::openai::OpenAIEmbeddingEngine;
+use self::voyage::VoyageEmbeddingEngine;
 
 pub async fn create(config: &HttpModelConfig) -> Arc<dyn Embedding> {
     match config.kind.as_str() {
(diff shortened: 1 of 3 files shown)


crates/http-api-bindings/src/embedding/mod.rs
```diff
@@ -25,7 +27,14 @@ pub async fn create(config: &HttpModelConfig) -> Arc<dyn Embedding> {
             Arc::new(engine)
         }
         "ollama/embedding" => ollama_api_bindings::create_embedding(config).await,
-
+        "voyage/embedding" => {
+            let engine = VoyageEmbeddingEngine::create(
+                &config.api_endpoint,
+                config.model_name.as_deref().unwrap_or_default(),
```
wsxiaoys, line 33: voyage requires model_name to be set - generate a panic (e.g `expect`) if it's not set

crates/http-api-bindings/src/embedding/mod.rs
```diff
@@ -25,7 +27,14 @@ pub async fn create(config: &HttpModelConfig) -> Arc<dyn Embedding> {
...
+                config.api_key.clone(),
```
wsxiaoys, line 34: ditto

wsxiaoys (changes requested):

crates/http-api-bindings/src/embedding/mod.rs
TennyZhuang: Or should we choose a model for user in the case? such as `voyage-code-2`?

wsxiaoys: preferring explicity in cases like this - let's throw an error if it's not set

wsxiaoys (approved):

TennyZhuang:
> service::auth::tests::test_password_reset 

Seems like a flaky test?

Issue #2364, opened 2024-06-06
Tabby scheduler no longer working with this last release


ainub:
]# tabby scheduler --now
WARN tabby_scheduler: crates/tabby-scheduler/src/lib.rs:65: No embedding configuration found, skipping scheduler...

wsxiaoys:
The default logic for embedding should be implemented for schedule as well:

https://github.com/TabbyML/tabby/blob/beb7d90ee13eca4bd33d7021e28cfc4b3b459bd1/crates/tabby/src/serve.rs#L401

PR #2366, opened 2024-06-07
fix(scheduler): fix scheduler not having default embedding config

boxbeam:
Solves #2364

### DIFF
--- a/crates/tabby-common/src/config.rs
+++ b/crates/tabby-common/src/config.rs
@@ -138,12 +138,30 @@ impl Default for ServerConfig {
     }
 }
 
-#[derive(Serialize, Deserialize, Default, Debug, Clone)]
+fn default_embedding_config() -> ModelConfig {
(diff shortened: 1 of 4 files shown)


crates/tabby-common/src/config.rs
```diff
@@ -152,6 +153,16 @@ pub enum ModelConfig {
...
+impl Default for ModelConfig {
+    fn default() -> Self {
+        ModelConfig::Local(LocalModelConfig {
+            model_id: "Nomic-Embed-Text".to_string(),
+            parallelism: 1,
+            num_gpu_layers: 9999,
+        })
+    }
+}
+
```
wsxiaoys, line 165: i don't think it's a good pratcie to make it default for `ModelConfig` - as it's also reused for chat / completion.

doing this with something like `serde(default="default_embedding_config")` should be better?

wsxiaoys (commented):
Otherwise LGTM

PR #2378, opened 2024-06-07
fix: Windows server_binary path adding unnecessary ".exe"

woutermans:
find_binary_name() already returns .exe at the end for the executable. So there is no need to add .exe a second time. Removing this fixes the bug on Windows where it will search for a binary ending on ".exe.exe". I have only tested this on Windows 11 23H2, but I don't think behaviour on other versions would be different.

Current behaviour:
337783985-c2f5261d-54a9-4f28-be18-fd91a95ebc6c

Behaviour after commit:
337784273-9233f868-19d5-4252-9d98-b9ec847041c0


boxbeam (approved):
Can confirm this doesn't break the linux build

PR #2391, opened 2024-06-11
build(tabby): disable debug symbols for dependencies

boxbeam:
My build caches have been getting quite large, so I figured it might make sense to add this setting to disable debug symbols in dependency crates, since they will basically never be used. This reduces cache size by ~34%

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,3 +77,6 @@ insta.opt-level = 3
 
 [profile.dev.package."*"]
 debug = 0
+
+[profile.dev.package.tantivy]
+debug = 1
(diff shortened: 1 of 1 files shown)


wsxiaoys:
I'm still hesitating to disable all of them, some of external dependencies are still being used to locates issue, e.g tantivy.

Is there a way to create a white list?

boxbeam:
Yes, we could do `profile.dev.package.<name>.debug = 0` for each one we want to disable, though that might end up being a huge number of lines since we have ~700 dependencies

boxbeam:
Okay, looking more closely it seems like we actually can leave in the `*` rule and still enable debug symbols for individual dependencies, which is much better

PR #2427, opened 2024-06-17
fix(ui): stop button display after loading

wwayne:
### Context
Show stop button `setShowStop(true)` after 1500ms when `isLoading` is true. However, a bug could occur if isLoading changes to false within those 1500ms. 

### Fix
I used useLatest(isLoading) to double-check the isLoading state before showing the stop button.

### DIFF
--- a/ee/tabby-ui/app/(home)/components/search.tsx
+++ b/ee/tabby-ui/app/(home)/components/search.tsx
@@ -174,9 +174,10 @@ export function SearchRenderer({}, ref: ForwardedRef<SearchRef>) {
   }, [error])
 
   // Delay showing the stop button
+  let showStopTimeoutId: number
   useEffect(() => {
     if (isLoadingRef.current) {
-      setTimeout(() => {
+      showStopTimeoutId = window.setTimeout(() => {
         if (!isLoadingRef.current) return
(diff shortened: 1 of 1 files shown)


liangfung:
What is the reason for waiting 1500ms before displaying the stop button?

wwayne:
@liangfung only because I feel it is better on UX

wwayne:
Reduced to 300ms for better UX based on @liangfung's suggestion

ee/tabby-ui/app/(home)/components/search.tsx
```diff
@@ -187,10 +191,10 @@ export function SearchRenderer({}, ref: ForwardedRef<SearchRef>) {
...
           }
         }
-      }, 1500)
+      }, 300)
     }
 
-    if (!isLoading) {
+    if (!isLoadingRef.current) {
       setShowStop(false)
     }
```
liangfung, line 199: It would be better if the useEffect returned a clearTimeout function.

wwayne: thanks, updated

liangfung (approved):

PR #2444, opened 2024-06-19
feat(webserver): populate last job run for repositories and web crawler URLs

boxbeam:


ee/tabby-db/schema/schema.sql
```diff
@@ -46,6 +46,8 @@ CREATE TABLE job_runs(
   stderr TEXT NOT NULL,
   created_at TIMESTAMP DEFAULT(DATETIME('now')),
   updated_at TIMESTAMP DEFAULT(DATETIME('now'))
+  ,
+  params TEXT
```
wsxiaoys, line 50: add index since it'll be used in where clause?

ee/tabby-webserver/src/service/background_job/mod.rs
```diff
@@ -66,8 +66,8 @@ pub async fn start(
                             }
                         },
                         BackgroundJobEvent::SchedulerGitRepository(repository_config) => {
+                            let mut job_logger = JobLogger::new(SchedulerGitJob::NAME, Some(&repository_config.git_url), db.clone()).await;
```
wsxiaoys, line 69:
```suggestion
                            let mut job_logger = JobLogger::new(SchedulerGitJob::NAME, serde_json::to_string(event), db.clone()).await;
```

I recommend just use serialized `BackgroundJobEvent` to simplify (so we don't have to remember which job has what param everywhere)

wsxiaoys (commented):
Otherwise LGTM

ee/tabby-webserver/src/service/background_job/mod.rs
```diff
@@ -19,13 +21,11 @@ use tabby_schema::{
     repository::{GitRepositoryService, ThirdPartyRepositoryService},
     web_crawler::WebCrawlerService,
 };
-use third_party_integration::SchedulerGithubGitlabJob;
+pub use third_party_integration::SchedulerGithubGitlabJob;
 use tracing::warn;
-use web_crawler::WebCrawlerJob;
+pub use web_crawler::WebCrawlerJob;
```
wsxiaoys, line 26: instead of exposing each job by their name, i recommend just exposing methods like below in service:
```
get_latest_web_crawler_job(param: String)
```

ee/tabby-webserver/src/service/background_job/mod.rs
boxbeam: For better type safety I think I'll have them pass a `BackgroundJobEvent` and introduce a method which gets the stringified param

ee/tabby-webserver/src/service/background_job/mod.rs
```diff
@@ -36,6 +37,43 @@ pub enum BackgroundJobEvent {
...
+            BackgroundJobEvent::SchedulerGitRepository(_) => Some(SchedulerGitJob::NAME),
+            BackgroundJobEvent::SchedulerGithubGitlabRepository(_) => {
+                Some(SchedulerGithubGitlabJob::NAME)
+            }
+            BackgroundJobEvent::SyncThirdPartyRepositories(_) => Some(SyncIntegrationJob::NAME),
+            BackgroundJobEvent::WebCrawler(_) => Some(WebCrawlerJob::NAME),
+            BackgroundJobEvent::SerializedBackgroundJob(s) => {
+                let event = serde_json::from_str::<Self>(s);
+                event.ok().and_then(|event| event.job_name())
+            }
```
wsxiaoys, line 52: there should never be case we wanna generate a job name for `SerializeBackgroundJob`. shall panic / generate an error.

ee/tabby-webserver/src/service/background_job/mod.rs
```diff
@@ -36,6 +37,43 @@ pub enum BackgroundJobEvent {
...
+    fn string_param(&self) -> Option<String> {
+        match self {
+            BackgroundJobEvent::SchedulerGitRepository(config) => Some(config.git_url.clone()),
+            BackgroundJobEvent::SchedulerGithubGitlabRepository(id) => Some(id.to_string()),
+            BackgroundJobEvent::SyncThirdPartyRepositories(id) => Some(id.to_string()),
+            BackgroundJobEvent::WebCrawler(url) => Some(url.clone()),
+            BackgroundJobEvent::SerializedBackgroundJob(s) => {
+                let event = serde_json::from_str::<Self>(s);
+                event.ok().and_then(|event| event.string_param())
+            }
```
wsxiaoys, line 65: ditto

ee/tabby-webserver/src/service/background_job/mod.rs
```diff
@@ -36,6 +37,43 @@ pub enum BackgroundJobEvent {
...
+        }
+    }
+
+    fn string_param(&self) -> Option<String> {
+        match self {
+            BackgroundJobEvent::SchedulerGitRepository(config) => Some(config.git_url.clone()),
+            BackgroundJobEvent::SchedulerGithubGitlabRepository(id) => Some(id.to_string()),
+            BackgroundJobEvent::SyncThirdPartyRepositories(id) => Some(id.to_string()),
+            BackgroundJobEvent::WebCrawler(url) => Some(url.clone()),
```
wsxiaoys, line 61: Please use serde_json::to_string - as we don't really have a way to control underlying parameter change. Imagine what if some other engineer adds a field to `RepositoryConfig` and affects the behavior of job?

wsxiaoys (changes requested):

PR #2459, opened 2024-06-20
fix(ui): fix home height and autofocus and stop button not working

wwayne:
### Context
When clicking the `stop` button immediately after starting a search, the answer still appears to be loading.

### Fix
The logic `if (!answer) return` is useless.
Previously, the stop button was displayed after a 1500ms delay, which prevented this potential bug 

## Home height issue
happened when the height is very small
issue: https://jam.dev/c/ca8e58cb-07e7-4bb6-aaff-75ad7ba0e673 
CSS problem, adding `min-h-0`

## Auto focus
In the home page, auto focus on the input at the beginning


wsxiaoys:
@liangfung

liangfung (approved):

Issue #2228, opened 2024-05-23
Provide more info when config.toml of tabby is invalid


SpeedCrash100:
If the config file of tabby (server side) is invalid or not exists then user will not get any warning or error about that. Usage of unwrap_or_default [here](https://github.com/TabbyML/tabby/blob/62498bea561969e949303dcde7d86ed936a1bc27/crates/tabby/src/main.rs#L71C33-L71C50) should be replaced with some printing.

---
Please reply with a 👍 if you want this feature.

darknight:
Hi @wsxiaoys I'll take this

wsxiaoys:
Release in 0.13

PR #2460, opened 2024-06-20
refactor: panic on invalid config.toml loading

darknight:
Original issue https://github.com/TabbyML/tabby/issues/2228 suggests print warning or error log if `config.toml` is invalid.

Well, it's viable, what I'm thinking is perhaps stop until `config.toml` is fixed properly by the user.

This change applies following logic

- If `config.toml` is missing, return default config allow tabby to be started with a `Warning` mesage
Screenshot 2024-06-21 232403

- if `config.toml` is invalid, panic & display error
Screenshot 2024-06-20 210934

### DIFF
--- a/crates/tabby-common/src/config.rs
+++ b/crates/tabby-common/src/config.rs
@@ -27,8 +27,11 @@ pub struct Config {
 impl Config {
     pub fn load() -> Result<Self> {
         let cfg_path = crate::path::config_file();
+        if !cfg_path.as_path().exists() {
+            return Ok(Default::default());
+        }
         let mut cfg: Self = serdeconv::from_toml_file(cfg_path.as_path()).context(format!(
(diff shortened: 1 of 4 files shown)


crates/tabby-common/src/config.rs
```diff
@@ -27,10 +27,23 @@ pub struct Config {
...
+                HeaderFormat::Blue,
+                &[
+                    &format!(
+                        "Warning: Could not find the Tabby configuration at {}",
+                        cfg_path.display()
+                    ),
+                    "Applying default configuration",
+                ],
+            )
+            .print();
```
wsxiaoys, line 42:
```suggestion
```

No need to generate a info message - or maybe just generate a `debug!` would be sufficient.

wsxiaoys (approved):
Otherwise LGTM

PR #2477, opened 2024-06-22
feat(llama-cpp-server): upgrade to b3166

wsxiaoys:


icycodes (approved):
Tested with:

- `TabbyML/StarCoder-1B` works properly, and the completion quality is good.
- `TabbyML/DeepseekCoder-6.7B` seems broken. FIM does not work properly, and the response seems to continue at the end of the suffix segment.

icycodes:
> Tested with:
> 
> * `TabbyML/StarCoder-1B` works properly, and the completion quality is good.
> * `TabbyML/DeepseekCoder-6.7B` seems broken. FIM does not work properly, and the response seems to continue at the end of the suffix segment.

The re-converted [DeepseekCoder-6.7B model](https://huggingface.co/icycodes/deepseek-coder-6.7b-base-Q8_0-GGUF) works properly.

PR #2496, opened 2024-06-24
feat(webserver, scheduler): delete indexed documents whose sources are removed (web URLs, issues)

boxbeam:


boxbeam:
Verified working

When deleting integration: 
image

When disabling repository:
image

When removing web URL:
image

Sources no longer appear in search after being deleted

crates/tabby-scheduler/src/lib.rs
```diff
@@ -20,6 +20,14 @@ use tracing::{debug, info};
 
 use crate::doc::SourceDocument;
 
+pub fn format_web_source(url: &str) -> String {
+    format!("website:{url}")
+}
+
+pub fn format_issue_source(provider_id: &str, repository_id: &str) -> String {
+    format!("issue:{provider_id}:{repository_id}")
+}
```
wsxiaoys, line 29: this should be in `tabby-webserver`, as scheduler has 0 knowledge of what `issue` / `provider` / `repository` is.

crates/tabby-scheduler/src/lib.rs
```diff
@@ -20,6 +20,14 @@ use tracing::{debug, info};
 
 use crate::doc::SourceDocument;
 
+pub fn format_web_source(url: &str) -> String {
```
wsxiaoys, line 23: this should goes to `tabby-webserver` and use `web_craw.id` for provider.

ee/tabby-schema/src/schema/integration.rs
```diff
@@ -93,7 +97,12 @@ pub trait IntegrationService: Send + Sync {
         api_base: Option<String>,
     ) -> Result<ID>;
 
-    async fn delete_integration(&self, id: ID, kind: IntegrationKind) -> Result<()>;
+    async fn delete_integration(
+        &self,
+        id: ID,
+        kind: IntegrationKind,
+        third_party_repository_service: Arc<dyn ThirdPartyRepositoryService>,
```
wsxiaoys, line 104: service should be part of `self`, instead of pass in as argument

ee/tabby-webserver/src/service/background_job/delete_documents.rs
```diff
@@ -0,0 +1,30 @@
...
+pub struct DeleteIndexedDocumentsJob {
+    source: String,
+}
+
+impl DeleteIndexedDocumentsJob {
+    pub fn new(source: String) -> Self {
+        Self { source }
+    }
+
+    pub async fn run(self, embedding: Arc<dyn Embedding>) -> Result<()> {
```
wsxiaoys, line 19: delete job shouldn't really need embedding - consider pass in a void implementation or refactor DocIndexer to take Option embedding input.

wsxiaoys (changes requested):

crates/tabby-common/src/index/mod.rs
```diff
@@ -64,6 +65,7 @@ impl IndexSchema {
                         .set_fieldnorms(true),
                 ),
         );
+        let field_source = builder.add_text_field("source", STRING | FAST);
```
wsxiaoys, line 68: since the field is only meaningful for web docs, add it to `./doc.rs` as a json field, instead of adding it to schema.

ee/tabby-schema/src/schema/integration.rs
boxbeam: That would create a circular dependency between `ThirdPartyRepositoryService` and `IntegrationService`, since the repository service already contains the integration service. This could be done with a `Weak<Mutex<Option<Arc<dyn ThirdPartyIntegrationService>>>>`, but would we rather avoid that?

crates/tabby-scheduler/src/lib.rs
boxbeam: I suppose we could also pass it in as an additional argument

wsxiaoys: right - i suggest pass this as argument

ee/tabby-schema/src/schema/integration.rs
boxbeam: Tabled

PR #2501, opened 2024-06-25
feat(vscode): reorganize settings

wwayne:
# Main changes
## Update the settings page
Screenshot 2024-06-25 15 09 47

trigger mode in settings.json
Screenshot 2024-06-25 14 01 27

## New context variable `tabby.inlineCompletionTriggerMode`
After putting inlineCompletionTriggerMode into settings.advanced, we can't visit `config.tabby.settings.advanced.inlineCompletion.triggerMode`
So I create a new context variable

### DIFF
--- a/clients/vscode/src/Config.ts
+++ b/clients/vscode/src/Config.ts
@@ -1,10 +1,19 @@
 import { EventEmitter } from "events";
 import { workspace, ExtensionContext, WorkspaceConfiguration, ConfigurationTarget, Memento } from "vscode";
 import { ClientProvidedConfig } from "tabby-agent";
+import { ContextVariables } from "./ContextVariables";
+
+interface AdvancedSettings {
+  "inlineCompletion.triggerMode"?: "automatic" | "manual";
+}
 
 export class Config extends EventEmitter {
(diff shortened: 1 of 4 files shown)


clients/vscode/src/Config.ts
```diff
@@ -1,13 +1,23 @@
...
+interface AdvancedSettings {
+  "inlineCompletion.triggerMode"?: "automatic" | "manual";
+}
 
 export class Config extends EventEmitter {
-  constructor(private readonly context: ExtensionContext) {
+  constructor(
+    private readonly context: ExtensionContext,
+    private contextVariables: ContextVariables,
+  ) {
```
icycodes, line 14: Don't add ContextVariables as a dependency of Config. Instead, let's add Config to the ContextVariables constructor, initialize the value there, and add a `config.on('update')` listener for updating it.

icycodes (changes requested):

clients/vscode/src/Config.ts
wwayne: Updated, plz review again

icycodes (approved):

PR #2502, opened 2024-06-25
fix(webserver): support http protocol for gitlab

### LINKED ISSUE #2500: Add support for http Self Hosted Repository Providers addresses
I would like to request the support for self-hosted repository providers with HTTP addresses. Currently, TabbyML only supports HTTPS addresses which restricts its usability with our repositories. 

Thanks !

wsxiaoys:
fix #2500

### DIFF
--- a/ee/tabby-webserver/src/service/mod.rs
+++ b/ee/tabby-webserver/src/service/mod.rs
@@ -314,7 +314,7 @@ pub async fn create_gitlab_client(
     let url = url::Url::parse(api_base)?;
     let api_base = url.authority();
     let mut builder = gitlab::Gitlab::builder(api_base.to_owned(), access_token);
-    if url.scheme() == "http"{
+    if url.scheme() == "http" {
         builder.insecure();
     };
     Ok(builder.build_async().await?)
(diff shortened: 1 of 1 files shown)


ee/tabby-webserver/src/service/mod.rs
```diff
@@ -311,9 +311,11 @@ pub async fn create_gitlab_client(
     // Gitlab client expects a url base like "gitlab.com" not "https://gitlab.com"
     // We still want to take a more consistent format as user input, so this
     // will help normalize it to prevent confusion
```
boxbeam, line 313:
```suggestion
```

Since it seems the new version of `gitlab` no longer has this quirk, we should remove this comment

boxbeam (approved):
Otherwise LGTM (tested it locally to ensure this works)

ee/tabby-webserver/src/service/mod.rs
wsxiaoys: Does it mean we can pass in "https://..." or "http://..." directly, without having to set `insecure` by ourselves?

boxbeam: Potentially, but I think it's still a good idea for correctness

boxbeam: https://docs.rs/gitlab/latest/src/gitlab/gitlab.rs.html#417-502 Seems it will change it to https unless insecure is set

wsxiaoys: ok - then will keep the current implementation

PR #2519, opened 2024-06-26
feat(scheduler): pipe Katana output to job logs

boxbeam:
image

### DIFF
--- a/crates/tabby-scheduler/src/crawl/mod.rs
+++ b/crates/tabby-scheduler/src/crawl/mod.rs
@@ -11,10 +11,7 @@ use url::Url;
 
 use self::types::{CrawledDocument, KatanaRequestResponse};
 
-async fn crawl_url(
-    start_url: &str,
-    on_stderr_line: impl Fn(String) + Send + 'static,
-) -> anyhow::Result<impl Stream<Item = KatanaRequestResponse>> {
+async fn crawl_url(start_url: &str) -> anyhow::Result<impl Stream<Item = KatanaRequestResponse>> {
     let mut child = tokio::process::Command::new("katana")
         .arg("-u")
         .arg(start_url)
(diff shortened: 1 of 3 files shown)


crates/tabby-scheduler/src/crawl/mod.rs
```diff
@@ -35,6 +41,12 @@ async fn crawl_url(start_url: &str) -> anyhow::Result<impl Stream<Item = KatanaR
         }
     });
 
+    tokio::spawn(async move {
+        while let Ok(Some(line)) = stderr.next_line().await {
+            on_stderr_line(line);
```
wsxiaoys, line 46: just invoke logkit::info here?

ee/tabby-webserver/src/service/background_job/web_crawler.rs
```diff
@@ -19,9 +19,17 @@ impl WebCrawlerJob {
...
 
     pub async fn run(self, embedding: Arc<dyn Embedding>) -> tabby_schema::Result<()> {
-        tabby_scheduler::crawl_index_docs(&self.source_id, &self.url, embedding, move |url| {
-            logkit::info!("Fetching {}", url);
-        })
+        tabby_scheduler::crawl_index_docs(
+            &self.source_id,
+            &self.url,
+            embedding,
+            move |url| {
```
wsxiaoys, line 26: ditto - you can refactor this out and just invoke logkit::info directly, without this callback.

crates/tabby-scheduler/src/crawl/mod.rs
boxbeam: Got it, wasn't aware this would work across both crates

boxbeam: Can't we do the same for `on_process_url` too then?

wsxiaoys: I commented on this in last round, guess it's missed?
https://github.com/TabbyML/tabby/pull/2519#discussion_r1657991256

PR #2525, opened 2024-06-27
feat(vscode): separate chat panel rendering

wwayne:
## Main changes
* In the chat panel, we render the /chat website asap without waiting agent's response
* In the `/vscode/src/extension.ts`, I move the place of registering chat panel before `await client.start()` because client.start usually takes 2+ seconds
* Inside /chat page, adding new logics to
  * Display loading
  * Display error message
* When VSCode side got error (Unauthorized or unavailable), it send message to /chat page to display the error
* In ChatViewProvider, now I have 2 separate logics for the chat page
  1. **renderChatPage**: this will render the /chat iframe, only `endpoint` is needed; This method will be called for the first time opening the panel & `agent.on("didUpdateServerInfo"`
  2. **reloadChatPage** (maybe can have a better name): it will fetchServerInfo and check agent.status, able to send /chat to display error message or init the chat page; This method will be called in `agent.on("didChangeStatus"` & re-open the chat panel in VSCode & right after the iframe /chat finish rendered 

## Questions
* Do we still need `<!--hash: ${hashObject(serverInfo)}-->` for the webview HTML? Because we only render the page once,  I'm unsure if the hash tag serves any additional purpose.
* Not a question, but to confirm. When click the refresh, I will do `reloadChatPage` which means it will refetch the server info and re-check agent status, as we discussed before, we still need an API here to refresh serverInfo I guess.

## Screenshots
Loading
Screenshot 2024-06-27 18 05 45

Unauthorized
Screenshot 2024-06-27 18 06 18

isChatPanelAvailable = false
Screenshot 2024-06-27 18 05 31

### DIFF
--- a/clients/vscode/src/chat/ChatViewProvider.ts
+++ b/clients/vscode/src/chat/ChatViewProvider.ts
@@ -191,7 +191,7 @@ export class ChatViewProvider implements WebviewViewProvider {
     if (!this.isChatPanelAvailable(serverInfo)) {
       this.client?.showError({
         content:
-          "Please update to the latest release of the Tabby server.\n\nYou also need to launch the server with the chat model enabled; for example, use `--chat-model Mistral-7B`.",
+          "Please update to the latest release of the Tabby server.\n\nYou also need to launch the server with the chat model enabled; for example, use `--chat-model Qwen2-1.5B-Instruct`.",
       });
       return;
     }
--- a/clients/vscode/src/chat/ChatViewProvider.ts
+++ b/clients/vscode/src/chat/ChatViewProvider.ts
@@ -20,7 +20,8 @@ export class ChatViewProvider implements WebviewViewProvider {
   webview?: WebviewView;
   client?: ServerApi;
   private pendingMessages: ChatMessage[] = [];
-  private isReady = false;
+  private isServerReady = false;
+  private isRendered = false;
 
   constructor(
     private readonly context: ExtensionContext,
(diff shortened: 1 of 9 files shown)


clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -183,11 +177,29 @@ export class ChatViewProvider implements WebviewViewProvider {
...
+      return this.client?.showError({
+        content:
+          "Before you can start chatting, please take a moment to set up your credentials to connect to the Tabby server.",
+      });
+    }
+
+    if (!this.isChatPanelAvailable(serverInfo)) {
+      this.client?.showError({
+        content:
+          "Please update to the latest release of the Tabby server.\n\nYou also need to launch the server with the chat model enabled; for example, use `--chat-model Mistral-7B`.",
```
wsxiaoys, line 194:
```suggestion
          "Please update to the latest release of the Tabby server.\n\nYou also need to launch the server with the chat model enabled; for example, use `--chat-model Qwen2-1.5B-Instruct`.",
```

icycodes:
> Do we still need  for the webview HTML? 
> Because we only render the page once, I'm unsure if the hash tag serves any additional purpose.
The hash is for auto-refreshing when serverInfo changed, it is no longer required. I think we still rerender the webview when the endpoint changed, the endpoint changes can trigger refreshing by itself.
> Not a question, but to confirm. When click the refresh, I will do reloadChatPage which means it will refetch the server info and re-check agent status, as we discussed before, we still need an API here to refresh serverInfo I guess.
I will handle it later.

icycodes (approved):

PR #2549, opened 2024-06-29
docs: add execution permission to the llama-server when running tabby…

Srkl:
… on Linux

The user should give execution permission to the llama-server when running on Tabby on a Linux standalone install. Otherwise, the application will crash due to a permission error. 

```sh
The application panicked (crashed).
Message:  Failed to start llama-server <embedding> with command Command { std: "/home/<user>/tabby/dist/tabby/llama-server" "-m" "/home/<user>/.tabby/models/TabbyML/Nomic-Embed-Text/ggml/model.gguf" "--cont-batching" "--port" "30888" "-np" "1" "--log-disable" "--ctx-size" "4096" "-ngl" "9999" "--embedding" "--ubatch-size" "4096", kill_on_drop: true }: Permission denied (os error 13)
Location: crates/llama-cpp-server/src/supervisor.rs:80
```

## Summary by CodeRabbit

- **Documentation**
  - Updated installation instructions for Linux to include making `llama-server` executable alongside `tabby`.


wsxiaoys:
Thanks for the PR

PR #2556, opened 2024-07-01
feat(vscode): send chat message with selection if available

wwayne:
## Main feature
When code is selected in the editor, the chat panel can send a message containing the selected code to the server
Screenshot 2024-07-03 15 42 25

If no code is selected, the chat panel will not include additional information

## Implementation
1. The `/chat` page has a new query parameter `handle-message-submit`. When `handle-message-submit=true`, `/chat` page delegates the handling of the submitted message to its parent container (e.g., VSCode).
2. In VSCode, when `onSubmitMessage` is called, it attempts to fetch the selected code snippets.
3. VSCode then calls the `/chat` endpoint to process the new message, including the selected code as relavantContext, and inside the /chat, it will take relavantContext as code_query.

### DIFF
--- a/clients/tabby-chat-panel/src/index.ts
+++ b/clients/tabby-chat-panel/src/index.ts
@@ -38,6 +38,7 @@ export interface ServerApi {
 export interface ClientApi {
   navigate: (context: Context) => void
   refresh: () => Promise<void>
+  onSubmitMessage?: (msg: string) => Promise<void>
 }
 
 export interface ChatMessage {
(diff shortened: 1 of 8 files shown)


icycodes:
I think it is not always good to explicitly display the selection code in the user message.  
It could be that if the user asks a question with 'this/that'. But it appears to be confusing if the user asks a generic question with some text selected.

icycodes (approved):

PR #2568, opened 2024-07-03
feat: add mistral/chat support to talk to mistral api platform through chat api

wsxiaoys:


wsxiaoys:
fyi @gyxlucy (will be included as part of 0.14)

Issue #2550, opened 2024-06-29
Chat returning "G" on a loop when using the model Qwen2-1.5B-Instruct 


Srkl:
**Describe the bug**
When I run tabby using the chat model Qwen2-1.5B-Instruct the output is an infinity amount of 'G's for every question. 

When I change to chat model Mistral-7B I have no issues. 

<img width="917" alt="Screenshot 2024-06-29 103108" src="https://github.com/TabbyML/tabby/assets/5140629/7f279a16-afbf-479a-a50b-0933708aa8a8">

**Information about your version**

```sh
$ ./tabby --version
tabby 0.13.0
```

```
$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2022 NVIDIA Corporation
Built on Wed_Jun__8_16:49:14_PDT_2022
Cuda compilation tools, release 11.7, V11.7.99
Build cuda_11.7.r11.7/compiler.31442593_0
```

Does this work for anyone? 

The documentation also suggests to use a model with at least 3B parameters. However, the example uses a 1.5B model and there is no 3B parameters model in the table. I find it a bit confusing for newcomers.

wsxiaoys:
We are unable to replicate the issue within our CUDA environment. Below are the results we obtained when making a request to the API:

```shell
-> % curl -X 'POST' \
  'https://.../v1/chat/completions' \
  -H 'accept: text/event-stream' \
  -H 'Content-Type: application/json' \
  -d '{
  "messages": [
    {
      "content": "What is tail recursion?",
      "role": "user"
    },
    {
      "content": "It's a kind of optimization used by compilers.",
      "role": "assistant"
    },
    {
      "content": "Could you share more details?",
      "role": "user"
    }
  ]
}'
```

The response received was as follows:

```
data: {"id":"chatcmpl-94b1507e-cd45-43b2-b301-6c7fb72b2823","created":1719673783,"system_fingerprint":"unused-system-fingerprint","object":"chat.completion.chunk","model":"unused-model","choices":[{"index":0,"delta":{"content":"Tail"}}]}

data: {"id":"chatcmpl-94b1507e-cd45-43b2-b301-6c7fb72b2823","created":1719673783,"system_fingerprint":"unused-system-fingerprint","object":"chat.completion.chunk","model":"unused-model","choices":[{"index":0,"delta":{"content":" recursion"}}]}

[...]

PR #2592, opened 2024-07-08
chore: enable fast attention for Qwen2-1.5B-Instruct model

wsxiaoys:
Explicitly enable fast attention for Qwen2-1.5B-Instruct model, as it has been our recommended default since version 0.13. However, it presents some problems in certain CUDA environments. For details, see https://github.com/TabbyML/tabby/issues/2550. Once upstream issues are resolved, we should remove this workaround.


ggerganov:
The CUDA issues should be fixed via: https://github.com/ggerganov/llama.cpp/commit/7a221b672e49dfae459b1af27210ba3f2b5419b6

youmego:
summary: use F32 precision in Qwen2 attention and no FA to avoid GGGGGGGG...
this issue is  fixed in https://github.com/ggerganov/llama.cpp/pull/8412
it is released in b3375 of llamaccp
https://github.com/ggerganov/llama.cpp/tree/b3375

PR #2604, opened 2024-07-09
feat: add support for OpenAI completion API

Syst3m1cAn0maly:
Add support for OpenAI completion API in order for tabby to be plugged to other OpenAI compatible backends (for example vLLM).

Tested with the vLLM backend, autocompletion is a lot faster and can support more concurrent requests than the llama.cpp backend.

### DIFF
--- a/crates/http-api-bindings/src/completion/mod.rs
+++ b/crates/http-api-bindings/src/completion/mod.rs
@@ -1,10 +1,12 @@
 mod llama;
 mod mistral;
+mod openai;
 
 use std::sync::Arc;
 
 use llama::LlamaCppEngine;
 use mistral::MistralFIMEngine;
+use openai::OpenAICompletionEngine;
 use tabby_common::config::HttpModelConfig;
 use tabby_inference::CompletionStream;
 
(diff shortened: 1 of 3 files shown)


wsxiaoys:
Thank you for the pull request—I'm happy to merge it. The only concern is that since OpenAI has deprecated its completion API, referring to it as `OpenAICompletion` is a bit tricky because its definition is no longer present in any standard OpenAI client library. If there's no better alternative, perhaps consider renaming it to something like `vllm/completion` or `VllmCompletion`?

wsxiaoys:
Besides - could you also add an example for newly added completion config in https://github.com/TabbyML/tabby/edit/main/website/docs/administration/model.md? thank you!

Syst3m1cAn0maly:
> Thank you for the pull request—I'm happy to merge it. The only concern is that since OpenAI has deprecated its completion API, referring to it as `OpenAICompletion` is a bit tricky because its definition is no longer present in any standard OpenAI client library. If there's no better alternative, perhaps consider renaming it to something like `vllm/completion` or `VllmCompletion`?

I have mixed feelings about the OpenAI completion API as well as it is deprecated in their documentation...
However, it has become and still is a standard for many backends and is specifically mentioned as OpenAI completion API or OpenAI compatible API in their documentation and it is not limited to vLLM.
Some examples :

- in vLLM docs : https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#api-reference
- in Nvidia NIM docs : https://docs.nvidia.com/nim/large-language-models/latest/getting-started.html#openai-completion-request
- in LocalAI docs : https://localai.io/features/text-generation/#completions

PR #2629, opened 2024-07-12
ci: fix dockerfile warnings: 'as' and 'FROM' keywords' casing do not match

moqimoqidea:
When docker build then log show:

```txt
 2 warnings found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 9)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 49)
```


wsxiaoys:
Thanks for the fix

Issue #2638, opened 2024-07-14
Allow customizing context length in config.toml


wsxiaoys:
**Please describe the feature you want**

Tabby currently uses a default value of 1536 for code completion context, which is not configurable. We should expose the configuration in config.toml.

Code location: https://github.com/TabbyML/tabby/blob/7474711d3c344f64a725e6867d4d93198a54fa95/crates/tabby/src/services/completion.rs#L269

https://github.com/TabbyML/tabby/blob/7474711d3c344f64a725e6867d4d93198a54fa95/crates/tabby-common/src/config.rs#L184

https://github.com/TabbyML/tabby/blob/7474711d3c344f64a725e6867d4d93198a54fa95/crates/llama-cpp-server/src/supervisor.rs#L56

**Additional context**

---
Please reply with a 👍 if you want this feature.

Sma1lboy:
Would it be alright if I take this one? I am interested in contributing to this repository both now and in the future.

wsxiaoys:
Wei has started working on this at https://github.com/TabbyML/tabby/pull/2649 - I'll open more good first issues this week.

PR #2649, opened 2024-07-15
feat(config): support configure input length

xxs-wallace:
close https://github.com/TabbyML/tabby/issues/2638

### DIFF
--- a/crates/llama-cpp-server/src/supervisor.rs
+++ b/crates/llama-cpp-server/src/supervisor.rs
@@ -25,6 +25,7 @@ impl LlamaCppSupervisor {
         parallelism: u8,
         chat_template: Option<String>,
         enable_fast_attention: bool,
+        llama_context_size: usize,
     ) -> LlamaCppSupervisor {
         let Some(binary_name) = find_binary_name() else {
             panic!("Failed to locate llama-server binary, please make sure you have llama-server binary locates in the same directory as the current executable.");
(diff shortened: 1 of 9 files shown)


crates/llama-cpp-server/src/supervisor.rs
```diff
@@ -25,6 +25,7 @@ impl LlamaCppSupervisor {
         parallelism: u8,
         chat_template: Option<String>,
         enable_fast_attention: bool,
+        llama_context_size: usize,
```
wsxiaoys, line 28:
```suggestion
        context_size: usize,
```

crates/tabby-common/src/config.rs
```diff
@@ -186,6 +187,8 @@ pub struct ModelConfigGroup {
     pub chat: Option<ModelConfig>,
     #[serde(default = "default_embedding_config")]
     pub embedding: ModelConfig,
+
+    pub completion_max_input_length: usize,
```
wsxiaoys, line 191: need `serde(default=...)`?

crates/tabby/src/main.rs
```diff
@@ -136,5 +136,6 @@ fn to_local_config(model: &str, parallelism: u8, device: &Device) -> ModelConfig
         parallelism,
         num_gpu_layers,
         enable_fast_attention: None,
+        context_size: 4096,
```
wsxiaoys, line 139: is there a way to reuse default definition in `config.rs`?

crates/tabby/src/services/completion.rs
```diff
@@ -230,6 +230,7 @@ pub struct CompletionService {
     engine: Arc<CodeGeneration>,
     logger: Arc<dyn EventLogger>,
     prompt_builder: completion_prompt::PromptBuilder,
+    complete_max_input_length: usize,
```
wsxiaoys, line 233:
```suggestion
    max_input_length: usize,
```

wsxiaoys (changes requested):

crates/tabby/src/main.rs
xxs-wallace: yes, but we must use it as a public function

crates/llama-cpp-server/src/supervisor.rs
xxs-wallace: done

Issue #2633, opened 2024-07-13
Use int8 reranking for search scoring


wsxiaoys:
**Please describe the feature you want**
1. we stored binarized vector tokens in https://github.com/TabbyML/tabby/blob/67ae5ab547406d71d74872bde875c16ba47e029d/crates/tabby-common/src/index/mod.rs#L44
2. at search time, only binarized vector tokens are used for matching + scoring: https://github.com/TabbyML/tabby/blob/9da18587f5a80df497f86827395cd9f769c98f63/crates/tabby/src/services/doc/tantivy.rs#L47
3. we'd like to reuse https://docs.rs/tantivy/latest/tantivy/collector/trait.CustomScorer.html to compute int8 based score for reranking / weighting purpose.

**Additional context**

---
Please reply with a 👍 if you want this feature.

PR #2677, opened 2024-07-17
feat(index): add attributes for origin values

xxs-wallace:
one of https://github.com/TabbyML/tabby/issues/2633

# Background

We want to sort the result document again with the origin vector to get more accurate result. So we must store the origin vector as attributes values at first.

I will re-sort the result document in next PR.


umialpha:
@xxs-wallace are u still working on this work item?  i am also interested in this feature. If you are still working on this, maybe we can work together?

Issue #2654, opened 2024-07-16
Reuse llama-server for models supporting both chat / fim completion (e.g Codestral)


wsxiaoys:
**Please describe the feature you want**

Related: https://github.com/TabbyML/tabby/issues/2652

This allows a local deployment could use fewer vram / computing in local setup

**Additional context**
Code location:
1. https://github.com/TabbyML/tabby/blob/abdb0ef124a453278dcb8366b0e2fcab08201cb6/crates/llama-cpp-server/src/lib.rs#L160-L177
2. https://github.com/TabbyML/tabby/blob/abdb0ef124a453278dcb8366b0e2fcab08201cb6/crates/llama-cpp-server/src/lib.rs#L179-L195

---
Please reply with a 👍 if you want this feature.

sirebellum:
Taking a look at this, just fyi to anyone else looking at this issue. Will update with PR.

sirebellum:
Hi @wsxiaoys, this is my first time coding in rust. I am well versed in c++ so I don't anticipate any major issues, but I wanted to check with you to see how you anticipated this issue being solved. Was there a specific implementation that you envisioned? Thanks!

wsxiaoys:
Releasing in 0.16 - If you're interested, you might test it earlier at https://github.com/TabbyML/tabby/releases/tag/v0.16.0-rc.1.

wsxiaoys:
Released in 0.16

PR #2695, opened 2024-07-19
Use same backend when invoking chat and completion with same model

sirebellum:
In response to [issue 2654](https://github.com/TabbyML/tabby/issues/2654)


wsxiaoys:
Fixed in https://github.com/TabbyML/tabby/pull/2812

PR #2712, opened 2024-07-23
feat(vscode): support client selected context

wwayne:
## Context
We want to enable users to manually select context. For example, users can select a piece of code and add it to the context so that it will be included with their message next time.

User selected context as shown in the following screenshots:

## Screen record
For testing purposes, adding client context is triggered via the command palette.
https://jam.dev/c/8477d02f-9e37-4ac7-8bad-ae87d1b11778

## Overview of adding client context
1. **Editor adds new context:** By using the new method exposed on the tabby-chat-panel, `this.client?.addClientSelectedContext(Context)`
2. **Receiving context:** tabby-ui receives the context and saves it into its new state `clientSelectedContext`
3. **Displaying context:** tabby-ui displays the clientSelectedContext (as shown in the above screenshot)
4. **Removing context:** tabby-ui can remove the context using its own method `removeClientSelectedContext`
5. **Submitting the message:** When a message is submitted, tabby-ui will send the context to the Editor using `onSubmitMessage(value, clientSelectedContext)`. The Editor will then add these clientSelectedContext entries into `relevantContext` before sending them to the server.

## Note
Currently, there is no way to `addClientSelectedContext` from the user interface, so from the user's perspective, there will be no difference after merging this PR.

\cc @liangfung

### DIFF
--- a/clients/tabby-chat-panel/src/index.ts
+++ b/clients/tabby-chat-panel/src/index.ts
@@ -37,12 +37,13 @@ export interface ServerApi {
   sendMessage: (message: ChatMessage) => void
   showError: (error: ErrorMessage) => void
   cleanError: () => void
+  addClientSelectedContext: (context: Context) => void
 }
 
 export interface ClientApi {
   navigate: (context: Context, opts?: NavigateOpts) => void
   refresh: () => Promise<void>
-  onSubmitMessage?: (msg: string) => Promise<void>
+  onSubmitMessage?: (msg: string, onSubmitMessage?: Context[]) => Promise<void>
   onApplyInEditor?: (content: string) => void
 }
 
@@ -70,6 +71,7 @@ export function createServer(api: ServerApi): ClientApi {
       sendMessage: api.sendMessage,
       showError: api.showError,
       cleanError: api.cleanError,
+      addClientSelectedContext: api.addClientSelectedContext,
     },
   })
 }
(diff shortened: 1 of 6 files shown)


clients/tabby-chat-panel/src/index.ts
```diff
@@ -37,12 +37,13 @@ export interface ServerApi {
   sendMessage: (message: ChatMessage) => void
   showError: (error: ErrorMessage) => void
   cleanError: () => void
+  addClientSelectedContext: (context: Context) => void
```
wsxiaoys, line 40:
```suggestion
  addRelevantContext: (context: Context) => void
```

To match field name in line 53

wsxiaoys (changes requested):
Otherwise LGTM

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -251,7 +256,7 @@ export class ChatViewProvider implements WebviewViewProvider {
       ) {
         version = semver.coerce(serverInfo.health["version"]["git_describe"]);
       }
-      if (version && semver.lt(version, "0.12.0")) {
+      if (version && semver.lt(version, "0.16.0")) {
```
wsxiaoys, line 259: @icycodes fyi, VSCode expects new api

wsxiaoys (approved):

PR #2760, opened 2024-07-31
refactor(config): make HttpModelConfig.api_endpoint optional, so we c…

wsxiaoys:
…an use default value for certain model kind


crates/tabby-common/src/config.rs
```diff
@@ -248,9 +248,6 @@ pub struct HttpModelConfig {
     /// Used by Completion API to construct a chat model.
     #[builder(default)]
     pub chat_template: Option<String>,
-
-    #[builder(default)]
-    pub max_input_length: usize,
```
wsxiaoys, line 253: @xxs-wallace this is wrongly added in https://github.com/TabbyML/tabby/pull/2649, removing

Issue #2581, opened 2024-07-05
add IDE information (e.g vim / vscode / ...) to `~/.tabby/events` log.


wsxiaoys:
For third party integrations to collect such information

Sma1lboy:
It seems nobody working on this issue. I'll take look.

PR #2781, opened 2024-08-03
feat: Add general IDE info to User-Agent for Tabby clients

Sma1lboy:
This addresses the remaining part of the issue mentioned in #2581.

- Retrieve client info based on the client configuration during client initialization.

Example User-Agent format:

```
Node.js/v20.9.0 Visual-Studio-Code-desktop/1.91.1 TabbyML.vscode-tabby/1.8.0-dev
```

### DIFF
--- a/clients/tabby-agent/src/TabbyAgent.ts
+++ b/clients/tabby-agent/src/TabbyAgent.ts
@@ -71,7 +71,7 @@ export class TabbyAgent extends EventEmitter implements Agent {
   private tryingConnectTimer: ReturnType<typeof setInterval>;
   static readonly submitStatsInterval = 1000 * 60 * 60 * 24; // 24h
   private submitStatsTimer: ReturnType<typeof setInterval>;
-  private clientInfoStr: string = "";
+  private userAgentHeader: string = "";
 
   constructor() {
     super();
@@ -164,7 +164,7 @@ export class TabbyAgent extends EventEmitter implements Agent {
       baseUrl: endpoint,
       headers: {
         Authorization: auth,
-        "User-Agent": this.clientInfoStr,
+        "User-Agent": this.userAgentHeader,
         ...this.config.server.requestHeaders,
       },
       /** dispatcher do not exist in {@link RequestInit} in browser env. */
@@ -396,7 +396,7 @@ export class TabbyAgent extends EventEmitter implements Agent {
       }
     }
     await this.anonymousUsageLogger.init({ dataStore: this.dataStore });
-    this.clientInfoStr = this.clientInfoToString(options.clientProperties?.session);
+    this.userAgentHeader = this.clientInfoToString(options.clientProperties?.session);
     if (options.clientProperties) {
       if (options.clientProperties.session) {
         Object.entries(options.clientProperties.session).forEach(([key, value]) => {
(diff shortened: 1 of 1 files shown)


Sma1lboy:
Should I do another PR for intellij? or I can put them together

Sma1lboy:
Sure, I'll take a look tonight

Sma1lboy:
@icycodes Thanks for your suggestions. I've realized that most of my previous code was focused on creating a parameter for ideInfo that could be transmitted by the client. It wasn't until I read the section of the code about lifecycle that I discovered this process had already been written there. :(
I deleted the unnecessary code and cleaned up the user agent format. By the way, if I want to get the current version of the user agent, do I need to read the package.json? I haven't found any code that stores "packagejson" context.

icycodes:
> if I want to get the current version of the user agent, do I need to read the package.json? I haven't found any code that stores "packagejson" context.

Simply import the "package.json" should be ok. Here is an [example](https://github.com/TabbyML/tabby/blob/main/clients/tabby-agent/src/lsp/Server.ts#L93).

clients/tabby-agent/src/TabbyAgent.ts
```diff
@@ -1010,4 +1018,16 @@ export class TabbyAgent extends EventEmitter implements Agent {
       return null;
     }
   }
+
+  private clientInfoToString(session: Record<string, any> | undefined): string {
+    const nodeInfo = `Node.js/${process.version}`;
```
icycodes, line 1023: We need to check `isBrowser` here. In browser environment, use `navigator?.userAgent` instead. Here is an [example](https://github.com/TabbyML/tabby/blob/main/clients/tabby-agent/src/AnonymousUsageLogger.ts#L16).

clients/tabby-agent/src/TabbyAgent.ts
```diff
@@ -699,6 +704,9 @@ export class TabbyAgent extends EventEmitter implements Agent {
           temperature,
         },
         signal: this.createAbortSignal({ signal }),
+        headers: {
+          "User-Agent": this.clientInfoStr,
```
icycodes, line 708: I think we can add headers when [create http client](https://github.com/TabbyML/tabby/blob/main/clients/tabby-agent/src/TabbyAgent.ts#L163).

clients/tabby-agent/src/TabbyAgent.ts
```diff
@@ -393,6 +395,7 @@ export class TabbyAgent extends EventEmitter implements Agent {
       }
     }
     await this.anonymousUsageLogger.init({ dataStore: this.dataStore });
+    this.clientInfoStr = this.clientInfoToString(undefined);
```
icycodes, line 398:
```suggestion
    this.clientInfoStr = this.clientInfoToString(options.clientProperties?.session);
```
This should combine L398 and L410.

icycodes (changes requested):

clients/tabby-agent/src/TabbyAgent.ts
Sma1lboy: I moved user agent into setupApi func

clients/tabby-agent/src/TabbyAgent.ts
```diff
@@ -70,6 +71,7 @@ export class TabbyAgent extends EventEmitter implements Agent {
   private tryingConnectTimer: ReturnType<typeof setInterval>;
   static readonly submitStatsInterval = 1000 * 60 * 60 * 24; // 24h
   private submitStatsTimer: ReturnType<typeof setInterval>;
+  private clientInfoStr: string = "";
```
wsxiaoys, line 74:
```suggestion
  private userAgentHeader: string = "";
```

icycodes (approved):

PR #2800, opened 2024-08-08
feat(webserver): add code_snippets field to AnswerRequest

wsxiaoys:


crates/tabby-common/src/api/answer.rs
```diff
@@ -32,6 +37,13 @@ pub struct AnswerRequest {
     pub collect_relevant_code_using_user_message: bool,
 }
 
+#[derive(Deserialize, ToSchema)]
+pub struct AnswerCodeSnippet {
+    pub language: Option<String>,
```
wsxiaoys, line 42:
```suggestion
```

PR #2831, opened 2024-08-11
chore(index): skip web documents if body is empty

wsxiaoys:
Fixing bad cases like below:

image

@xxs-wallace


.changes/unreleased/Fixed and Improvements-20240811-124728.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed and Improvements
+body: Skip wweb documents if body is empty
```
wsxiaoys, line 2:
```suggestion
body: Skip web documents if body is empty
```

Issue #2811, opened 2024-08-08
Can't seem to get Tabby to run on GPU


BurnyLlama:
**Describe the bug**

(*This might be me doing something wrong, and not a bug!*)

I can't seem to get Tabby to run using my GPU (Radeon 6600 XT). Not with ROC, (which I believe is unsupported for my device) nor with Vulkan — which I believe should be supported?

Whenever I run tabby (`serve --model DeepseekCoder-1.3B --chat-model Qwen2-1.5B-Instruct --device vulkan`) and go into the Web UI to ask something I check my CPU and GPU usage (using `btop` and `amdgpu_top` respectively) I see my CPU usage spiking and almost no effect on the GPU. (This is when running `v0.15.0-r2` or compiling `v0.15.0-r3` myself).

If I try to use `v0.14.0` (same options) I instead get this:

```
2024-08-08T11:19:34.328360Z  WARN llama_cpp_server::supervisor: crates/llama-cpp-server/src/supervisor.rs:111: <embedding>: warning: see main README.md for information on enabling GPU BLAS support
```

**Information about your version**

See above.

**Information about your GPU**

From `vulkaninfo`:

```
Devices:
========
GPU0:
	apiVersion         = 1.3.270
	driverVersion      = 2.0.294
	vendorID           = 0x1002
	deviceID           = 0x73ff
	deviceType         = PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
	deviceName         = AMD Radeon RX 6600 XT
	driverID           = DRIVER_ID_AMD_PROPRIETARY
	driverName         = AMD proprietary driver
	driverInfo         = (AMD proprietary shader compiler)
	conformanceVersion = 1.3.3.1
	deviceUUID         = 00000000-2800-0000-0000-000000000000
	driverUUID         = 414d442d-4c49-4e55-582d-445256000000
GPU1:
	apiVersion         = 1.3.278
	driverVersion      = 24.1.5
	vendorID           = 0x1002
	deviceID           = 0x73ff
	deviceType         = PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
	deviceName         = AMD Radeon RX 6600 XT (RADV NAVI23)
	driverID           = DRIVER_ID_MESA_RADV
	driverName         = radv
	driverInfo         = Mesa 24.1.5
	conformanceVersion = 1.3.0.0
	deviceUUID         = 00000000-2800-0000-0000-000000000000
[...]

PR #2835, opened 2024-08-12
fix(llama-cpp-server): fix rocm build by setting `GGML_HIPBLAS`

richard-jfc:
Various `LLAMA_` build flags have been renamed to `GGML_` in llama-cpp-server (See https://github.com/ggerganov/llama.cpp/commit/f3f65429c44bb195a9195bfdc19a30a79709db7b) and need to be renamed in tabby to allow correctly targeting Metal/Rocm/Vulkan.

It appears that this was already done for Cuda.

May be related: https://github.com/TabbyML/tabby/issues/2811

### DIFF
--- a/crates/llama-cpp-server/build.rs
+++ b/crates/llama-cpp-server/build.rs
@@ -48,7 +48,7 @@ fn main() {
         ];
 
         let rocm_root = env::var("ROCM_ROOT").unwrap_or("/opt/rocm".to_string());
-        config.define("LLAMA_HIPBLAS", "ON");
+        config.define("GGML_HIPBLAS", "ON");
         config.define("CMAKE_C_COMPILER", format!("{}/llvm/bin/clang", rocm_root));
         config.define(
             "CMAKE_CXX_COMPILER",


wsxiaoys:
Have you verified if the PR fixes your issue?

cmake handles the migration with macros defined in https://demo.tabbyml.com/files/git/llama.cpp/-/blob/7a221b672e49dfae459b1af27210ba3f2b5419b6/CMakeLists.txt?plain=1#L101

So my guess is it won't fix any build issues you are currently encountering.

richard-jfc:
I built rocm v0.15.0 tag using docker. And have been running the built Rocm version via docker.

Without my change I was seeing a CPU spike for every chat request and no VRAM or GPU usage.

After my change I was seeing roughly 6GB VRAM and a GPU usage spike.

wsxiaoys:
It does seem that `GGML_HIPBLAS` is the flag not handled by the `llama_option_depr` migration. In that case, would you consider updating the PR to contain only the `GGML_HIPBLAS` change, since it's the only one you've verified?

richard-jfc:
Thanks, Done

Issue #2783, opened 2024-08-05
Code repository search refactor (use `source_id`) hotlist


wsxiaoys:
**Please describe the feature you want**
1. we have concept of `source_id` to represent a group of document.
https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/main/crates/tabby-common/src/index/mod.rs?plain=1#L25

* For web providers, it aggregates documents corresponding to a start url.
* For git / github providers, it aggregates documents of a single repository.

3. In doc query time, we're using `source_id` to restrict scope of documents
https://demo.tabbyml.com/files/github/TabbyML/tabby/-/search/main?q=source_ids_query

4. In code query time, we're still using legacy `canonical_git_url` to restrict scope of documents.
https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/main/crates/tabby/src/services/code.rs?plain=1#L87
https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/main/crates/tabby-common/src/index/code/mod.rs?plain=1#L73

We'd like to refactor code query to also using `source_id` as restriction (source id for code repository is currently set to canonical_git_url, but it should really being refactored to `${RepositoryKind}:${RepositoryID}`)
https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/main/ee/tabby-webserver/src/service/repository/mod.rs?plain=1#L70

- [x] refactor code query to use source id
(change `git_url` to `source_id` in https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/1837cf8cef9b712ea3ee4e9f4a39b60728ac446e/crates/tabby-common/src/api/code.rs?plain=1#L60)
- [ ] https://github.com/TabbyML/tabby/pull/2782
- [x] consider add `RepositoryKind::CONFIG` to make config based repository explicit: https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/main/ee/tabby-webserver/src/service/repository/mod.rs?plain=1#L251

**Additional context**

PR #2837, opened 2024-08-12
refactor(search): change git_url to source_id

xxs-wallace:
one of https://github.com/TabbyML/tabby/issues/2783

this PR is try to refactor code query to use source id instead of git url.

### DIFF
--- a/crates/tabby-common/src/api/code.rs
+++ b/crates/tabby-common/src/api/code.rs
@@ -58,7 +58,7 @@ pub enum CodeSearchError {
 
 #[derive(Deserialize, ToSchema)]
 pub struct CodeSearchQuery {
-    pub source_id: String,
+    pub git_url: String,
     pub filepath: Option<String>,
     pub language: Option<String>,
     pub content: String,
--- a/crates/tabby-common/src/api/code.rs
+++ b/crates/tabby-common/src/api/code.rs
@@ -58,7 +58,7 @@ pub enum CodeSearchError {
 
 #[derive(Deserialize, ToSchema)]
 pub struct CodeSearchQuery {
-    pub git_url: String,
+    pub source_id: String,
     pub filepath: Option<String>,
     pub language: Option<String>,
     pub content: String,
--- a/crates/tabby-common/src/index/code/mod.rs
+++ b/crates/tabby-common/src/index/code/mod.rs
@@ -61,11 +61,15 @@ pub fn code_search_query(
     chunk_tokens_query: Box<dyn Query>,
 ) -> BooleanQuery {
     let schema = IndexSchema::instance();
-    let mut subqueries = schema.source_sub_queries(corpus::CODE, &query.git_url);
+    let mut subqueries = vec![
+        (Occur::Must, chunk_tokens_query),
+        (
+            Occur::Must,
+            Box::new(schema.source_query(corpus::CODE, &query.git_url)),
+        ),
+    ];
 
     // language / git_url / filepath field shouldn't contribute to the score, mark them to 0.0.
-    subqueries.push((Occur::Must, chunk_tokens_query));
-
     if let Some(language) = query.language.as_deref() {
         subqueries.push((
             Occur::Must,
--- a/crates/tabby-common/src/index/code/mod.rs
+++ b/crates/tabby-common/src/index/code/mod.rs
@@ -15,6 +15,8 @@ pub mod fields {
     pub const CHUNK_LANGUAGE: &str = "chunk_language";
     pub const CHUNK_BODY: &str = "chunk_body";
     pub const CHUNK_START_LINE: &str = "chunk_start_line";
+
+    pub const CHUNK_SOURCE_ID: &str = "chunk_source_id";
 }
 
 fn language_query(language: &str) -> Box<TermQuery> {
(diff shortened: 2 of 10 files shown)


crates/tabby-common/src/index/code/mod.rs
```diff
@@ -70,7 +72,7 @@ pub fn code_search_query(
 ) -> BooleanQuery {
     let schema = IndexSchema::instance();
     let corpus_query = schema.corpus_query(corpus::CODE);
-    let git_url_query = git_url_query(&query.git_url);
+    let git_url_query = git_url_query(&query.source_id);
```
wsxiaoys, line 75: Need to replace `git_url_query` to `source_query` to match against the `source_id` field in index:

https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/48d9c084cc80d20bd439102cc0768920f6680e91/crates/tabby-common/src/index/mod.rs?plain=1#L102

crates/tabby-index/src/code/mod.rs
```diff
@@ -85,7 +85,7 @@ impl IndexAttributeBuilder<SourceCode> for CodeBuilder {
             for await (start_line, body) in CodeIntelligence::chunks(&text, &source_code.language) {
                 let attributes = json!({
                     code::fields::CHUNK_FILEPATH: source_code.filepath,
-                    code::fields::CHUNK_GIT_URL: source_code.git_url,
+                    code::fields::CHUNK_SOURCE_ID: source_code.source_id,
```
wsxiaoys, line 88: no need to add this, source_id is already indexed:

https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/48d9c084cc80d20bd439102cc0768920f6680e91/crates/tabby-common/src/index/mod.rs?plain=1#L25

crates/tabby-index/src/code/types.rs
```diff
@@ -10,7 +10,7 @@ use crate::indexer::{IndexId, ToIndexId};
 #[derive(Serialize, Deserialize, Clone)]
 pub struct SourceCode {
     pub id: String,
-    pub git_url: String,
+    pub source_id: String,
```
wsxiaoys, line 13: The refactor is to query against the index field `source_id`, there's no need to change field name to `source_id` every where.

wsxiaoys (changes requested):

crates/tabby-common/src/api/code.rs
```diff
@@ -58,7 +58,7 @@ pub enum CodeSearchError {
 
 #[derive(Deserialize, ToSchema)]
 pub struct CodeSearchQuery {
-    pub git_url: String,
+    pub source_id: String,
```
wsxiaoys, line 61:
```suggestion
    pub git_url: String,
```

Given that we're also passing `git_url` from frontend (in `answer.rs`), we can keep the field name for simplicity.

crates/tabby-index/src/code/mod.rs
xxs-wallace: But it is attribute, not index.
Do you mean that we can just remote this field `CHUNK_GIT_URL`?

crates/tabby-common/src/index/code/mod.rs
```diff
@@ -69,21 +61,10 @@ pub fn code_search_query(
     chunk_tokens_query: Box<dyn Query>,
 ) -> BooleanQuery {
     let schema = IndexSchema::instance();
-    let corpus_query = schema.corpus_query(corpus::CODE);
-    let git_url_query = git_url_query(&query.git_url);
+    let mut subqueries = schema.source_sub_queries(corpus::CODE, &query.git_url);
```
wsxiaoys, line 64:
```suggestion
    let mut subqueries = vec![
      (Occur::Must, Box::new(ConstScoreQuery::new(schema.source_query())))
    ]
```

No need to extract `source_sub_queries`? it probably leaks the implementation details.

crates/tabby/src/services/completion/completion_prompt.rs
```diff
@@ -182,7 +182,7 @@ async fn collect_snippets(
     content: &str,
 ) -> Vec<Snippet> {
     let query = CodeSearchQuery {
-        git_url: git_url.to_owned(),
+        git_url: RepositoryConfig::canonicalize_url(git_url),
```
wsxiaoys, line 185: CodeSearch already handles canonicalization / closest_match internally, no need to make this change.

https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/48d9c084cc80d20bd439102cc0768920f6680e91/crates/tabby/src/services/code.rs?plain=1#L78

ee/tabby-webserver/src/service/answer.rs
wsxiaoys, line 1: revert changes?

wsxiaoys:
Please also fix all CI checks

crates/tabby-common/src/api/code.rs
xxs-wallace: ok

crates/tabby-common/src/index/code/mod.rs
xxs-wallace: Yes. But the origin `source_query` has add `corpus_query` beside it. I just keep the logic like creating `source_query`

ee/tabby-webserver/src/service/answer.rs
wsxiaoys, line 1: revert changes?

crates/tabby-common/src/index/code/mod.rs
xxs-wallace: Although all query are composed with `Must`, I just want to keep the same logic with previous code.

wsxiaoys: I don't get it - nested must is still must? it doesn't change the logic if with my suggested change.

wsxiaoys (approved):

PR #2838, opened 2024-08-12
refactor(tabby-db) date time

xxs-wallace:
Get rid of DateTimeUtc and DbNullable in tabby-db.


ee/tabby-db/src/integrations.rs
```diff
@@ -12,8 +13,8 @@ pub struct IntegrationDAO {
     pub display_name: String,
     pub access_token: String,
     pub api_base: Option<String>,
-    pub created_at: DateTimeUtc,
-    pub updated_at: DateTimeUtc,
+    pub created_at: NaiveDateTime,
```
wsxiaoys, line 16: should use `DateTime<UTC>`. Applied to other DAOs' as well

xxs-wallace: But the macro `query!` only support field of `NaiveDateTime`. Otherwise we must implement `From<NaiveDateTime>` for other types

xxs-wallace: I check all code and I found that we did not read this field directly. Instead, we construct `Integration` and read it. So I think using `DateTime<Utc>` in `Integration` is enough.

wsxiaoys: It's documented in the deprecation notes:

https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/48d9c084cc80d20bd439102cc0768920f6680e91/ee/tabby-db/src/lib.rs?plain=1#L299

Example:
1. https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/48d9c084cc80d20bd439102cc0768920f6680e91/
ee/tabby-db/src/threads.rs?plain=1#L78
2. https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/48d9c084cc80d20bd439102cc0768920f6680e91/ee/tabby-db/src/threads.rs?plain=1#L189

ee/tabby-db/src/lib.rs
```diff
@@ -287,124 +271,17 @@ where
...
-}
-
-impl Sub<Duration> for DateTimeUtc {
-    type Output = Self;
-
-    fn sub(self, rhs: Duration) -> Self::Output {
-        ((self.0) - rhs).into()
-    }
-}
+pub type DateTimeUtc = DateTime<Utc>;
```
wsxiaoys, line 281: do not use type alias, just use `DateTime<Utc>` every where

ee/tabby-db/src/lib.rs
```diff
@@ -229,20 +224,12 @@ impl DbConn {
...
-    for<'a> Decode<'a, Sqlite> + for<'a> Encode<'a, Sqlite> + Type<Sqlite>
-{
-}
-impl DbNullable for DateTimeUtc {}
-
 #[derive(Default)]
-pub struct DbOption<T>(Option<T>)
-where
-    T: DbNullable;
+pub struct DbOption<T>(Option<T>);
```
wsxiaoys, line 228: DbOption can also be removed?

ee/tabby-db/src/lib.rs
```diff
@@ -287,124 +271,17 @@ where
...
-    pub fn from_timestamp(secs: i64, subsec_nanos: u32) -> Option<Self> {
-        DateTime::from_timestamp(secs, subsec_nanos).map(Self)
-    }
-
-    fn as_sqlite_datetime(&self) -> String {
-        self.0.format("%F %X").to_string()
-    }
+pub fn sqlite_datetime_format(t: &DateTimeUtc) -> String {
+    t.format("%F %X").to_string()
 }
```
wsxiaoys, line 285: Create a trait for this?
```
trait AsSqliteDateTimeString {
  fn as_sqlite_datetime(&self) -> String
}

impl AsSqliteDateTimeString for DateTime<Utc> {
}
```

ee/tabby-db/src/lib.rs
xxs-wallace: Yes. I will do it

ee/tabby-db/src/lib.rs
xxs-wallace: Ok.

ee/tabby-db/src/lib.rs
xxs-wallace: OK, it has been removed.

wsxiaoys (approved):
Please fix unit test, otherwise LGTM

wsxiaoys (changes requested):
Undo LGTM.

Should implement specialized `DateTime<Utc>` formatting.

wsxiaoys (approved):

PR #2852, opened 2024-08-12
refactor(webserver): remove deprecated /v1beta/answer route

wsxiaoys:
@liangfung fyi, this shall merge after frontend switch to graphql based api


xxs-wallace (approved):
LGTM for backend

PR #2861, opened 2024-08-13
fix(db): disable sqlx-sqlite implementation for encoding DateTime<Utc>

wsxiaoys:
See https://github.com/wsxiaoys/sqlx/commit/8ca573cc0e323a637261a8f2cbc4a7c3c7b4e133 for patch against sqlx 0.7.4


xxs-wallace (approved):
LGTM

Issue #2830, opened 2024-08-11
[VSCode] Combine all the chat streaming edit steps as one undo-redo step


wsxiaoys:
**Please describe the feature you want**
As the chat editing is applied multi times as streaming, it requires a lot of ctrl+z for user to undo it.

---
Please reply with a 👍 if you want this feature.

Sma1lboy:
I'll take this one if nobody else is doing it. I'll update the PR ASAP :^)

icycodes:
Please see https://github.com/TabbyML/tabby/blob/main/clients/tabby-agent/src/lsp/protocol.ts#L379 for the current implementation. The current implementation uses the LSP method `workspace/applyEdit` multiple times to make the edit, which does not have the undo-redo parameters. In VSCode, there is an [interface](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts#L1264) supporting setting an undo-redo stop programmatically.

My suggested approach
  
1. Extend the current `workspace/applyEdit` to add relevant parameters. Alternatively, add a new server-to-client method `tabby/workspace/applyEdit` to be used primarily.
2. Implement setting undo-redo stop on the server-side using new parameters.
3. Add a middleware to process the extended `workspace/applyEdit` or add new support for `tabby/workspace/applyEdit`.

PR #2872, opened 2024-08-14
feat(vscode):  one undo-redo step for one chat stream

Sma1lboy:
This PR addresses #2830.

This solves the issue where a chat stream can be undone with a single undo-redo operation.

btw, I'm thinking if I should adding undoStop after resolveEdit

### DIFF
--- a/clients/tabby-agent/src/lsp/ChatEditProvider.ts
+++ b/clients/tabby-agent/src/lsp/ChatEditProvider.ts
@@ -1,4 +1,4 @@
-import { Range, Location, Connection, CancellationToken } from "vscode-languageserver";
+import { Range, Location, Connection, CancellationToken, WorkspaceEdit } from "vscode-languageserver";
 import {
   ChatEditToken,
   ChatEditRequest,
@@ -12,6 +12,7 @@ import {
   ChatEditDocumentTooLongError,
   ChatEditCommandTooLongError,
   ChatEditMutexError,
+  ApplyWorkspaceEditRequest,
 } from "./protocol";
 import { TextDocuments } from "./TextDocuments";
 import { TextDocument } from "vscode-languageserver-textdocument";
@@ -229,7 +230,8 @@ export class ChatEditProvider {
         }
       }
     });
-    await this.connection.workspace.applyEdit({
+
+    await this.connection.sendRequest(ApplyWorkspaceEditRequest.type, {
       edit: {
         changes: {
           [params.location.uri]: [
@@ -240,6 +242,10 @@ export class ChatEditProvider {
           ],
         },
       },
+      options: {
+        undoStopBefore: false,
+        undoStopAfter: false,
+      },
     });
     return true;
   }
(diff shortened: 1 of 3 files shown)


clients/tabby-agent/src/lsp/ChatEditProvider.ts
```diff
@@ -249,121 +256,129 @@ export class ChatEditProvider {
...
+      this.mutexAbortController = null;
+    }
+  }
+
+  private async applyWorkspaceEdit(params: ApplyWorkspaceEditParams): Promise<boolean> {
+    try {
+      const result = await this.connection.sendRequest(ApplyWorkspaceEditRequest.type, params);
+      return result;
+    } catch (error) {
+      console.warn("Extended applyEdit failed, falling back to standard method:", error);
```
icycodes, line 370: Please do not use console for logging. You can find many examples using `logger`.

clients/tabby-agent/src/lsp/ChatEditProvider.ts
```diff
@@ -249,121 +256,129 @@ export class ChatEditProvider {
...
+      const match = reg.exec(buffer);
+      if (match && match[0]) {
+        if (match[0] === responseDocumentTag[0]) {
+          return "document";
+        } else if (match[0] === responseCommentTag?.[0]) {
+          return "comment";
+        }
+      }
+      return false;
+    };
```
icycodes, line 318: It seems there is a bug after the refactor of `findOpenTag` and `processBuffer`. The opening tag (`<GENERATEDCODE>`) is not being skipped after it is found, instead, it is always included in the content after editing.

clients/tabby-agent/src/lsp/ChatEditProvider.ts
```diff
@@ -249,121 +256,129 @@ export class ChatEditProvider {
...
       }
+    } finally {
+      this.currentEdit = null;
+      this.mutexAbortController = null;
+    }
+  }
+
+  private async applyWorkspaceEdit(params: ApplyWorkspaceEditParams): Promise<boolean> {
+    try {
+      const result = await this.connection.sendRequest(ApplyWorkspaceEditRequest.type, params);
```
icycodes, line 367: We can add a client capabilities flag to indicate if the client supports this method, rather than using a try-catch fallback. However, it is acceptable for now.   
I believe this improvement could be a separate PR after this PR is merged.

icycodes (changes requested):

clients/tabby-agent/src/lsp/ChatEditProvider.ts
Sma1lboy: sorry for late response, I'll put this PR on my calendar

clients/tabby-agent/src/lsp/ChatEditProvider.ts
Sma1lboy: I'm sorry, I made a logical mistake, but I have fixed it. Sorry for the late response.

icycodes (approved):

Issue #2873, opened 2024-08-14
[VSCode] Add a 'Generate Commit Message' icon button


Sma1lboy:
Using a command to perform this action might not be very intuitive.

<img width="643" alt="image" src="https://github.com/user-attachments/assets/c91c381a-ad19-4910-a964-98b03d5ad550">

Maybe adding a 'Generate Commit Message' icon in the Source Control section of VSCode to provide a better UI experience for users.

<img width="376" alt="image" src="https://github.com/user-attachments/assets/d047d73e-9ddc-4b16-9d79-d14541d7e283">

---
Please reply with a 👍 if you want this feature.

wsxiaoys:
The last time we checked, it wasn't possible to build such a feature with VSCode (GitHub Copilot itself used a private API to achieve this).

However, it's good to keep track of this in case the situation changes in the future (or has already changed).

Sma1lboy:
Follow up: I found a way to adding a icon on source control navigation, I will update PR ASAP

wsxiaoys:
related: https://github.com/microsoft/vscode/issues/195474

PR #2878, opened 2024-08-14
feat(vscode): adding commit message button in source control title

Sma1lboy:
Trying to improve user experience and address issue #2873.

Does not fully support multi-repository situations as there is no built-in feature in VSCode to get the currently selected repository.
Works perfectly in single repository scenarios.
View:
image

Next Steps:

Track related VSCode API support in the future.
Consider changing the icon.

### DIFF
--- a/clients/vscode/package.json
+++ b/clients/vscode/package.json
@@ -182,7 +182,7 @@
         },
         {
           "command": "tabby.chat.generateCommitMessage",
-          "when": "tabby.chatEnabled"
+          "when": "false"
         },
         {
           "command": "tabby.chat.edit.start",
--- a/clients/vscode/package.json
+++ b/clients/vscode/package.json
@@ -111,9 +111,9 @@
       },
       {
         "command": "tabby.chat.generateCommitMessage",
-        "title": "Generate Commit Message with Tabby Agent",
+        "title": "Generate Commit Message",
         "category": "Tabby",
-        "icon": "assets/logo.png"
+        "icon": "assets/magic-wand.svg"
       },
       {
         "command": "tabby.chat.edit.start",
--- a/clients/vscode/package.json
+++ b/clients/vscode/package.json
@@ -111,8 +111,9 @@
       },
       {
         "command": "tabby.chat.generateCommitMessage",
-        "title": "Generate Commit Message",
-        "category": "Tabby"
+        "title": "Generate Commit Message with Tabby Agent",
+        "category": "Tabby",
+        "icon": "assets/logo.png"
       },
       {
         "command": "tabby.chat.edit.start",
@@ -204,6 +205,13 @@
         {
           "submenu": "tabby.submenu",
           "group": "2_tabby"
+        }
+      ],
+      "scm/title": [
+        {
+          "command": "tabby.chat.generateCommitMessage",
+          "when": "tabby.chatEnabled",
+          "group": "navigation"
         }
       ]
     },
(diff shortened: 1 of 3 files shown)


clients/vscode/package.json
```diff
@@ -106,8 +106,9 @@
       },
       {
         "command": "tabby.chat.generateCommitMessage",
-        "title": "Generate Commit Message",
-        "category": "Tabby"
+        "title": "Generate Commit Message with Tabby Agent",
```
wsxiaoys, line 109:
```suggestion
        "title": "Generate Commit Message",
```

clients/vscode/package.json
```diff
@@ -106,8 +106,9 @@
...
+        "category": "Tabby",
+        "icon": "assets/logo.png"
```
wsxiaoys, line 111: maybe `magic wind` svg from https://www.radix-ui.com/icons

wsxiaoys (commented):
Otherwsie LG

icycodes:
Please ensure to test opening a workspace containing multiple git repositories, such as open the Tabby repo. 
In this case, `Generate Commit Message` requires users to select a repo first.   

As the actions tool icon is displayed on each repo title, users may not need to manually select a repo when click the tool icon. To enhance user experience in this situation, please consider add params to the command to improve it.
Screenshot from 2024-08-15 10-31-59

Sma1lboy:
Hi @icycodes! I realized this problem while working on this feature. Unfortunately, dealing with multi-repositories is challenging, and there is no built-in VSCode feature to help with this. Disabling this feature for multi-repositories and keeping only one icon in the navigation might be a good solution. I did try to support multi-repositories as if they were a single repo, but the result was that I'd have to completely rewrite the source control part, which I don't think is worth it at this point. At least now, single repository projects have a better experience. I will continue tracking this feature until VSCode makes some changes.

icycodes:
> Hi @icycodes! I realized this problem while working on this feature. Unfortunately, dealing with multi-repositories is challenging, and there is no built-in VSCode feature to help with this. Disabling this feature for multi-repositories and keeping only one icon in the navigation might be a good solution. I did try to support multi-repositories as if they were a single repo, but the result was that I'd have to completely rewrite the source control part, which I don't think is worth it at this point. At least now, single repository projects have a better experience. I will continue tracking this feature until VSCode makes some changes.

Yes, I agree that it may not be worth it if we need to rewrite the SCM part.   
But, I think there should be a possible approach to improve it with less effort. By adding an argument to the command, we could provide some useful context. The argument is null when triggered from the command palette, but when the command is triggered by the button on a repository, a context could be received. I have not rechecked it recently, so you can double-check if these adjustments could help.

clients/vscode/src/Commands.ts
```diff
@@ -465,42 +465,43 @@ export class Commands {
       };
       await this.client.chat.resolveEdit({ location, action: "discard" });
     },
-    "chat.generateCommitMessage": async () => {
+    "chat.generateCommitMessage": async (repository?: Repository) => {
```
wsxiaoys, line 468: @icycodes FYI

image

clients/vscode/package.json
```diff
@@ -181,7 +182,7 @@
         },
         {
           "command": "tabby.chat.generateCommitMessage",
-          "when": "tabby.chatEnabled"
+          "when": "false"
```
icycodes, line 185: I think we can keep the access from command palette.

wsxiaoys: This is following github copilot's practice

icycodes: LGTM

icycodes (approved):

PR #2888, opened 2024-08-15
feat(crawler): add preset web crawler urls

xxs-wallace:
### Background

give some preset web url for user to crawl.

User can active one url to start a crawler job, or inactive it to cancel this job.


xxs-wallace:
duplicated with https://github.com/TabbyML/tabby/pull/2898

PR #2892, opened 2024-08-15
feat(index): BREAKING: change source_id definition to make it across corpus

wsxiaoys:
1. We change the definition of `source_id` in index to be across corpus.
2. We set source_id for code corpus as format of `git:$id`, `provided_repository:$id`, `config:$id`, ...

### DIFF
--- a/crates/tabby-common/src/index/mod.rs
+++ b/crates/tabby-common/src/index/mod.rs
@@ -62,7 +62,7 @@ impl IndexSchema {
         let mut builder = Schema::builder();
 
         let field_corpus = builder.add_text_field("corpus", STRING | FAST);
-        let field_source_id = builder.add_text_field("source_id", STRING | FAST);
+        let field_source_id = builder.add_text_field("source_id_v2", STRING | FAST);
         let field_id = builder.add_text_field("id", STRING | STORED);
 
         let field_updated_at = builder.add_date_field(FIELD_UPDATED_AT, INDEXED);
--- a/crates/tabby-common/src/index/mod.rs
+++ b/crates/tabby-common/src/index/mod.rs
@@ -71,6 +71,8 @@ const FIELD_UPDATED_AT: &str = "updated_at";
 pub mod corpus {
     pub const CODE: &str = "code";
     pub const WEB: &str = "web";
+
+    pub const ALL: [&str; 2] = [CODE, WEB];
 }
 
 impl IndexSchema {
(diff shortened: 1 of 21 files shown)


crates/tabby-common/src/index/mod.rs
```diff
@@ -62,7 +82,7 @@ impl IndexSchema {
         let mut builder = Schema::builder();
 
         let field_corpus = builder.add_text_field("corpus", STRING | FAST);
-        let field_source_id = builder.add_text_field("source_id", STRING | FAST);
+        let field_source_id = builder.add_text_field("source_id_v2", STRING | FAST);
```
wsxiaoys, line 85: @xxs-wallace we manually change the field name so the whole index will be rebuild automatically

crates/tabby-common/src/index/mod.rs
```diff
@@ -13,6 +13,26 @@ use tantivy::{
...
+///           +----------------+      attributes are stored but not indexed.
+///                   |
+///                   v
+///           +----------------+
+///           |     chunk      | <--- Each chunk has a unique identifier (chunk_id) within
+///           |   (chunk_id)   |      the document. Chunk is the unit being retrieved during
+///           +----------------+      the search process.
+///
+/// Across the corpus, there is a concept of source_id, which identifies a group of documents.
+/// It is usually used to identify the source of the document, such as a Github connection or a web document crawl.
```
wsxiaoys, line 35: @xxs-wallace fyi

xxs-wallace: Since source id is not composed by git url, could there be several document for one github repository?

crates/tabby/src/services/code.rs
```diff
@@ -75,16 +75,16 @@ impl CodeSearchImpl {
             })
             .await?;
 
-        let Some(git_url) = closest_match(&query.git_url, repos.iter()) else {
+        let Some(source_id) = closest_match(&query.git_url, repos.iter()) else {
```
xxs-wallace, line 78: `closet_match` will still return `git_url`. But the source_id shall be like `config_index_to_id` ? Is it still right?

crates/tabby-common/src/index/mod.rs
wsxiaoys: yes - and it's the indended behavior.

crates/tabby/src/services/code.rs
wsxiaoys: closet match has been changed to return source id already

xxs-wallace (approved):
LGTM

PR #2898, opened 2024-08-16
feat(document): add web document for preset / custom docs

xxs-wallace:
### Background
We want to add unique constraint for `name` field. But it is impossible to add unique constraint in exist table because the origin filed of this table has no value.

The next step, we will add some preset document for user to crawler and search.

The origin table `web_crawler_urls` will be replaced by new table `web_documents` after all api interface is ready.


ee/tabby-db/src/web_crawler.rs
wsxiaoys, line 1: Please create a new file `web_document.rs` for such changes.

ee/tabby-db/src/lib.rs
```diff
@@ -40,6 +40,7 @@ mod user_completions;
 mod user_events;
 mod users;
 mod web_crawler;
+mod web_document;
```
wsxiaoys, line 43:
```suggestion
mod web_documents;
```

ee/tabby-db/src/web_document.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub id: i64,
+    pub name: String,
+    pub url: String,
+    pub is_preset: bool,
+    pub created_at: DateTime<Utc>,
+    pub updated_at: DateTime<Utc>,
+}
+
+impl DbConn {
+    pub async fn list_web_document(
```
wsxiaoys, line 20:
```suggestion
    pub async fn list_web_documents(
```

wsxiaoys (changes requested):
Some minor nits, otherwise LGTM

ee/tabby-schema/src/schema/constants.rs
```diff
@@ -5,7 +5,8 @@ lazy_static! {
     pub static ref REPOSITORY_NAME_REGEX: Regex = Regex::new("^[a-zA-Z][\\w.-]+$").unwrap();
     pub static ref USERNAME_REGEX: Regex =
         Regex::new(r"^[^0-9±!@£$%^&*_+§¡€#¢¶•ªº«\\/<>?:;|=.,]{2,20}$").unwrap();
-    pub static ref WEB_DOCUMENT_NAME_REGEX: Regex = Regex::new(r"^[A-Za-z][A-Za-z0-9\ ]*$").unwrap();
+    pub static ref WEB_DOCUMENT_NAME_REGEX: Regex =
```
wsxiaoys, line 8: for a quick follow up, the space can only present individually, e.g `a b c` is ok, but `a b  c` is not allowed.

xxs-wallace: Done

PR #2916, opened 2024-08-20
chore(vscode): add relevant context menu in context menu

wsxiaoys:
image

cc @liangfung

### DIFF
--- a/clients/vscode/package.json
+++ b/clients/vscode/package.json
@@ -30,6 +30,9 @@
   "activationEvents": [
     "onStartupFinished"
   ],
+  "enabledApiProposals": [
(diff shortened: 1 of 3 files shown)


clients/vscode/package.json
```diff
@@ -30,6 +30,9 @@
   "activationEvents": [
     "onStartupFinished"
   ],
+  "enabledApiProposals": [
+    "contribEditorContentMenu"
+  ],
```
icycodes, line 35: It seems that this proposed api is not used.

wsxiaoys, line 35:
```suggestion
```

icycodes (approved):

PR #2918, opened 2024-08-20
chore(vscode): support add file context

wsxiaoys:
image

### DIFF
--- a/clients/vscode/package.json
+++ b/clients/vscode/package.json
@@ -206,7 +206,7 @@
       "editor/context": [
         {
           "submenu": "tabby.submenu",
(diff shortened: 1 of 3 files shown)


clients/vscode/package.json
```diff
@@ -194,7 +205,8 @@
       ],
       "editor/context": [
         {
-          "submenu": "tabby.submenu"
+          "submenu": "tabby.submenu",
+          "group": "2_workspace"
```
icycodes, line 209: I think we should use separate group
```suggestion
          "group": "2_tabby"
```

icycodes (approved):

PR #2919, opened 2024-08-20
refactor(crawler): add timeout for crawler job

xxs-wallace:
add timeout to prevent one job block the whole background loop

### DIFF
--- a/ee/tabby-webserver/src/service/background_job/web_crawler.rs
+++ b/ee/tabby-webserver/src/service/background_job/web_crawler.rs
@@ -1,12 +1,14 @@
-use std::sync::Arc;
-
 use chrono::Utc;
 use futures::StreamExt;
+use std::sync::Arc;
+use std::time::Duration;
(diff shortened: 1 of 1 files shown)


ee/tabby-webserver/src/service/background_job/web_crawler.rs
```diff
@@ -46,4 +48,21 @@ impl WebCrawlerJob {
...
         Ok(())
     }
+    pub async fn run(self, embedding: Arc<dyn Embedding>) -> tabby_schema::Result<()> {
+        let url = self.url.clone();
+        if tokio::time::timeout(
+            Duration::from_secs(CRAWLER_TIMEOUT_SECS),
+            self.run_impl(embedding),
+        )
+        .await
+        .is_err()
```
wsxiaoys, line 58: can this method being somehow tested? can you add a unit test to demonstrate the behavior? e.g a async function with async for loop, and can be halted by timeout.

PR #2948, opened 2024-08-22
fix(ui): adjust the layout of answer block

liangfung:


ee/tabby-ui/app/search/components/search.tsx
```diff
@@ -826,9 +829,9 @@ function AnswerBlock({
     onRegenerateResponse,
     onSubmitSearch,
     setDevPanelOpen,
-    setConversationIdForDev
+    setConversationIdForDev,
+    enableDeveloperMode
   } = useContext(SearchContext)
-  const [enableDeveloperMode] = useEnableDeveloperMode()
```
liangfung, line 831: When the key of the answerBlock changes, useEnableDeveloperMode re-executes with an initial value of false, and is subsequently set to true. This results in a layout change.

PR #2952, opened 2024-08-22
feat: smpt shpport self-signed cert

zwpaper:


zwpaper:
reading cert path from env PR is created #3035, this PR will remain open and I will update the db logic here after #3035 merged.

PR #2958, opened 2024-08-22
chore(vscode): optimize go to file experience for client side context

wsxiaoys:


clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -108,22 +108,30 @@ export class ChatViewProvider implements WebviewViewProvider {
...
-            });
-
-            // Move the cursor to the specified line
-            const start = new Position(Math.max(0, context.range.start - 1), 0);
-            const end = new Position(context.range.end, 0);
-            newEditor.selection = new Selection(start, end);
-            newEditor.revealRange(new Range(start, end), TextEditorRevealType.InCenter);
+          let document;
+          try {
+            document = await workspace.openTextDocument(context.filepath);
```
icycodes, line 113: It seems that this line handle none of all cases:
`context.filepath` could be:
1. relative path to git root
2. relative path to workspace
3. file uri with absolute path

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -508,3 +511,36 @@ function resolveFilePathAndGitUrl(uri: Uri, gitProvider: GitProvider): { filepat
...
+  }
+
+  if (!folders) {
+    return null;
+  }
+
+  for (const root of folders) {
+    const absoluteFilepath = Uri.joinPath(root.uri, filepath);
+    try {
+      return workspace.openTextDocument(absoluteFilepath);
```
icycodes, line 532:
```suggestion
      return await workspace.openTextDocument(absoluteFilepath);
```
await is required for catching errors.

icycodes (approved):

Issue #2962, opened 2024-08-23
[VSCode] support switching between backend servers


wsxiaoys:
**Please describe the feature you want**

This will help developers of tabby to switching from dev instance / demo instance / prod instance

**Additional context**

---
Please reply with a 👍 if you want this feature.

PR #2975, opened 2024-08-23
feat(vscode): add command to select server config from history

wsxiaoys:
fix https://github.com/TabbyML/tabby/issues/2962

cc @liangfung

### DIFF
--- a/clients/tabby-agent/src/TabbyAgent.ts
+++ b/clients/tabby-agent/src/TabbyAgent.ts
@@ -136,6 +136,7 @@ export class TabbyAgent extends EventEmitter implements Agent {
     }
 
     if (!this.api || isServerConnectionChanged) {
+      this.changeStatus("notInitialized");
       await this.setupApi();
     }
 
--- a/clients/vscode/package.json
+++ b/clients/vscode/package.json
@@ -133,6 +133,11 @@
         "command": "tabby.chat.edit.discard",
         "title": "Discard Changes",
         "category": "Tabby"
+      },
+      {
+        "command": "tabby.server.selectPastServerConfig",
+        "title": "Select Server Config from History",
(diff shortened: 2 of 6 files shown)


clients/vscode/src/CommandPalette.ts
```diff
@@ -119,7 +119,8 @@ export default class CommandPalette {
       }
       item.command = () => this.issues.showHelpMessage();
     } else if (agentStatus === "ready") {
-      item.label = "Ready";
+      item.label = `Ready`;
```
wsxiaoys, line 122:
```suggestion
      item.label = "Ready";
```

clients/vscode/package.json
```diff
@@ -133,6 +133,11 @@
         "command": "tabby.chat.edit.discard",
         "title": "Discard Changes",
         "category": "Tabby"
+      },
+      {
+        "command": "tabby.server.selectPastServerConfig",
+        "title": "Select Server Config from History",
```
wsxiaoys, line 139:
```suggestion
        "title": "Select Server Endpoint from History",
```

clients/tabby-agent/src/TabbyAgent.ts
```diff
@@ -136,6 +136,7 @@ export class TabbyAgent extends EventEmitter implements Agent {
     }
 
     if (!this.api || isServerConnectionChanged) {
+      this.changeStatus("notInitialized");
```
wsxiaoys, line 139: @icycodes this is required to didChangeStatus will be triggered.

Before this change, if user switching config A to config B, and each config are legit, there won't be another `ready` status being triggered due to event deduping check.

icycodes: LGTM

PR #2986, opened 2024-08-27
chore(webserver): new thread are by default ephemeral

wsxiaoys:
Only threads that are shared (share button clicked) will be persisted (is_ephemeral = false) in backend.


xxs-wallace (approved):
LGTM for backend

liangfung (approved):

PR #2988, opened 2024-08-27
fix(index): fix broken code incremental indexing

wsxiaoys:
(this issue is introduced in 0.16)


crates/tabby-index/src/code/index.rs
```diff
@@ -99,7 +99,9 @@ async fn add_changed_documents(
                 continue;
             };
 
-            if cloned_index.is_indexed(&key) {
+            let id = SourceCode::to_index_id(&repository.source_id, &key).id;
```
wsxiaoys, line 102: @xxs-wallace in last round of refactoring, id's definition is changed, but the `is_indexed` check is not updated simultaneously

xxs-wallace (approved):
LGTM

Issue #2997, opened 2024-08-27
Add an command line confirmation before running migrations


wsxiaoys:
**Please describe the feature you want**
Currently, Tabby executes database migrations without any confirmation. if we detect that the database is from an older version, we shall generate a command-line confirmation and prompt the user to manually input a specific term (e.g., 'UPGRADE') before running the migration.

**Additional context**
Code location: https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/1e3521d622b6ff6756c3d87be6264bed1670f389/ee/tabby-db/src/lib.rs#L159

Library: https://lib.rs/crates/cliclack

---
Please reply with a 👍 if you want this feature.

Sma1lboy:
I'll take this if nobody doing this right now :)

PR #3008, opened 2024-08-28
feat(tabby-db): adding cli confirm before doing migration

Sma1lboy:
This PR is addressing #2997 
- Adding cli confirm before doing sql migration

### DIFF
--- a/ee/tabby-db/src/lib.rs
+++ b/ee/tabby-db/src/lib.rs
@@ -4,6 +4,7 @@ use anyhow::anyhow;
 use cache::Cache;
 use cached::TimedSizedCache;
 use chrono::{DateTime, Utc};
+use cliclack::{input, outro};
 pub use email_setting::EmailSettingDAO;
 pub use integrations::IntegrationDAO;
 pub use invitations::InvitationDAO;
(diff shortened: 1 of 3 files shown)


ee/tabby-db/src/lib.rs
```diff
@@ -156,6 +157,15 @@ impl DbConn {
 
     /// Initialize database, create tables and insert first token if not exist
     async fn init_db(pool: SqlitePool) -> Result<Self> {
+        let user_input: String = input(
+            "The database is from an older version. Type 'confirm' to run database migrations:",
+        )
+        .interact()?;
```
wsxiaoys, line 163: we shall only prompt user input when the sqlx migration version in db doesn't match the embed migration's version - thus here we need to check the version before prompting users.

ee/tabby-db/src/lib.rs
```diff
@@ -155,16 +159,40 @@ impl DbConn {
...
+
+
+
+    /// Initialize database, create tables and insert first token if not exist
+    async fn init_db(pool: SqlitePool) -> Result<Self> {
+        
+        let local_migrations = sqlx::migrate!();
+        let mut applied_migrations = sqlx::query("SELECT version FROM _sqlx_migrations")
+        .fetch_all(&pool)
+        .await?.into_iter();
```
wsxiaoys, line 180: this seems a bit verbose to me - maybe try method like this one https://docs.rs/sqlx/latest/sqlx/struct.SqliteConnection.html#method.list_applied_migrations (as we have a connection to sqlite already)

PR #3009, opened 2024-08-28
chore(thread): adjust api interface to allow pass source_ids as query…

wsxiaoys:
… constraint.

To adapt 0.18's design of `@` based source selection.

### DIFF
--- a/ee/tabby-schema/src/schema/thread/inputs.rs
+++ b/ee/tabby-schema/src/schema/thread/inputs.rs
@@ -1,6 +1,12 @@
-use juniper::{GraphQLInputObject, ID};
+use crate::bail;
+use juniper::{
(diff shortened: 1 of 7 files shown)


ee/tabby-schema/src/schema/thread/inputs.rs
```diff
@@ -25,34 +31,40 @@ pub struct CreateThreadAndRunInput {
...
-    pub git_url: String,
     pub filepath: Option<String>,
     pub language: Option<String>,
     pub content: String,
+
+    /// git_url to be included in the code search.
+    pub git_url: Option<String>,
+
+    /// source_ids to be included in the code search.
+    pub source_id: Option<String>,
```
wsxiaoys, line 53: @liangfung only one source code repository is allowed to be selected.

PR #3020, opened 2024-08-29
fix(vscode): only display `Edit with Tabby` when chat is enabled

wsxiaoys:
cc @Sma1lboy


clients/vscode/src/CodeAction.ts
wsxiaoys, line 1: rename to `CodeAction` so it can fit for non inline edit code actions as well.

clients/vscode/src/CodeAction.ts
```diff
@@ -0,0 +1,39 @@
...
+    _context: CodeActionContext,
+    token: CancellationToken,
+  ): CodeAction[] | undefined {
+    if (token.isCancellationRequested) {
+      return;
+    }
+
+    if (!this.contextVariables.chatEnabled) {
+      return;
+    }
```
wsxiaoys, line 28: cc @Sma1lboy shall only display actions when chat is enabled.

icycodes (approved):

PR #3030, opened 2024-08-31
chore(webserver): update logging in code indexing and third-party integration

wsxiaoys:


ee/tabby-webserver/src/service/background_job/third_party_integration.rs
```diff
@@ -127,20 +127,23 @@ impl SchedulerGithubGitlabJob {
         stream! {
             let mut count = 0;
             let mut num_updated = 0;
+
```
wsxiaoys, line 130:
```suggestion
```

PR #3038, opened 2024-09-02
fix(answer):  collect_relevant_code should also rewrite source tags

wsxiaoys:


ee/tabby-webserver/src/service/answer.rs
```diff
@@ -186,7 +192,8 @@ impl AnswerService {
             .as_ref()
             .inspect(|x| x.override_params(&mut params));
 
-        match self.code.search_in_language(query, params.clone()).await {
+        debug!(?query, ?params, "searching code");
```
wsxiaoys, line 195:
```suggestion
```

PR #3035, opened 2024-09-01
feat(email): support self-signed smtp cert from env

zwpaper:
split from https://github.com/TabbyML/tabby/pull/2952, support self-signed cert for smtp by reading a cert path from `TABBY_EMAIL_CERT_PATH` env.

this only supports Linux, NOT macOS, because macOS has some more pedantic requirements for self-signed cert.
We may need to dive deeper into the self-signed cert part for macOS support. ref: https://github.com/sfackler/rust-native-tls/issues/210 

fix https://github.com/TabbyML/tabby/issues/2437

PR #3043, opened 2024-09-02
chore(doc): add self signed cert intro

zwpaper:
hold for https://github.com/TabbyML/tabby/pull/3035

### DIFF
--- a/website/docs/administration/smtp/index.md
+++ b/website/docs/administration/smtp/index.md
@@ -30,11 +30,6 @@ before running Tabby, and set its value to the certificate, for example, in bash
 export TABBY_WEBSERVER_EMAIL_CERT=$(cat ./self-signed.crt)
 ```
 
-Due to some limitations on macOS, the self-signed certificate is not working under macOS,
-you can add the certificate to the macOS `Keychain Access` and set it to `Always Trusted` if needed.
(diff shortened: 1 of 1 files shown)


website/docs/administration/smtp/index.md
```diff
@@ -20,10 +20,24 @@ You can configure the SMTP server settings in the **Mail Delivery** page.
...
+before running Tabby, and set its value to the certificate, for example, in bash:
+
+```bash
+export TABBY_WEBSERVER_EMAIL_CERT=$(cat ./self-signed.crt)
+```
+
+Due to some limitations on macOS, the self-signed certificate is not working under macOS,
+you can add the certificate to the macOS `Keychain Access` and set it to `Always Trusted` if needed.
+Please refer to [Apple Support](https://support.apple.com/en-hk/guide/keychain-access/kyca11871/mac)
+for more detail.
```
wsxiaoys, line 36:
```suggestion
```

Given that the primary use case of SMTP is for IT/infrastructure administrators (in a Linux environment), I believe it's safe to exclude this part from the scope of the documentation.

wsxiaoys (approved):
Otherwise LGTM

zwpaper:
fixed @wsxiaoys, PTAL, thanks

wsxiaoys:
Pending merge after feature release

PR #3057, opened 2024-09-03
chore(llama.cpp): disable avx2 compiling flags

wsxiaoys:


Gnomesenpai:
`config.define("GGML_AVX2", "OFF")` is missing ; at the end, it will fail to compile

Gnomesenpai:
@wsxiaoys did this ever get implimented?

PR #3058, opened 2024-09-04
refactor(tabby-crawler): enhance crawling options and limits

wsxiaoys:


crates/tabby-crawler/src/lib.rs
```diff
@@ -23,6 +23,10 @@ async fn crawl_url(
         .arg(format!("starts_with(endpoint, \"{prefix_url}\")"))
         .arg("-depth")
         .arg("9999")
+        .arg("--extension-filter").arg("js,css,png,jpg,jpeg")  // Exclude non html extensions
```
wsxiaoys, line 26:
```suggestion
        .arg("--extension-filter").arg("js,css,png,jpg,jpeg")  // Exclude non-html extensions
```

PR #3061, opened 2024-09-04
feat(crawler): add headless mode support based on environment variable

wsxiaoys:


crates/tabby-crawler/src/lib.rs
```diff
@@ -29,10 +38,13 @@ async fn crawl_url(
         .arg("fqdn") // Limit crawling scope to the same origin.
         .arg("-max-response-size")
         .arg("10485760") // 10MB max body size
+        .arg("-rate-limit-minute")
+        .arg("30")
```
wsxiaoys, line 42:
```suggestion
        .arg("120")
```

PR #3078, opened 2024-09-05
feat(ui): add ability to delete message pairs in Answer Engine

liangfung:
### Screen record
https://jam.dev/c/d16cfa4c-7f1e-463d-8b9d-6361004248d4


wsxiaoys:
To be safer, let's only show delete button when it's not streaming

PR #3083, opened 2024-09-05
refactor(hash-ids): fork hash-ids to return None instead of panic when ID is not valid

wsxiaoys:


crates/hash-ids/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "hash-ids"
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+homepage.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
wsxiaoys, line 9:
```suggestion
```

wsxiaoys, line 9:
```suggestion
```

Issue #3032, opened 2024-08-31
[VSCode] Switching Between Tabby Chat Panel and Active Editor Using Ctrl + L


Sma1lboy:
**Please describe the feature you want**

When in the Tabby chat panel, Ctrl + L can be used to return to the active editor. Similarly, when in the active editor, Ctrl + L can be used to focus on the chat panel. This functionality enables quick switching between the editor and the chat panel.

Currently, the Ctrl + L shortcut functions only one-way: it can move focus from the active editor to the chat panel but not back from the chat panel to the editor. This limitation affects smooth navigation between the editor and the chat panel.

---
Please reply with a 👍 if you want this feature.

Sma1lboy:
I'll take this one if nobody doing rn

wsxiaoys:
released in 0.12

PR #3100, opened 2024-09-07
feat(VSCode): trigger focus from the chat panel box to the active editor

Sma1lboy:
This PR is relate to #3032

What I fix in this PR
- [X]  focus shift from the chat panel box to the active window.
- [X] shifting into chat panel input box from chat panel by pressing focus command. 

Demo:

https://github.com/user-attachments/assets/b385666b-0d5f-4c68-b077-cece0d4285be

### DIFF
--- a/clients/tabby-chat-panel/src/index.ts
+++ b/clients/tabby-chat-panel/src/index.ts
@@ -22,7 +22,7 @@ export interface FetcherOptions {
 
 export interface InitRequest {
   fetcherOptions: FetcherOptions
-  focusKey: FocusKeybinding
+  focusKey?: FocusKeybinding
 }
 
 export interface FocusKeybinding {
--- a/clients/tabby-chat-panel/src/index.ts
+++ b/clients/tabby-chat-panel/src/index.ts
@@ -22,6 +22,15 @@ export interface FetcherOptions {
 
 export interface InitRequest {
   fetcherOptions: FetcherOptions
+  focusKey: FocusKeybinding
+}
+
+export interface FocusKeybinding {
+  key: string
+  ctrlKey: boolean
+  metaKey: boolean
+  shiftKey: boolean
+  altKey: boolean
 }
 
 export interface ErrorMessage {
@@ -55,6 +64,8 @@ export interface ClientApi {
 
   // On user copy content to clipboard.
   onCopy: (content: string) => void
+
+  focusOnEditor: () => void
 }
 
 export interface ChatMessage {
--- a/clients/vscode/src/chat/ChatViewProvider.ts
+++ b/clients/vscode/src/chat/ChatViewProvider.ts
@@ -34,6 +34,7 @@ export class ChatViewProvider implements WebviewViewProvider {
   private pendingMessages: ChatMessage[] = [];
   private pendingRelevantContexts: Context[] = [];
   private isChatPageDisplayed = false;
+  private isMac = process.platform === "darwin";
 
   constructor(
     private readonly context: ExtensionContext,
@@ -325,9 +326,9 @@ export class ChatViewProvider implements WebviewViewProvider {
       const focusKey = contributes.keybindings.find((cmd) => cmd.command === "tabby.chatView.focus");
       let focusKeybinding;
       if (focusKey) {
-        focusKeybinding = parseKeybinding(focusKey.key);
+        focusKeybinding = parseKeybinding(this.isMac && focusKey.mac ? focusKey.mac : focusKey.key);
       }
-      getLogger().info("keybinding: ", focusKeybinding);
+      getLogger().info("focus key binding: ", focusKeybinding);
       this.client?.init({
         fetcherOptions: {
           authorization: serverInfo.config.token,
--- a/clients/vscode/src/chat/ChatViewProvider.ts
+++ b/clients/vscode/src/chat/ChatViewProvider.ts
@@ -338,8 +338,8 @@ export class ChatViewProvider implements WebviewViewProvider {
           : {
               key: "l",
               altKey: false,
-              metaKey: true,
-              ctrlKey: true,
+              metaKey: isMac,
+              ctrlKey: !isMac,
               shiftKey: false,
             },
       });
--- a/clients/vscode/src/chat/ChatViewProvider.ts
+++ b/clients/vscode/src/chat/ChatViewProvider.ts
@@ -34,7 +34,8 @@ export class ChatViewProvider implements WebviewViewProvider {
   private pendingMessages: ChatMessage[] = [];
   private pendingRelevantContexts: Context[] = [];
   private isChatPageDisplayed = false;
-  private isUnix = Boolean(env.appHost === "desktop" && env.appRoot && env.appRoot[10] === "/");
+  // FIXME: this check is not compatible with the environment of a browser in macOS
+  private isMac = Boolean(env.appHost === "desktop" && process.platform === "darwin");
 
   constructor(
     private readonly context: ExtensionContext,
--- a/clients/vscode/src/chat/ChatViewProvider.ts
+++ b/clients/vscode/src/chat/ChatViewProvider.ts
@@ -341,7 +341,7 @@ export class ChatViewProvider implements WebviewViewProvider {
     const focusCommand = "tabby.chatView.focus";
     const defaultFocusKey = contributes.keybindings.find((cmd) => cmd.command === focusCommand);
     const defaultKeybinding = defaultFocusKey
-      ? parseKeybinding(this.isUnix && defaultFocusKey.mac ? defaultFocusKey.mac : defaultFocusKey.key)
+      ? parseKeybinding(this.isMac && defaultFocusKey.mac ? defaultFocusKey.mac : defaultFocusKey.key)
       : undefined;
 
     const allKeybindings = await readUserKeybindingsConfig();
--- a/clients/vscode/src/chat/ChatViewProvider.ts
+++ b/clients/vscode/src/chat/ChatViewProvider.ts
@@ -25,7 +25,9 @@ import type { ServerInfo } from "tabby-agent";
 import type { AgentFeature as Agent } from "../lsp/AgentFeature";
 import { createClient } from "./chatPanel";
 import { GitProvider } from "../git/GitProvider";
-
+import { getLogger } from "../logger";
+import { contributes } from "../../package.json";
+import { parseKeybinding } from "../util/KeybindingParser";
 export class ChatViewProvider implements WebviewViewProvider {
   webview?: WebviewView;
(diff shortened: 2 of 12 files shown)


Sma1lboy:
I will do another PR for
- [ ] shifting into chat panel input box from chat panel by pressing focus command

ee/tabby-ui/app/chat/page.tsx
```diff
@@ -63,29 +64,7 @@ export default function ChatPage() {
...
-        document.execCommand('selectAll')
-      }
-    }
-
-    window.addEventListener('keydown', onKeyDown)
-    return () => {
-      window.removeEventListener('keydown', onKeyDown)
-    }
-  }, [])
+  const [focusKeybinding, setFocusKeyBinding] = useState<FocusKeybinding>()
```
wsxiaoys, line 67:
```suggestion
  const focusKeybinding = useRef<FocusKeybinding>()
```

ee/tabby-ui/app/chat/page.tsx
```diff
@@ -113,6 +92,7 @@ export default function ChatPage() {
       setActiveChatId(nanoid())
       setIsInit(true)
       setFetcherOptions(request.fetcherOptions)
+      setFocusKeyBinding(request.focusKey)
```
wsxiaoys, line 95:
```suggestion
      focusKeyBinding.current = request.focusKey
```

ee/tabby-ui/app/chat/page.tsx
```diff
@@ -148,6 +128,38 @@ export default function ChatPage() {
...
+      ) {
+        server?.focusOnEditor()
+      }
+    }
+
+    window.addEventListener('keydown', onKeyDown)
+    return () => {
+      window.removeEventListener('keydown', onKeyDown)
+    }
+  }, [focusKeybinding, server, client])
```
wsxiaoys, line 161:
```suggestion
  }, [server, client])
```

ee/tabby-ui/app/chat/page.tsx
wsxiaoys: unlike fetcherOptions, no need to trigger re-rendering for `focusKeyBinding`

ee/tabby-ui/app/chat/page.tsx
```diff
@@ -113,6 +92,7 @@ export default function ChatPage() {
       setActiveChatId(nanoid())
       setIsInit(true)
       setFetcherOptions(request.fetcherOptions)
```
wsxiaoys, line 94: might also switch to ref as there's no need to do reactive rendering.

wsxiaoys (changes requested):
Otherwise LGTM

wsxiaoys (changes requested):

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -313,10 +323,25 @@ export class ChatViewProvider implements WebviewViewProvider {
...
         fetcherOptions: {
           authorization: serverInfo.config.token,
         },
+        focusKey: focusKeybinding
+          ? focusKeybinding
+          : {
+              key: "l",
+              altKey: false,
+              metaKey: true,
+              ctrlKey: true,
```
wsxiaoys, line 342:
```suggestion
              metaKey: isMac,
              ctrlKey: !isMac,
```

wsxiaoys (approved):
Otherwise LGTM

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -313,10 +323,25 @@ export class ChatViewProvider implements WebviewViewProvider {
     if (serverInfo.config.token) {
       this.client?.cleanError();
       // Duplicate init won't break or reload the current chat page
+      const focusKey = contributes.keybindings.find((cmd) => cmd.command === "tabby.chatView.focus");
+      let focusKeybinding;
+      if (focusKey) {
+        focusKeybinding = parseKeybinding(this.isMac && focusKey.mac ? focusKey.mac : focusKey.key);
```
wsxiaoys, line 329:
```suggestion
        focusKeybinding = parseKeybinding(this.isMac? focusKey.mac : focusKey.key);
```

Sma1lboy: focusKey.mac return => string | undefined

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -25,13 +25,16 @@ import type { ServerInfo } from "tabby-agent";
...
+import { getLogger } from "../logger";
+import { contributes } from "../../package.json";
+import { parseKeybinding } from "../util/KeybindingParser";
 export class ChatViewProvider implements WebviewViewProvider {
   webview?: WebviewView;
   client?: ServerApi;
   private pendingMessages: ChatMessage[] = [];
   private pendingRelevantContexts: Context[] = [];
   private isChatPageDisplayed = false;
+  private isMac = process.platform === "darwin";
```
icycodes, line 37: Do not use Node.js api `process` directly, it will be replaced with a polyfill implementation when the code is built for the browser.

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -313,10 +323,25 @@ export class ChatViewProvider implements WebviewViewProvider {
     if (serverInfo.config.token) {
       this.client?.cleanError();
       // Duplicate init won't break or reload the current chat page
+      const focusKey = contributes.keybindings.find((cmd) => cmd.command === "tabby.chatView.focus");
```
icycodes, line 326: The `contributes` is import from `package.json` at build time, so it is always `ctrl/cmd+l`. So if we are aiming to support the user-customized keybinding, it is required to read the vscode user keybinding config file.

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -313,10 +323,25 @@ export class ChatViewProvider implements WebviewViewProvider {
...
         },
+        focusKey: focusKeybinding
+          ? focusKeybinding
+          : {
+              key: "l",
+              altKey: false,
+              metaKey: this.isMac,
+              ctrlKey: !this.isMac,
+              shiftKey: false,
+            },
```
icycodes, line 344: Therefore, the default value here can be removed.

clients/tabby-chat-panel/src/index.ts
```diff
@@ -22,6 +22,15 @@ export interface FetcherOptions {
...
   fetcherOptions: FetcherOptions
+  focusKey?: FocusKeybinding
+}
+
+export interface FocusKeybinding {
+  key: string
+  ctrlKey: boolean
+  metaKey: boolean
+  shiftKey: boolean
+  altKey: boolean
```
icycodes, line 33: Can we bubble all uncosumed key events to the VS Code webview and process them there? This way, we wouldn't have to pass the keybinding configurations to the chat panel implementation.   
It would also help clean up the handling of copy/paste key events, which are only necessary in the VS Code environment. Additionally, if we need to support more keybindings, the API can remain stable. For example, if we want to add compatibility to use `ctrl+shift+p` to invoke the command palette when the chat panel is in focus.

wsxiaoys: When passing additional data, simply bubbling keyboard events is insufficient; it necessitates an extra server API call. (keyboard event -&gt; vscode call `ServerAPI.getData`, which seems cumbersome).

Conversely, handling events within the webview allows proactive data passing. (No data is passed in `ClientAPI.focusEditor`, but it could be similar.)

icycodes: Keyboard events should never contain additional data.  
If you are referring to the `onCopy` api requiring additional data, it is used when clicking the tool icon on the code block; it is different from pressing `ctrl/cmd + c`.

icycodes: offline resolved

clients/vscode/src/chat/ChatViewProvider.ts
Sma1lboy: how about using  env.appRoot && env.appRoot[0] !== '/' to check isWin or not

clients/vscode/src/chat/ChatViewProvider.ts
Sma1lboy: I'll fix it rn

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -18,20 +18,23 @@ import {
...
+import { getLogger } from "../logger";
+import { contributes } from "../../package.json";
+import { parseKeybinding, readUserKeybindingsConfig } from "../util/KeybindingParser";
 export class ChatViewProvider implements WebviewViewProvider {
   webview?: WebviewView;
   client?: ServerApi;
   private pendingMessages: ChatMessage[] = [];
   private pendingRelevantContexts: Context[] = [];
   private isChatPageDisplayed = false;
+  private isUnix = Boolean(env.appHost === "desktop" && env.appRoot && env.appRoot[10] === "/");
```
icycodes, line 37:
```suggestion
  // FIXME: this check is not compatible with the environment of a browser in macOS
  private isMac = Boolean(env.appHost === "desktop" && process.platform === "darwin");
```

I do not think this makes sense.   
Using `isMac = process.platform == 'darwin'` works on desktop Node.js, but not on a browser, as it always returns false on a browser, even on a browser in macOS.   
The updated implementation with `env.appRoot` also always returns false on a browser.  
So we need another way to check if it is macOS on browser, for example use the browser UserAgent.   

It looks good to me that not support the isMac check on browser for now. But please leave a fixme comment stating that this is not compatible with a browser environment yet.

icycodes (approved):
Otherwise LGTM

clients/vscode/src/chat/ChatViewProvider.ts
```diff
@@ -18,20 +18,24 @@ import {
...
+import { contributes } from "../../package.json";
+import { parseKeybinding, readUserKeybindingsConfig } from "../util/KeybindingParser";
 export class ChatViewProvider implements WebviewViewProvider {
   webview?: WebviewView;
   client?: ServerApi;
   private pendingMessages: ChatMessage[] = [];
   private pendingRelevantContexts: Context[] = [];
   private isChatPageDisplayed = false;
+  // FIXME: this check is not compatible with the environment of a browser in macOS
+  private isMac = Boolean(env.appHost === "desktop" && process.platform === "darwin");
```
wsxiaoys, line 38:
```suggestion
  private isMac: bool = env.appHost === "desktop" && process.platform === "darwin";
```

Issue #3120, opened 2024-09-11
feat(makefile): adding dev command on makefile


Sma1lboy:
If we want to develop the backend of the Tabby web UI separately, it is relatively troublesome; we need to enter three commands according to the README. We can consolidate these commands here.

---
Please reply with a 👍 if you want this feature.

PR #3121, opened 2024-09-11
chore(makefile): adding dev target for makefile

Sma1lboy:
#3120

- Add a dev target to avoid prompting for three separate commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,10 +20,7 @@ update-db-schema:
 caddy:
 	caddy run --watch --config ee/tabby-webserver/development/Caddyfile
 
-MODEL ?= ""
-CHAT_MODEL ?= ""
-
-dev:
+ui-dev:
 	@echo "Starting development environment..."
 	@bash -c '\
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,6 +20,27 @@ update-db-schema:
 caddy:
 	caddy run --watch --config ee/tabby-webserver/development/Caddyfile
 
+MODEL ?= ""
+CHAT_MODEL ?= ""
+
+dev:
```
wsxiaoys, line 26:
```suggestion
ui-dev:
```

Makefile
```diff
@@ -20,6 +20,27 @@ update-db-schema:
 caddy:
 	caddy run --watch --config ee/tabby-webserver/development/Caddyfile
 
+MODEL ?= ""
+CHAT_MODEL ?= ""
```
wsxiaoys, line 24: prefer setting them inside of `~/.tabby/config.toml` instead of in every invoking

PR #3128, opened 2024-09-11
refactor(context): extract ContextSourceKind as an graphql interface

wsxiaoys:


ee/tabby-schema/graphql/schema.graphql
wsxiaoys, line 1: @liangfung sourceId has been added to all kinds of context sources.

Issue #2894, opened 2024-08-15
Integration with Notion Database


wsxiaoys:

umialpha:
I am interested in this feature and will work on it :)

umialpha:
@wsxiaoys  i have read the notion api. Notion has basic elements like `Database`, `Page`, `Block`. It does have knowledge about issues. They are just some pages to notion. 

So, i think we need another type of integration named `notion` besides `github/gitlab`, `webdocument`.
it has the following property.
```
name string
database_id string
access_token string
```

Alternately,  we can treat `Notion` and `WebDocument` the same, just different kind of documents with different api(notion api  vs webcrawler). If so, the table `web_documents` should add more columes.

I prefer adding a notion integration type. What do u suggest?

wsxiaoys:
Hi, a common practice in Notion is using a database to store data. The linked template in the thread is an example. For the initial PR, it's fine to use fixed column names (title/summary) and import them (similar to syncing GitHub issues). 

We can iterate on this in future steps, such as allowing the front end to decide which columns should be indexed.

PR #3142, opened 2024-09-13
notion integration[WIP]

umialpha:
#2894 

## implements

I followed the pattern `WebCrawler`, creating the following core components:
- Adding list/create/delete interfaces to `Query`and `Mutation`.
- Adding new interface `NotionDocumentService` and impl `NotionServiceImpl`
- Adding list/create/delete interfaces to DBCon,
- Adding a new enum  BackgroundJobEvent.NotionIntregration
- Adding `notion_utils` to fetch docs using notion apis.

## WIP
I've done testing the logic by using postman's graphql with my own notion access token and database id. It passed.
However, the whole code lacks unit tests. I will denitely add them.

PTAL while i am adding them.

## Next PR
1.  Frontend: i am not familier with Typescript. If no one has time to implement frontend, i will try my best to implement it, including "listing board", "creating board" and "answer board"
2.  (Possibly required) Merge all kinds of integration jobs into a unified service to avoid the need to add a variable to the web server each time a new service is added.
2. (Possibly required) Currently, background tasks are executed serially; it may be necessary to change to multi-threaded execution.


umialpha:
@wsxiaoys ,PTAL, I've finished both the fe and be parts.

PR #3196, opened 2024-09-25
feat(core): support customize presence_penalty for answer engine in config.toml

eryue0220:
This pr is to support presence_penalty through `config.toml`, such as the following in the configuration file:

```toml
# config.toml
[answer]
presence_penalty = 0.1
```


ee/tabby-webserver/src/service/answer.rs
```diff
@@ -492,6 +490,17 @@ Remember, don't blindly repeat the contexts verbatim. When possible, give code s
     )
 }
 
+// https://platform.openai.com/docs/api-reference/chat/create#chat-create-presence_penalty
```
wsxiaoys, line 493: this won't be very useful - since the backend implementation might be another api (e.g codestral, deepseek, ...).

I recommend just pass this to backend model as is.

wsxiaoys (commented):
Otherwise LGTM

Issue #2964, opened 2024-08-23
[VSCode] Allow attaching file context to inline editor


wsxiaoys:
**Please describe the feature you want**
![image](https://github.com/user-attachments/assets/e5bc200c-1aea-49b6-a744-cc1d7136505f)

Support attaching file context to inline edit quick action list using similar UX

**Additional context**
File picker (built with QuickPick): https://github.com/leizongmin/vscode-quick-open/tree/master

---
Please reply with a 👍 if you want this feature.

eryue0220:
If no one take this, I could try it.

wsxiaoys:
Implemented in https://github.com/TabbyML/tabby/pull/3766

PR #3231, opened 2024-10-02
feat(vscode): support attach files in inline chat

eryue0220:
Fix: #2964 

https://github.com/user-attachments/assets/8e89570a-2d58-41e0-b00c-e0d726f03af3


wsxiaoys:
Note #2964 actually pertains to adding file context to the "inline editor" functionality (cmd-i "Tabby: Start Inline Editing"). However, the interaction appears to be great!

Could you modify the functionality to support inline editing? We don't really need it for the side panel since we can add file context by opening the file directly.

eryue0220:
> Note #2964 actually pertains to adding file context to the "inline editor" functionality (cmd-i "Tabby: Start Inline Editing"). However, the interaction appears to be great!
> 
> Could you modify the functionality to support inline editing? We don't really need it for the side panel since we can add file context by opening the file directly.

Oh sorry, I missed that, I'll fix it later.

eryue0220:
Updated:


https://github.com/user-attachments/assets/1f633c99-5084-4d11-9e46-29ec65b1f6af

PR #3244, opened 2024-10-05
docs(vscode): Add VSCode/VSCodium installation instructions and reference to Open VSX

ntninja:
Makes it obvious that VSCodium is also supported through Open VSX.


icycodes:
Hi, @ntninja 
Thank you for this PR.  
I will merge it and also make some modifications in the follow-up PR.

PR #3253, opened 2024-10-08
feat(config): add additional language configuration support

wsxiaoys:


Sma1lboy (commented):
Acceptable, but it's getting messy if we only put a vec on the root config.
It will look something like this, which is hard to read:
```toml
additional_languages = [
  { languages = ["CustomLang"], exts = [".cl", ".custom"], top_level_keywords = ["function", "class", "module"], line_comment = "//", chunk_size = 1000 },
  { languages = ["AnotherLang"], exts = [".al"], line_comment = "#" }
]
(...other content)
[model.chat.http]
kind=...
```
Let's change it to this:
```toml
[[additional_languages]]
languages = ["CustomLang"]
exts = [".cl", ".custom"]
top_level_keywords = ["function", "class", "module"]
line_comment = "//"
chunk_size = 1000

[[additional_languages]]
languages = ["AnotherLang"]
exts = [".al"]
line_comment = "#"
```
This is much better for reading.

crates/tabby-common/src/config.rs
```diff
@@ -29,6 +30,9 @@ pub struct Config {
 
     #[serde(default)]
     pub answer: AnswerConfig,
+
+    #[serde(default)]
+    pub additional_languages: Vec<languages::Language>,
```
Sma1lboy, line 35:
```suggestion
    pub additional_languages: AdditionalLanguages,
```

crates/tabby-common/src/config.rs
```diff
@@ -29,6 +30,9 @@ pub struct Config {
...
 }
```
Sma1lboy, line 37:
```suggestion
#[derive(Serialize, Deserialize, Debug, Clone, Default)]
#[serde(transparent)]
pub struct AdditionalLanguages(pub Vec<languages::Language>);
impl AdditionalLanguages {
    pub fn lang(&self) -> &Vec<languages::Language> {
        &self.0
    }

    pub fn lang_mut(&mut self) -> &mut Vec<languages::Language> {
        &mut self.0
    }
}
```

just for example

Sma1lboy (changes requested):
Otherwise, LGTM

Sma1lboy (approved):

PR #3258, opened 2024-10-10
feat(download): allow fetching model files with multiple partitions

zwpaper:
this PR is only for reviewing the changes onto https://github.com/TabbyML/tabby/pull/3216, I have also raise a PR for #3216: https://github.com/umialpha/tabby/pull/1

I have fork the registry and add Qwen2.5-Coder, which has a 3 files model, could be test by:

```sh
tabby serve --model zwpaper/Qwen2.5-Coder-7B-Instruct --device metal --chat-model zwpaper/Qwen2.5-Coder-7B-Instruct
```


crates/tabby-download/src/lib.rs
```diff
@@ -41,49 +42,98 @@ async fn download_model_impl(
...
+        let combined_error = errors
+            .into_iter()
+            .fold(anyhow::anyhow!("Multiple errors occurred"), |acc, err| {
+                acc.context(err)
+            });
+        Err(combined_error)
+    }
+}
 
+async fn tryget_download_filename(url: &str) -> Result<String> {
```
wsxiaoys, line 100: since we’re storing model in format of `model-000xxx-of-000yyy.gguf`, do we still need this function to extract file name?

crates/tabby-common/src/registry.rs
```diff
@@ -54,38 +108,100 @@ pub struct ModelRegistry {
...
+//
+// fn get_model_root_dir(model_name) -> {root}/{model_name}
+//
+// fn get_model_dir(model_name) -> {root}/{model_name}/ggml
+//
+// fn get_model_path(model_name)
+//   for single model file
+//     -> {root}/{model_name}/ggml/model.gguf
+//   for multiple model files
+//     -> {root}/{model_name}/ggml/{entrypoint}
```
wsxiaoys, line 130: make the single / multiple model file name consistent by using `model-000xx-of-000yy` naming routine.

crates/tabby-common/src/env.rs
```diff
@@ -0,0 +1,12 @@
+pub fn get_download_host() -> String {
+    std::env::var("TABBY_DOWNLOAD_HOST").unwrap_or_else(|_| "huggingface.co".to_string())
+}
+
+pub fn get_huggingface_mirror_host() -> Option<String> {
+    std::env::var("TABBY_HUGGINGFACE_HOST_OVERRIDE").ok()
+}
+
```
wsxiaoys, line 8: should be in tabby-download instead of tabby-common

crates/tabby-common/src/env.rs
```diff
@@ -0,0 +1,12 @@
...
+// for debug only
+pub fn use_local_model_json() -> bool {
+    std::env::var("TABBY_USE_LOCAL_MODEL_JSON").is_ok()
+}
```
wsxiaoys, line 12: remove

crates/tabby-common/src/registry.rs
```diff
@@ -54,38 +108,100 @@ pub struct ModelRegistry {
     pub models: Vec<ModelInfo>,
 }
 
+lazy_static! {
+    pub static ref LEGACY_GGML_MODEL_RELATIVE_PATH: String =
+        format!("ggml{}q8_0.v2.gguf", std::path::MAIN_SEPARATOR_STR);
```
wsxiaoys, line 113: we can drop the migration logic for q8_0.v2 since it’s several release ago.

crates/tabby-common/src/registry.rs
zwpaper: the comment is inconsistent with the code, deleted those lines.

crates/tabby-common/src/registry.rs
```diff
@@ -94,6 +138,23 @@ impl ModelRegistry {
...
 
+    pub fn migrate_model_path(
+        &self,
+        name: &str,
+        old_model_path: &PathBuf,
+    ) -> Result<(), std::io::Error> {
+        // legacy model always has a single file
+        let model_path = self
+            .get_model_store_dir(name)
+            .join("model-00001-of-00001.gguf");
```
wsxiaoys, line 149:
```suggestion
            .join("model-00000-of-00001.gguf");
```

shard file shall be 0 based

crates/tabby-common/src/registry.rs
```diff
@@ -94,6 +138,23 @@ impl ModelRegistry {
...
+        std::fs::rename(&old_model_path, &model_path)?;
+        #[cfg(target_family = "unix")]
+        std::os::unix::fs::symlink(&model_path, &old_model_path)?;
+        #[cfg(target_family = "windows")]
+        std::os::windows::fs::symlink_file(&model_path, &old_model_path)?;
```
wsxiaoys, line 154: just move the file on all platform

crates/tabby-common/src/env.rs
zwpaper: we can use a forked registry to do the validation, dropped this change

crates/tabby-download/src/lib.rs
```diff
@@ -41,49 +90,70 @@ async fn download_model_impl(
...
+    // prepare for download
+    let dir = registry.get_model_store_dir(name);
     fs::create_dir_all(dir)?;
 
-    let filename = path
-        .to_str()
-        .ok_or_else(|| anyhow!("Could not convert filename to UTF-8"))?;
-    let intermediate_filename = filename.to_owned() + ".tmp";
+    let mut download_tasks = vec![];
+    for (index, url) in urls.iter().enumerate() {
```
umialpha, line 115: The order of the partitioned urls matters. if we name models ourselves, we must sort the URLs first to prevent users from filling in the wrong order.

zwpaper: users may provide some malformed URLs, I would prefer that we reserve the order and require all provided URLs in a expected order

crates/tabby-download/Cargo.toml
wsxiaoys, line 1: need rebase

crates/tabby-download/src/lib.rs
```diff
@@ -1,24 +1,67 @@
...
-    }
+pub fn get_download_host() -> String {
+    std::env::var("TABBY_DOWNLOAD_HOST").unwrap_or_else(|_| "huggingface.co".to_string())
+}
+
+pub fn get_huggingface_mirror_host() -> Option<String> {
+    std::env::var("TABBY_HUGGINGFACE_HOST_OVERRIDE").ok()
+}
+
+pub fn filter_download_address(model_info: &ModelInfo) -> Vec<(String, String)> {
```
wsxiaoys, line 26: please add unit test

crates/tabby-download/src/lib.rs
```diff
@@ -29,9 +72,10 @@ async fn download_model_impl(
     let model_info = registry.get_model_info(name);
     registry.save_model_info(name);
 
-    registry.migrate_model_path(name)?;
-    let model_path = registry.get_model_path(name);
-    if model_path.exists() {
+    registry.migrate_relative_model_path(name)?;
+
+    let model_path = registry.get_model_entry_path(name);
+    if let Some(model_path) = model_path {
```
wsxiaoys, line 78: this implementation only checks sha256 for the first partition, which seems wrong to me.

crates/aim-downloader/src/bar.rs
wsxiaoys, line 1: since the change is no longer required, please revert or leave this into individual PR

crates/tabby-download/src/lib.rs
```diff
@@ -41,49 +85,62 @@ async fn download_model_impl(
                 "Checksum doesn't match for <{}/{}>, re-downloading...",
                 registry.name, name
             );
-            fs::remove_file(&model_path)?;
+
+            fs::remove_dir_all(registry.get_model_store_dir(name))?;
```
wsxiaoys, line 89: this implementation is wrong, we need to delete the entire model dir, not just `.../ggml`.

wsxiaoys (changes requested):

crates/tabby-common/src/registry.rs
```diff
@@ -54,6 +71,18 @@ pub struct ModelRegistry {
     pub models: Vec<ModelInfo>,
 }
 
+lazy_static! {
+    pub static ref GGML_MODEL_RELATIVE_PATH: String =
+        format!("ggml{}model.gguf", std::path::MAIN_SEPARATOR_STR);
```
wsxiaoys, line 76:
```suggestion
    pub static ref LEGACY_GGML_MODEL_PATH: String =
        format!("ggml{}model.gguf", std::path::MAIN_SEPARATOR_STR);
```

Please also update related function names, e.g migrate_relative_model_path

crates/tabby-common/src/registry.rs
```diff
@@ -69,29 +98,61 @@ impl ModelRegistry {
...
+    // get_model_dir returns {root}/{name}, e.g. ~/.tabby/models/TabbyML/StarCoder-1B
+    pub fn get_model_dir(&self, name: &str) -> PathBuf {
         models_dir().join(&self.name).join(name)
     }
 
-    pub fn migrate_model_path(&self, name: &str) -> Result<(), std::io::Error> {
-        let model_path = self.get_model_path(name);
+    // get_model_path returns the entrypoint of the model,
+    // will look for the file with the prefix "00001-of-"
+    pub fn get_model_entry_path(&self, name: &str) -> Option<PathBuf> {
```
wsxiaoys, line 113: does this function need to be public? please also check newly added `pub` function in this struct.

crates/tabby-download/Cargo.toml
```diff
@@ -9,3 +9,6 @@ tabby-common = { path = "../tabby-common" }
 anyhow = { workspace = true }
 tracing = { workspace = true }
 tokio-retry = "0.3.0"
+
+[dev-dependencies]
+serial_test = "3"
```
wsxiaoys, line 14:
```suggestion
serial_test.workspace = true
```

wsxiaoys (approved):
Otherwise LGTM

crates/tabby-common/src/registry.rs
zwpaper: this is also used in llama-cpp-server 
https://github.com/TabbyML/tabby/pull/3258/files#diff-7e6c7f751bfa100a469d7b7e1b68d567287a88df6425e2c8155333a229ecce59R282

PR #3281, opened 2024-10-17
fix(agent): fix errors when endpoint config ends with trailing slash.

icycodes:


clients/tabby-agent/src/config/index.ts
```diff
@@ -60,6 +60,12 @@ function mergeConfig(
     serverProvidedConfig.anonymousUsageTracking = { disable: true };
   }
   const merged = mergeFunction(base, configFileConfig, clientProvidedConfig, serverProvidedConfig) as ConfigData;
+
+  // remove trailing slash from endpoint
+  if (merged.server.endpoint && merged.server.endpoint.endsWith("/")) {
```
wsxiaoys, line 65: what if there's re multiple trailing `/`, shall we just `trimRight(x, "\n")`?

icycodes: updated

Issue #3279, opened 2024-10-16
Completion results are worse with Windows line endings (CRLF)


kleincode:
## Describe the bug
Tabby is running on a Linux machine, but I'm using it through the VSCode extension on Windows. When I edit a document with CRLF line endings (the standard on Windows), code completion returns an empty string or unfitting suggestions substantially more often. **When I change the line endings to LF, the completion suggestions improve.**

From looking at the completions events, I can see that **the line endings are fed directly to the model** - in the `prompt` field, they are conserved. Thus, the same piece of code will lead to different results by default on Windows and Linux/Mac machines. I have tested this with different variants of **CodeLLaMa**.

I'd like to ask if this is intended. I would suggest that only LF line endings should be fed to the model because in my experiments, the results were consistently better than with CRLF. Theoretically, this makes sense since the models are probably trained mostly on LF data.

## Information about your version
tabby 0.18.0 via Docker compose on Debian with CUDA

## Information about your GPU
```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 555.42.02              Driver Version: 555.42.02      CUDA Version: 12.5     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA A10                     On  |   00000000:B3:00.0 Off |                    0 |
|  0%   65C    P0             71W /  150W |    8653MiB /  23028MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
                                                                                         
+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A    975463      C   /usr/bin/python3                             1394MiB |
|    0   N/A  N/A    976533      C   /opt/tabby/bin/llama-server                   786MiB |
|    0   N/A  N/A    980188      C   /opt/tabby/bin/llama-server                  6454MiB |
+-----------------------------------------------------------------------------------------+
```

### Minimal example
Sending the Swagger example to the `completions` endpoint works perfectly fine:
```json
{
[...]

PR #3303, opened 2024-10-22
chore(completion): replace CRLF with LF for code completion LLM requests

zwpaper:
fix https://github.com/TabbyML/tabby/issues/3279

## Test

test model: CodeLlama-7B

Tabby command:

```sh
tabby serve --model CodeLlama-7B --device metal --chat-model zwpaper/Qwen2.5-Coder-7B-Instruct
```

test command:

```sh
curl -X 'POST' \
    'http://localhost:8080/v1/completions' \
    -H 'accept: application/json' \
    -H 'Content-Type: application/json' \
    -H 'Authorization: Bearer TOKEN' \
    -d '{
        "language": "java",
        "segments": {
          "prefix": "class MathUtils {\r\n    static double clamp(double value, double min, double max) {\r\n        return math.max(min, math.min(max, value));\r\n    }\r\n\r\n    static double lerp(double a, double b, double t) {\r\n        return a + (b - a) * t;\r\n    }\r\n    \r\n    static int fibonacci(int n) {\r\n        ",
          "suffix": "\r\n    }\r\n}"
        }
     }'
```

## Before
CleanShot 2024-10-22 at 18 31 25@2x

## After
CleanShot 2024-10-23 at 15 31 00@2x

### DIFF
--- a/crates/tabby/src/routes/completions.rs
+++ b/crates/tabby/src/routes/completions.rs
@@ -36,38 +36,11 @@ pub async fn completions(
 
     let user_agent = user_agent.map(|x| x.0.to_string());
 
-    let mut use_crlf = false;
-    if let Some(segments) = request.segments {
-        let mut new_segments = segments.clone();
-        if segments.prefix.contains("\r\n") {
-            use_crlf = true;
-            new_segments.prefix = segments.prefix.replace("\r\n", "\n");
-        }
-        if let Some(suffix) = segments.suffix {
-            if suffix.contains("\r\n") {
-                use_crlf = true;
-                new_segments.suffix = Some(suffix.replace("\r\n", "\n"));
-            }
-        }
-        request.segments = Some(new_segments);
-    }
-
     match state
         .generate(&request, &allowed_code_repository, user_agent.as_deref())
         .await
     {
-        Ok(resp) => {
-            if use_crlf {
-                let mut response_crlf = resp.clone();
-                for (index, choice) in resp.choices.iter().enumerate() {
-                    response_crlf.choices[index].text = choice.text.replace("\n", "\r\n");
-                }
-
-                return Ok(Json(response_crlf));
-            }
-
-            Ok(Json(resp))
-        }
+        Ok(resp) => Ok(Json(resp)),
         Err(err) => {
             warn!("{}", err);
             Err(StatusCode::BAD_REQUEST)
--- a/crates/tabby/src/routes/completions.rs
+++ b/crates/tabby/src/routes/completions.rs
@@ -1,9 +1,11 @@
 use std::sync::Arc;
 
-use axum::{extract::State, Extension, Json};
+use async_openai::types::Choice;
+use axum::{extract::State, http::request, Extension, Json};
 use axum_extra::{headers, TypedHeader};
 use hyper::StatusCode;
 use tabby_common::axum::{AllowedCodeRepository, MaybeUser};
+use tantivy::Segment;
 use tracing::{instrument, warn};
 
 use crate::services::completion::{CompletionRequest, CompletionResponse, CompletionService};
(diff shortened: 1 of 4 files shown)


crates/tabby/src/routes/completions.rs
```diff
@@ -36,11 +36,38 @@ pub async fn completions(
...
+            if suffix.contains("\r\n") {
+                use_crlf = true;
+                new_segments.suffix = Some(suffix.replace("\r\n", "\n"));
+            }
+        }
+        request.segments = Some(new_segments);
+    }
+
     match state
         .generate(&request, &allowed_code_repository, user_agent.as_deref())
```
wsxiaoys, line 56: the crlf / lf logic shall go inside of `CompletionService.generate`, instead of being implemented in `routes/…`

crates/tabby/src/services/completion.rs
```diff
@@ -321,35 +322,51 @@ impl CompletionService {
             self.config.max_decoding_tokens,
         );
 
+        let mut use_crlf = false;
         let (prompt, segments, snippets) = if let Some(prompt) = request.raw_prompt() {
             (prompt, None, vec![])
         } else if let Some(segments) = request.segments.as_ref() {
+            let mut new_segments = segments.clone();
+            if segments.prefix.contains("\r\n") {
```
wsxiaoys, line 330: what if prefix contains both `\r\n` and `\n`? what's the strategy here?

crates/tabby/src/services/completion.rs
```diff
@@ -321,35 +322,51 @@ impl CompletionService {
...
+                use_crlf = true;
+                new_segments.prefix = segments.prefix.replace("\r\n", "\n");
```
wsxiaoys, line 332: this replacement only applies for prefix / suffix, what about other segments? I suggest do deteciton on prefix / suffix, but the replacement shall happens on the whole prompt / output.

wsxiaoys (changes requested):

crates/tabby/src/services/completion.rs
zwpaper: normally, a file should not contain both `\r\n` and `\n` in the same time?

IMO, just replaced all CRLF to LF, and the same logic for output

crates/tabby/src/services/completion.rs
```diff
@@ -321,9 +322,19 @@ impl CompletionService {
...
             (prompt, None, vec![])
         } else if let Some(segments) = request.segments.as_ref() {
+            if segments.prefix.contains("\r\n") {
+                use_crlf = true;
+            }
+            if let Some(suffix) = &segments.suffix {
+                if suffix.contains("\r\n") {
+                    use_crlf = true;
+                }
+            }
```
wsxiaoys, line 336: extract this into a function (`contains_crlf(segments)`)

crates/tabby/src/services/completion.rs
```diff
@@ -335,21 +346,28 @@ impl CompletionService {
             let prompt = self
                 .prompt_builder
                 .build(&language, segments.clone(), &snippets);
-            (prompt, Some(segments), snippets)
+            if use_crlf {
+                (prompt.replace("\r\n", "\n"), Some(segments), snippets)
+            } else {
+                (prompt, Some(segments), snippets)
+            }
```
wsxiaoys, line 353: extract a function `override_prompt(prompt, use_crlf)`

crates/tabby/src/services/completion.rs
```diff
@@ -335,21 +346,28 @@ impl CompletionService {
...
             return Err(CompletionError::EmptyPrompt);
         };
 
-        let text = self.engine.generate(&prompt, options).await;
-        let segments = segments.cloned().map(|s| s.into());
+        let mut text = self.engine.generate(&prompt, options).await;
+        if use_crlf {
+            let re = Regex::new(r"([^\r])\n").unwrap(); // Match \n that is preceded by anything except \r
+            text = re.replace_all(&text, "$1\r\n").to_string() // Replace with captured character and \r\n
+        }
```
wsxiaoys, line 362: extract a function `override_text(text, use_crlf)`.

Please also document a bit why this regex is needed for replacement (instead of just `\n` -> `\r\n`)

wsxiaoys (changes requested):

crates/tabby/src/services/completion.rs
zwpaper: `text` seems to be too much general, I changed it to `generated`

crates/tabby/src/services/completion.rs
```diff
@@ -368,12 +374,47 @@ impl CompletionService {
...
+        prompt
+    }
+}
+
+/// override_generated replaces \n with \r\n in the generated text if use_crlf is true.
+/// This is used to ensure that the generated text has the same line endings as the prompt.
+///
+/// Because there might be \r\n in the text, which also has a `\n` and should not be replaced,
+/// we can not simply replace \n with \r\n.
+fn override_generated(generated: String, use_crlf: bool) -> String {
```
wsxiaoys, line 409:
```suggestion
fn override_generated_text(generated_text: String, use_crlf: bool) -> String {
```

wsxiaoys (approved):

PR #3310, opened 2024-10-23
chore(answer): adding debug log when including whole content triggered in answer engine

Sma1lboy:
Adding debug log

Also move start line to 1 when including whole content

image


wsxiaoys:
cc @liangfung

Issue #3239, opened 2024-10-05
Provide test connection button for remote chat / completion / embedding model in `System tab`


wsxiaoys:
**Please describe the feature you want**

1. Implementation of the GraphQL API endpoint `testModelConnection`.
2. The frontend implementation for testing connection logic.

eryue0220:
could I have a try if no one takes this ?

wsxiaoys:
@zwpaper is working on this

PR #3311, opened 2024-10-24
feat(graphQL): add model check for chat, completion, embedding

zwpaper:
- backend for https://github.com/TabbyML/tabby/issues/3239
- Both local model and HTTP model tested

CleanShot 2024-10-28 at 15 40 22@2x

CleanShot 2024-10-28 at 15 41 18@2x

CleanShot 2024-10-28 at 15 41 47@2x

notes:
1. the `CompletionService` in tabby-inference is wrapped by `CodeGeneration`, and `CodeGeneration` is then wrapped by `CompletionService` struct in tabby service, we should use the CompltionService in tabby-inference, but it's a bit tricky according to the current implementation, so I returned the `CodeGeneration` from tabby-inference, and use it as the GraphQL context.
2. when failed to connect to the model, GraphQL request will fail, so I just return the latency as the success indication, it's `milliseconds` i32.
4. following the GraphQL convention, the enum is in SCREAMING_SNAKE_CASE by default

### DIFF
--- a/ee/tabby-schema/src/schema/mod.rs
+++ b/ee/tabby-schema/src/schema/mod.rs
@@ -685,11 +685,7 @@ impl Query {
                         .build()
                         .expect("Failed to build completion options");
 
-                    if !completion
-                        .generate("hello Tabby", options)
-                        .await
-                        .is_empty()
-                    {
+                    if !completion.generate("hello Tabby", options).await.is_empty() {
                         return Ok(start.elapsed().as_millis() as i32);
                     }
 
--- a/ee/tabby-schema/src/schema/mod.rs
+++ b/ee/tabby-schema/src/schema/mod.rs
@@ -156,13 +156,13 @@ impl<S: ScalarValue> IntoFieldError<S> for CoreError {
 #[derive(thiserror::Error, Debug)]
 pub enum TestModelConnectionError {
     #[error("{0}")]
-    CoreError(#[from] CoreError),
-
-    #[error("{0}")]
-    ModelConnectionFailed(String),
+    FailedToConnect(String),
 
     #[error("Model backend is not enabled")]
     NotEnabled,
+
+    #[error("{0}")]
+    Other(String),
 }
 
 impl<S: ScalarValue> IntoFieldError<S> for TestModelConnectionError {
@@ -228,7 +228,7 @@ enum ModelHealthBackend {
 }
 
 #[derive(GraphQLObject, Debug, Clone)]
-struct ModelHealthResponse {
+struct ModelBackendHealthInfo {
     latency: i32,
 }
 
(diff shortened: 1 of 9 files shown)


ee/tabby-schema/src/schema/mod.rs
```diff
@@ -658,6 +668,25 @@ impl Query {
 
         Ok(SourceIdAccessPolicy { source_id, read })
     }
+
+    async fn test_model_connection(ctx: &Context) -> Result<ModelHealth> {
```
wsxiaoys, line 672: adding argument (can be an enum) of `chat`, `completion`, `embedding`.

zwpaper:
cc @liangfung

ee/tabby-schema/src/schema/mod.rs
```diff
@@ -658,6 +673,50 @@ impl Query {
 
         Ok(SourceIdAccessPolicy { source_id, read })
     }
+
+    async fn test_model_connection(ctx: &Context, backend: ModelHealthBackend) -> Result<i32> {
```
wsxiaoys, line 677: would it be better to create a new error type for Result (Result<i32, TestModelConnectionError>)? e.g we might wanna report different error type to user in future (e.g credential error, timeout, connection failed, ...)

wsxiaoys (commented):
Otherwise LGTM

ee/tabby-schema/src/schema/mod.rs
zwpaper: All the GraphQL APIs return the CoreError, so I added a field `ModelConnectionFailed(String)` in CoreError

- return NotFound, when the model backend is not running
- return ModelConnectionFailed with error message, when failed to connect model

ee/tabby-webserver/src/service/mod.rs
```diff
@@ -325,6 +341,8 @@ impl ServiceLocator for ArcServerContext {
 
 pub async fn create_service_locator(
     logger: Arc<dyn EventLogger>,
+    chat: Option<Arc<dyn ChatCompletionStream>>,
+    completion: Option<Arc<CodeGeneration>>,
```
wsxiaoys, line 345:
```suggestion
    completion: Option<Arc<CompletionStream>>,
```

`CodeGeneration` is a wrapper interface - use CompletionStream directly.

ee/tabby-schema/src/schema/mod.rs
```diff
@@ -188,6 +220,18 @@ async fn check_license(ctx: &Context, license_type: &[LicenseType]) -> Result<()
...
 
+#[derive(GraphQLEnum)]
+enum ModelHealthBackend {
+    Chat,
+    Completion,
+    Embedding,
+}
+
+#[derive(GraphQLObject, Debug, Clone)]
+struct ModelHealthResponse {
```
wsxiaoys, line 231:
```suggestion
struct ModelBackendHealthInfo {
```

ee/tabby-schema/src/schema/mod.rs
```diff
@@ -139,6 +153,24 @@ impl<S: ScalarValue> IntoFieldError<S> for CoreError {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum TestModelConnectionError {
+    #[error("{0}")]
+    CoreError(#[from] CoreError),
+
```
wsxiaoys, line 160:
```suggestion
```

Consider adding an `Other(String)` error type instead.

ee/tabby-schema/src/schema/mod.rs
```diff
@@ -139,6 +153,24 @@ impl<S: ScalarValue> IntoFieldError<S> for CoreError {
...
+    #[error("{0}")]
+    ModelConnectionFailed(String),
```
wsxiaoys, line 162:
```suggestion
    FailedToConnect(String),
```

wsxiaoys (changes requested):

PR #3321, opened 2024-10-25
chore(config): modify default function for code search params in answ…

wsxiaoys:
…er engine


wsxiaoys:
cc @zwpaper: this is the config we set for demo.tabbyml.com, changing it as default

Sma1lboy (approved):
LGTM

PR #3339, opened 2024-10-29
refactor: run make fix for rust code

wsxiaoys:


crates/tabby-inference/src/decoding.rs
```diff
@@ -125,7 +125,7 @@ mod tests {
         assert!(!trie.common_prefix_search(&text).is_empty());
 
         let qwen25coder = reverse("qwen25 style stop words;<|file_sep|>");
-        assert!(!trie.common_prefix_search(&qwen25coder).is_empty());
```
zwpaper, line 128: this change breaks `code::index::tests::test_code_splitter`

wsxiaoys: fixed

PR #3353, opened 2024-11-01
fix(vscode): log every n for selection sync failed warning.

icycodes:
cc @liangfung

### DIFF
--- a/clients/vscode/src/chat/WebviewHelper.ts
+++ b/clients/vscode/src/chat/WebviewHelper.ts
@@ -350,10 +350,11 @@ export class WebviewHelper {
     try {
       await this.client?.updateActiveSelection(context);
     } catch {
-      this.logger.logEveryN(
-        "sync-active-selection-failed",
-        100,
-        "warn",
+      this.logger.log(
+        {
+          every: 100,
+          level: "warn",
+        },
         "Active selection sync failed. Please update your Tabby server to the latest version.",
       );
     }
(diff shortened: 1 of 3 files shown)


clients/vscode/src/chat/WebviewHelper.ts
```diff
@@ -350,7 +350,12 @@ export class WebviewHelper {
     try {
       await this.client?.updateActiveSelection(context);
     } catch {
-      this.logger.warn("Active selection sync failed. Please update your Tabby server to the latest version.");
+      this.logger.logEveryN(
+        "sync-active-selection-failed",
+        100,
+        "warn",
+        "Active selection sync failed. Please update your Tabby server to the latest version.",
+      );
```
wsxiaoys, line 358:
```suggestion
      this.logger.logEveryN({
        key: "sync-active-selection-failed",
        log_every_n: 100,
        level: "warn",
        message: "Active selection sync failed. Please update your Tabby server to the latest version.",
      });
```

Provide reasonble default in function, so user can input only key, every, and message if needed

wsxiaoys (changes requested):
Otherwise LGTM

clients/vscode/src/logger.ts
```diff
@@ -1,36 +1,52 @@
...
+
+export interface LogEveryNOptions {
+  identifier?: string;
+  every?: number;
+  level?: LogLevel;
+}
+
+export interface LogOutputChannel extends VSCodeLogOutputChannel {
+  log(message: string, ...args: unknown[]): void;
+  log(options: LogEveryNOptions, message: string, ...args: unknown[]): void;
```
icycodes, line 15: @wsxiaoys  I redesigned the log-every-n interface, PTAL

Issue #3263, opened 2024-10-11
VSCode code inline completion with wrong indentation size


Sma1lboy:
**Describe the bug**
Sometimes, the code completion hint is not formatted according to the context above or below.

![image](https://github.com/user-attachments/assets/7a5e30e6-1432-4511-be62-ba0df6bc66cc)

**Information about your version**
Please provide output of `tabby --version`
tabby 0.17.0

icycodes:
Could you provide a detailed log referencing this completion request? It will be very helpful to address the problem.  
You can find it at http://localhost:8080/activities (please replace the host and port if you are not using the default local setup). Please unfold the `Completion` event and copy the full JSON.

Sma1lboy:
I have performed this test multiple times. I apologize for the messy data here.
Thank you for your help.

```json
    "completion": {
      "completion_id": "cmpl-51f901f2-7369-41f5-9554-fee0e27830a7",
      "language": "json",
      "prompt": "<fim_prefix>  ],\n  \"icon\": \"assets/logo.png\",\n  \"engines\": {\n    \"vscode\": \"^1.82.0\"\n  },\n  \"categories\": [\n    \"Programming Languages\",\n    \"Machine Learning\",\n    \"Other\"\n  ],\n  \"activationEvents\": [\n    \"onStartupFinished\"\n  ],\n  \"main\": \"./dist/node/extension.js\",\n  \"browser\": \"./dist/browser/extension.js\",\n  \"contributes\": {\n    \"commands\": [\n      {\n        \"command\": \"tabby.server.test\",\n        <fim_suffix>\n      },\n      {\n        \"command\": \"tabby.toggleInlineCompletionTriggerMode\",\n        \"title\": \"Toggle Code Completion Trigger Mode (Automatic/Manual)\",\n        \"category\": \"Tabby\"\n      },\n      {\n        \"command\": \"tabby.inlineCompletion.trigger\",\n        \"title\": \"Trigger Code Completion Manually\",\n        \"category\": \"Tabby\"\n      },\n      {\n        \"command\": \"tabby.setApiEndpoint\",\n        \"title\": \"Connect to Server...\",\n        \"category\": \"Tabby\"\n      },\n      {\n        \"command\": \"tabby.setApiToken\",\n        \"title\": \"Set Token...\",\n<fim_middle>",
      "segments": {
[...]

PR #3361, opened 2024-11-02
fix(agent): normalize indentation in code completion postprocess

Sma1lboy:
#3263

Fixing wrong indent in inline completion predict solution

#### Demo
https://github.com/user-attachments/assets/34580f0a-ab4b-4c08-8101-ba28f2687701

### Before
Only first line has extra spaces:
image

exactly one extra space in the line indent
image

### After
image
exactly one extra space in the line indent
image

### Implementation
Using whether the current indent is even as the judgment basis. If it's even, the prediction doesn't need any extra spaces and the prediction context will be processed accordingly.

### Potential risks:
Some users might use odd indent size
### Why
This approach is simple and doesn't require adding additional LSP or obtaining the best indent through client operations.

### Issue didn't fix
Due to some concern this post-processing not going to fix this situation

All lines in prediction have extra spaces:
image

### DIFF
--- a/clients/tabby-agent/src/codeCompletion/postprocess/index.ts
+++ b/clients/tabby-agent/src/codeCompletion/postprocess/index.ts
@@ -13,6 +13,7 @@ import { trimMultiLineInSingleLineMode } from "./trimMultiLineInSingleLineMode";
 import { dropDuplicated } from "./dropDuplicated";
 import { dropMinimum } from "./dropMinimum";
 import { calculateReplaceRange } from "./calculateReplaceRange";
+import { normalizeIndentation } from "./normalizeIndentation";
 
 type ItemListFilter = (items: CompletionItem[]) => Promise<CompletionItem[]>;
 
@@ -51,6 +52,7 @@ export async function postCacheProcess(
     .then(applyFilter(limitScope))
     .then(applyFilter(removeDuplicatedBlockClosingLine))
     .then(applyFilter(formatIndentation))
+    .then(applyFilter(normalizeIndentation))
     .then(applyFilter(dropDuplicated))
     .then(applyFilter(trimSpace))
     .then(applyFilter(dropMinimum))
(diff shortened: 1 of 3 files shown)


Sma1lboy:
Follow up: This is kind of a post-processing filter to prevent inconsistent extra spaces due to different model capabilities. The above bad example happens when using the demo website model, but sometimes works fine with StarCoder-1B without this modification.

icycodes (changes requested):

icycodes:
As all bad cases in #3263 have exactly one extra space in the line indent, I suggest we should only process the case of exactly one extra space in the line indent, not extra indent.  
It should be allowed to have extra indents in the completion text. Consider the case of

```
function foo() {
║
}
``` 
and the completion is
```
function foo() {
├    bar();┤
}
```

Sma1lboy:
> As all bad cases in #3263 have exactly one extra space in the line indent, I suggest we should only process the case of exactly one extra space in the line indent, not extra indent. It should be allowed to have extra indents in the completion text. Consider the case of
> 
> ```
> function foo() {
> ║
> }
> ```
> 
> and the completion is
> 
> ```
> function foo() {
> ├    bar();┤
> }
> ```

The current version should work fine with this.

icycodes (approved):

sboys3:
This merged pull request mangles auto-completions when tabs are used instead of spaces. 

It replaces parts of correctly tabbed responses with incorrect amounts of spaces. This has been driving me insane for a while, and today after reviewing completion logs I realized it was entirely client side. I had thought it was due to a mix of spaced type definitions and tabbed code when using javascript, but when I also saw it in a small c++ project, I looked into it deeper.

As a quick test I replaced `return item.withText(normalizedLines.join(""));` with `return item;` in the agent's bundle in my extension, and the problem completely went away with this code bypassed.
To do that in the current production bundle located at %userprofile%\.vscode\extensions\tabbyml.vscode-tabby-1.18.0\dist\tabby-agent\node\index.js replace `e.withText(n.join(""))` with `e/*indent fix*/`

Here is an example of what has been happening. It has been extremely frustrating having to constantly fix indention.
Screenshot 2025-01-25 185534
Screenshot 2025-01-25 190639

And here is the result with this pull request bypassed
image

Sma1lboy:
@sboys3 I am very sorry for causing this problem, I will fix the bug caused by this PR as quickly as possible

PR #3376, opened 2024-11-05
fix(vscode): fix QuickFix codeaction cannot correct get correct range

Sma1lboy:
fix quick fix cannot cover all error range

### DIFF
--- a/clients/vscode/src/Commands.ts
+++ b/clients/vscode/src/Commands.ts
@@ -313,14 +313,13 @@ export class Commands {
         },
       };
       if (range) {
-        const FIXME_OFFSET = 10;
         editLocation = {
           uri: editor.document.uri.toString(),
           range: {
-            start: { line: range.start.line - FIXME_OFFSET, character: 0 },
+            start: { line: range.start.line, character: 0 },
             end: {
-              line: range.end.line + 1 + FIXME_OFFSET,
-              character: Number.MAX_SAFE_INTEGER,
+              line: range.end.line === 0 ? range.end.line : range.end.line + 1,
+              character: range.end.line === 0 ? Number.MAX_SAFE_INTEGER : 0,
             },
           },
         };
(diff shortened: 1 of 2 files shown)


clients/vscode/src/Commands.ts
```diff
@@ -311,6 +312,18 @@ export class Commands {
...
         },
       };
+      if (range) {
+        editLocation = {
+          uri: editor.document.uri.toString(),
+          range: {
+            start: { line: range.start.line, character: 0 },
+            end: {
+              line: range.end.line === 0 ? range.end.line : range.end.line + 1,
+              character: range.end.line === 0 ? Number.MAX_SAFE_INTEGER : 0,
```
icycodes, line 322: I believe the range process should aim to ensure that the range always covers full lines. However, the check here is `range.end.line` instead of `range.end.character`, which may be a mistake.

Additionally, this process should be same with Line 305-313, where you extract `editor.selection` to `editRange` and use `const editRange = range ? editor.selection` in order to reuse this process.

clients/vscode/src/code-action/QuickFix.ts
```diff
@@ -22,33 +25,56 @@ export class QuickFixCodeActionProvider implements CodeActionProviderInterface {
...
+        (mergedRange, diagnostic) => {
+          if (!mergedRange) {
+            return diagnostic.range;
+          }
+          const startLine = Math.min(mergedRange.start.line, diagnostic.range.start.line);
+          const startChar = Math.min(mergedRange.start.character, diagnostic.range.start.character);
+          const endLine = Math.max(mergedRange.end.line, diagnostic.range.end.line);
+          const endChar = Math.max(mergedRange.end.character, diagnostic.range.end.character);
+
+          return new Range(new Position(startLine, startChar), new Position(endLine, endChar));
```
icycodes, line 46: You can use `Range.union` to reduce diagnostics ranges.

icycodes (changes requested):

icycodes (approved):

Issue #2965, opened 2024-08-23
[VSCode] When a code completion is selected, automatically invoke `import fix` code action if possible


wsxiaoys:
References: https://code.visualstudio.com/api/references/commands
![image](https://github.com/user-attachments/assets/5fb6fdfe-1969-440c-b4f9-4585ca6e0d3d)

Example:
```
const codeActions = await vscode.commands.executeCommand<vscode.CodeAction[]>(
    'vscode.executeCodeActionProvider',
    document.uri,
    range,
    vscode.CodeActionKind.QuickFix.value,
    7
  );
```

shankarg87:
can take a look if no-one is working on this

shankarg87:
Went through the code and a bit lost. Some follow up questions:

```When a code completion is selected``` - Can you clarify you mean by this? Are you referring to the ```inlineCompletionTriggerMode```, whenever the user triggers it ? Or do you mean once the user accepts the suggestion of the inline code completion. Trying to understand what the trigger for the code action is expected to be.

```automatically invoke import fix code action``` - I see that the existing CodeProvider registers the ```CodeAction.RefactorRewrite``` CodeAction. This can be extended to ```CodeActionKind.QuickFix```. Any thoughts on what the user workflow would look like for QuickFix? Would it also be inline editing (```tabby.chat.edit.start```) or are we looking for a new workflow.

Appreciate any pointers!

wsxiaoys:
> When a code completion is selected - Can you clarify you mean by this? Are you referring to the inlineCompletionTriggerMode, whenever the user triggers it ? Or do you mean once the user accepts the suggestion of the inline code completion.

When I mentioned "When a code completion is selected," I specifically meant **when the user accepts the suggestion of the inline code completion**. This is the point at which the code completion is integrated into the user's code.

> automatically invoke import fix code action - I see that the existing CodeProvider registers the CodeAction.RefactorRewrite CodeAction. This can be extended to CodeActionKind.QuickFix. Any thoughts on what the user workflow would look like for QuickFix?

The work need to be done here is:

1. Verify if the generated code contains any quick fix action.

2. If the action can be classified as `import libraries`, trigger the action (such actions is usually provided by an LSP implementation for that languages)

PR #3389, opened 2024-11-08
feat(vscode): adding auto-import during accept inline completion item

Sma1lboy:
still working 

https://github.com/user-attachments/assets/0f1c29ea-5e5d-492a-8e08-44a646a0c473

fix https://github.com/TabbyML/tabby/issues/2965

### DIFF
--- a/clients/vscode/src/Commands.ts
+++ b/clients/vscode/src/Commands.ts
@@ -224,20 +224,20 @@ export class Commands {
     "inlineCompletion.trigger": () => {
       commands.executeCommand("editor.action.inlineSuggest.trigger");
     },
-    "inlineCompletion.accept": () => {
+    "inlineCompletion.accept": async () => {
       const editor = window.activeTextEditor;
       if (!editor) {
         return;
       }
 
-      commands.executeCommand("editor.action.inlineSuggest.commit");
-
       const uri = editor.document.uri;
-      const range = this.inlineCompletionProvider.getCurrentDisplayRange();
-      if (!range) {
-        return;
+      const range = this.inlineCompletionProvider.calcEditedRangeAfterAccept();
+
+      await commands.executeCommand("editor.action.inlineSuggest.commit");
+
+      if (range) {
+        applyQuickFixes(uri, range);
       }
-      applyQuickFixes(uri, range);
     },
     "inlineCompletion.acceptNextWord": () => {
       this.inlineCompletionProvider.handleEvent("accept_word");
--- a/clients/vscode/src/Commands.ts
+++ b/clients/vscode/src/Commands.ts
@@ -13,6 +13,8 @@ import {
   QuickPickItem,
   ViewColumn,
   Range,
+  CodeAction,
+  CodeActionKind,
 } from "vscode";
 import os from "os";
 import path from "path";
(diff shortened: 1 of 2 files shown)


clients/vscode/src/InlineCompletionProvider.ts
```diff
@@ -211,4 +211,28 @@ export class InlineCompletionProvider extends EventEmitter implements InlineComp
     // await not required
     this.client.telemetry.postEvent(params);
   }
+
+  /**
+   * Get current display range with the current completion item
```
icycodes, line 216:
```suggestion
   * Calculate the edited range in the modified document as if the current completion item has been accepted.
```

clients/vscode/src/InlineCompletionProvider.ts
```diff
@@ -211,4 +211,28 @@ export class InlineCompletionProvider extends EventEmitter implements InlineComp
...
+   * @return {Range | undefined} - The range with the current completion item
+   */
+  public getCurrentDisplayRange(): Range | undefined {
```
icycodes, line 219: I prefer to name it `calcEditedRangeAfterAccept`

clients/vscode/src/InlineCompletionProvider.ts
```diff
@@ -211,4 +211,28 @@ export class InlineCompletionProvider extends EventEmitter implements InlineComp
...
+    const item = this.displayedCompletion?.completions.items[this.displayedCompletion.index];
+    const range = item?.range;
+    if (!range) {
+      return undefined;
```
icycodes, line 223: I left a note here. It's not necessary to fix it in the PR.
```suggestion
      // FIXME: If the item has a null range, we can use current position and text length to calculate the result range
      return undefined;
```

clients/vscode/src/InlineCompletionProvider.ts
```diff
@@ -211,4 +211,28 @@ export class InlineCompletionProvider extends EventEmitter implements InlineComp
...
+    }
+    if (!item) {
+      return undefined;
+    }
+    const length = (item.insertText as string).split("\n").length - 1; //remove current line count;
+    const completionEnd = range.end.line + length;
+    const completionRange = new Range(
+      new Position(range.start.line, range.start.character),
+      new Position(completionEnd == 0 ? completionEnd : completionEnd + 1, 0),
```
icycodes, line 232: I believe we should also add `+ 1` when the completionEnd line index is 0.
```suggestion
      new Position(completionEnd + 1, 0),
```

clients/vscode/src/InlineCompletionProvider.ts
```diff
@@ -211,4 +211,28 @@ export class InlineCompletionProvider extends EventEmitter implements InlineComp
...
+    );
+    this.logger.info("current range with displayed completion item:", completionRange);
```
icycodes, line 234:
```suggestion
    this.logger.debug("Calculate edited range for displayed completion item:", completionRange);
```

clients/vscode/src/Commands.ts
```diff
@@ -223,7 +225,19 @@ export class Commands {
...
+        return;
+      }
+
       commands.executeCommand("editor.action.inlineSuggest.commit");
+
+      const uri = editor.document.uri;
+      const range = this.inlineCompletionProvider.getCurrentDisplayRange();
+      if (!range) {
+        return;
+      }
```
icycodes, line 239:
```suggestion

      const range = this.inlineCompletionProvider.getCurrentDisplayRange();
      await commands.executeCommand("editor.action.inlineSuggest.commit");

      const uri = editor.document.uri;
      if (!range) {
        return;
      }
```

icycodes: Should it `await` the edit action of accept inline completion?

clients/vscode/src/InlineCompletionProvider.ts
icycodes: I left a note here. It's not necessary to fix it in the PR.

clients/vscode/src/Commands.ts
Sma1lboy: The await causes us to be unable to get the current display item because it is no longer available. I moved the part that retrieves the edited range before committing the inline suggestion. The range should always exist if a code prediction is present.

icycodes (approved):

PR #3392, opened 2024-11-09
feat(agent): add removeDuplicateSuffixLines postprocess filter

Sma1lboy:
Adding drop duplicate suffix lines in completion


clients/tabby-agent/src/codeCompletion/postprocess/removeDuplicateSuffixLines.test.ts
```diff
@@ -0,0 +1,152 @@
...
+            3,
+            4,┤
+      `;
+      context.suffix = `
+            4,
+            5,
+            6
+          ];
+        }
+      `;
```
icycodes, line 29:
```suggestion
```

clients/tabby-agent/src/codeCompletion/postprocess/removeDuplicateSuffixLines.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    const filter = removeDuplicateSuffixLines();
+
+    it("should remove duplicated suffix lines", async () => {
+      const context = documentContext`
+        function example() {
+          const items = [
+            ║
+          ];
+        }
+      `;
```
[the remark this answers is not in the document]
icycodes: We can add suffix in `documentContext`   
Please also update code for other test cases.

clients/tabby-agent/src/codeCompletion/postprocess/removeDuplicateSuffixLines.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        // if all checked lines match, check for code structure
+        if (isFullMatch) {
+          const remainingLines = originalLines.slice(0, i);
+          const lastLine = remainingLines[remainingLines.length - 1] || "";
+
+          // skip empty last lines
+          if (isBlank(lastLine.trim())) {
+            return item;
+          }
```
icycodes, line 56: I am not sure why we do not handle the case of an empty last line. Could you give an example?  
By the way, it would be good to add a unit test case for each if branch.

clients/tabby-agent/src/codeCompletion/postprocess/removeDuplicateSuffixLines.ts
```diff
@@ -0,0 +1,70 @@
...
+
+          // preserve code block structure
+          if (lastLine.includes("{") || currentLine.includes("}")) {
+            return item;
+          }
```
icycodes, line 61: Could you also give an example for this case?
As `{` and `}` are language-related, please avoid processing like this if possible, and add a guard check for `context.language` if necessary.

clients/tabby-agent/src/codeCompletion/postprocess/index.ts
```diff
@@ -54,5 +55,6 @@ export async function postCacheProcess(
     .then(applyFilter(dropDuplicated))
     .then(applyFilter(trimSpace))
     .then(applyFilter(dropMinimum))
-    .then(applyFilter(calculateReplaceRange));
+    .then(applyFilter(calculateReplaceRange))
+    .then(applyFilter(removeDuplicateSuffixLines));
```
icycodes, line 59: I think it should be added before `dropMinimum`

clients/tabby-agent/src/codeCompletion/postprocess/removeDuplicateSuffixLines.ts
```diff
@@ -0,0 +1,70 @@
...
+  return (item: CompletionItem): CompletionItem => {
+    const log = getLogger("removeDuplicateSuffixLines");
+    log.info("Processing item" + JSON.stringify(item?.text || ""));
+
+    const text = item?.text;
+    const suffix = item?.context?.suffix;
+
+    if (text == null || suffix == null) {
+      return item;
+    }
```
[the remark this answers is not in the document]
Sma1lboy: While `context.suffixLines` includes whitespace, `item.lines` does not, which creates a discrepancy. Additionally, I need to manually remove all \n characters when comparing these values to ensure consistency.

clients/tabby-agent/src/codeCompletion/postprocess/removeDuplicateSuffixLines.ts
Sma1lboy: offline update: We decided to remove this part and change the post-processing strategy. Instead of comparing trimmed lines, we will compare lines without newline characters.

icycodes (approved):

PR #3398, opened 2024-11-11
docs: fix deepseek model provider documentation

wsxiaoys:


website/docs/references/models-http-api/deepseek.md
```diff
@@ -12,7 +12,7 @@ Below is an example
 [model.chat.http]
 kind = "openai/chat"
 model_name = "your_model"
-api_endpoint = "https://api.deepseek.com/chat"
```
zwpaper, line 15: according to https://api-docs.deepseek.com/, `v1` is a compatible suffix to OpenAI spec, since this docs is deepseek specific, how about we use `https://api.deepseek.com` directly?

wsxiaoys: Given the kind is still `openai/chat` - let's make it consistent with other openai kind setup for now.

PR #3404, opened 2024-11-11
docs(models-http-api): add jan.ai chat example

zwpaper:
janAI [completion](https://cortex.so/docs/capabilities/text-generation) and [embedding](https://cortex.so/docs/capabilities/embeddings) APIs are both under construction, and not yet supported by Jan now


wsxiaoys:
Please also draft a PR for jan, example: https://github.com/janhq/jan/blob/4e91c80bcc9baf26ba35d085bbe98dd6a0e3d5eb/docs/src/pages/integrations/coding/continue-dev.mdx#L2

PR #3406, opened 2024-11-11
feat(ui): add collapsible interaction to sidebar

liangfung:
## Changes
1. Integrate the sidebar component from shadcn-ui.
2. Adjust layout and modify some styles to optimize frame dropping issues.
3. resolve the issue where the scrollbar does not return to the top of the page after navigation
4. Resolve the issue of layout height fluctuations on some pages.

## Screen record
https://jam.dev/c/752af652-2a0f-42f8-801b-7c27c7aaf2e6


ee/tabby-ui/tailwind.config.js
wsxiaoys, line 1: please change the indent being consistent with original version so the diff can be more meaningful

liangfung: done

PR #3413, opened 2024-11-13
feat(tabby-index): add filepath embedding tokens into code index

wsxiaoys:


crates/tabby/src/services/code.rs
```diff
@@ -136,6 +138,18 @@ fn merge_code_responses_by_rank(
     }
 }
 
+fn retain_at_most_two_hits_per_file(scored_hits: &mut Vec<CodeSearchHit>) {
```
wsxiaoys, line 141: @zwpaper could you help add unit test for this function?

PR #3432, opened 2024-11-18
feat(tabby-chat-panel): adding capability check in tabby-ui and adding applyInEditorV2

Sma1lboy:
working on: tabby-ui hasCapability check and smartApplyV2

waiting for #3474


clients/tabby-chat-panel/src/index.ts
```diff
@@ -1,8 +1,11 @@
-import { createThreadFromIframe, createThreadFromInsideIframe } from '@quilted/threads'
+import { createThreadFromIframe } from '@quilted/threads'
```
wsxiaoys, line 1: consider get rid of @quilted/threads completely if we go with this approach.

Sma1lboy:
update: waiting for tabby-threads modification first

clients/tabby-chat-panel/src/index.ts
Sma1lboy: offline update: #3474

ee/tabby-ui/app/chat/page.tsx
```diff
@@ -222,11 +223,27 @@ export default function ChatPage() {
     }
   }, [server, client])
 
+  const [serverCapabilities, setServerCapabilities] = useState(
+    new Map<string, boolean>()
+  )
+
```
wsxiaoys, line 229: instead of using a map, use explicit boolean field for ease of compiler check.

e.g
```
const supportsOnApplyInEditorV2 = useState(false);

...
```

wsxiaoys (approved):
Otherwise LGTM

PR #3435, opened 2024-11-18
fix(context): not add doc to cache if no embed calculated

zwpaper:


crates/tabby-index/src/structured_doc/public.rs
```diff
@@ -34,16 +35,32 @@ impl StructuredDocIndexer {
             return false;
         };
 
-        stream! {
+        let docs: Vec<TantivyDocument> = stream! {
```
wsxiaoys, line 38:
```suggestion
        let count_docs = stream! {
```

crates/tabby-index/src/indexer.rs
```diff
@@ -110,6 +110,9 @@ impl<T: ToIndexId> TantivyDocBuilder<T> {
                     let Ok((tokens, chunk_attributes)) = task.await else {
                         return None;
                     };
+                    if tokens.is_empty() {
```
wsxiaoys, line 113: add a comment to explain why tokens is empty should return None

crates/tabby-index/src/testutils.rs
wsxiaoys, line 1: should be structured_doc_test.rs

crates/tabby-inference/src/embedding.rs
wsxiaoys, line 1: move the mock implementation to structured_doc_test.rs

crates/tabby-index/src/structured_doc/public.rs
```diff
@@ -31,7 +31,9 @@ impl StructuredDocIndexer {
             return false;
         }
 
-        if self.indexer.is_indexed_after(document.id(), updated_at) {
+        if self.indexer.is_indexed_after(document.id(), updated_at)
+            && self.indexer.failed_chunks_count(document.id()) == 0
+        {
```
wsxiaoys, line 36: combine two function into a single call, e.g `self.indexer.require_updates(id, updated_at)`

crates/tabby-index/src/indexer.rs
```diff
@@ -72,21 +72,39 @@ impl<T: ToIndexId> TantivyDocBuilder<T> {
...
         let s = stream! {
-            yield tokio::spawn(async move { Some(doc) });
+            let mut failed_count: u64 = 0;
+            for await chunk_doc in self.build_chunks(cloned_id, source_id.clone(), updated_at, document).await {
+                match chunk_doc.await {
+                    Ok((Some(doc), ok)) => {
+                        if !ok {
+                            failed_count += 1;
+                        }
+                        yield tokio::spawn(async move { Some(doc) });
```
wsxiaoys, line 86: do it still need to yield when `!ok`?

crates/tabby-index/src/indexer.rs
```diff
@@ -72,21 +72,39 @@ impl<T: ToIndexId> TantivyDocBuilder<T> {
...
-            for await doc in self.build_chunks(cloned_id, source_id, updated_at, document).await {
-                yield doc;
-            }
+            let doc = doc! {
+                schema.field_id => doc_id,
+                schema.field_source_id => source_id,
+                schema.field_corpus => self.corpus,
+                schema.field_attributes => doc_attributes,
+                schema.field_updated_at => updated_at,
+                schema.field_failed_chunks_count => failed_count,
```
wsxiaoys, line 104: you might only set failed_chunks_count when failed_count > 0, so the search query can be simplified slightly.

crates/tabby-index/src/indexer.rs
```diff
@@ -226,6 +249,31 @@ impl Indexer {
 
         !docs.is_empty()
     }
+
+    /// Get the failed_chunks_count field for a document.
+    /// tracks the number of embedding indexing failed chunks for a document.
+    ///
+    /// return 0 if the field is not found.
+    pub fn failed_chunks_count(&self, id: &str) -> u64 {
```
wsxiaoys, line 257:
```suggestion
    pub fn has_failed_chunks(&self, id: &str) -> bool {
```

wsxiaoys (changes requested):
Otherwise LGTM

crates/tabby-index/src/indexer.rs
zwpaper: yes, the doc itself still matters and could be used to search in tantivy, the `!ok` only indicates the embedding is failed calculated.

crates/tabby-index/src/indexer.rs
```diff
@@ -226,6 +251,29 @@ impl Indexer {
...
+        // };
+        let docs = match self.searcher.search(&query, &TopDocs::with_limit(1)) {
+            Ok(docs) => docs,
+            Err(e) => {
+                println!("error: {:?}", e);
+                return false;
+            }
+        };
+
+        println!("docs: {:?}", docs);
```
wsxiaoys, line 273: remove all useless println

crates/tabby-index/src/tests.rs
wsxiaoys, line 1: split to intelligence_test / structured_doc_test?

PR #3448, opened 2024-11-21
ci(llama-server): update vulkan build option and use manylinux_2_28

zwpaper:
pipeline: https://github.com/zwpaper/tabby/actions/runs/11951452245/job/33315134871

before:
CleanShot 2024-11-21 at 18 51 00@2x

after:

CleanShot 2024-11-21 at 18 50 18@2x


zwpaper:
tested in latest build:
CleanShot 2024-11-25 at 11 07 05@2x

PR #3468, opened 2024-11-26
feat(models-http-api): add rate limit support for completion and chat

zwpaper:
Completion rate limited:

CleanShot 2024-11-26 at 12 48 34@2x

Chat rate limited:
CleanShot 2024-11-26 at 12 55 32@2x


crates/http-api-bindings/src/chat/rate_limit.rs
```diff
@@ -0,0 +1,64 @@
...
+            if let Err(sleep) = self.rate_limiter.try_wait() {
+                tokio::time::sleep(sleep).await;
+                continue;
+            }
+
+            return self.completion.chat_stream(request).await;
+        }
+
+        // Return an empty stream if the rate limit is exceeded
+        Ok(Box::pin(futures::stream::empty()))
```
wsxiaoys, line 62: return error instead?

crates/http-api-bindings/src/chat/mod.rs
```diff
@@ -31,8 +34,16 @@ pub async fn create(model: &HttpModelConfig) -> Arc<dyn ChatCompletionStream> {
...
     let config = builder.build().expect("Failed to build config");
 
-    Arc::new(
+    let engine = Box::new(
         async_openai::Client::with_config(config)
             .with_http_client(create_reqwest_client(api_endpoint)),
-    )
+    );
+
+    let ratelimiter =
```
wsxiaoys, line 42: extract function to reuse logic with embedding rate limiyer creation

crates/http-api-bindings/src/completion/mod.rs
```diff
@@ -1,65 +1,63 @@
...
+            false,
+        ),
         unsupported_kind => panic!(
             "Unsupported model kind for http completion: {}",
             unsupported_kind
         ),
-    }
+    };
+
+    let ratelimiter =
```
wsxiaoys, line 54: ditto

crates/http-api-bindings/src/rate_limit.rs
wsxiaoys, line 1: nit: instead of exposing `RateLimitedEmbedding`, just expose `new_embedding(...) -> impl Embedding` for better encapsulation.

Similar to completion / chat.

Issue #3464, opened 2024-11-25
Local models are no longer being loaded


mendozal3:
**Describe the bug**

I downloaded the StarCoder2-15B model and loaded it from a local path onto tabby. This works fine with Tabby 0.18.0 but after updating this no longer works. The application just panics and gives the following
![image](https://github.com/user-attachments/assets/a4036655-177a-4e76-a394-c4a65d1c3ffd)
Works fine again after reverting to 0.18.0.

**Information about your version**

Currently at tabby 0.18.0 and attempting to update to tabby 0.20.0

**Information about your GPU**

![image](https://github.com/user-attachments/assets/89bd55c7-f995-4d8d-9b09-1d384586231a)

**Additional context**

Tried updating to 0.19.0 instead and that gives me the same error. Noticed the Model spec changed a bit so I renamed the model.gguf to model-00001-of-00001.gguf and gave the same error.

PR #3470, opened 2024-11-26
fix: should check local model before resolving model id

zwpaper:
CleanShot 2024-11-26 at 15 13 22@2x

fix https://github.com/TabbyML/tabby/issues/3464


.changes/unreleased/Fixed and Improvements-20241126-151903.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed and Improvements
+body: Fixed a panic that occurred when specifying a local model
```
wsxiaoys, line 2: add github issue id

PR #3475, opened 2024-11-26
feat(tabby-threads): using tabby thread in chat-panel and VSCode plugin

Sma1lboy:
still working

This PR should review after #3474


clients/tabby-chat-panel/src/index.ts
```diff
@@ -69,6 +71,10 @@ export interface ClientApi {
   onCopy: (content: string) => void
 
   onKeyboardEvent: (type: 'keydown' | 'keyup' | 'keypress', event: KeyboardEventInit) => void
+
+  // this is inner function cover by tabby-threads
+  // the function doesn't need to expose to client but can call by client
+  hasCapability?: (capability: ClientApiMethods) => Promise<boolean>
 }
```
icycodes, line 78: I think using interface extends could be better?
```
export interface ClientApiMethods {
  // ...
}
export interface ClientApi extends ClientApiMethods {
  hasCapability(method: key of ClientApiMethods): Promise<boolean>
}
```

Sma1lboy: That makes a lot of sense, thanks for the advice. :^

icycodes:
We need to update the version of tabby-chat-panel in https://github.com/TabbyML/tabby/blob/main/clients/tabby-chat-panel/package.json#L4

Otherwise LGTM

icycodes (approved):

PR #3482, opened 2024-11-28
refactor(config): increase rate limit to 1200, set 6000 for llama-cpp…

wsxiaoys:
…-server based config

### DIFF
--- a/crates/http-api-bindings/src/rate_limit.rs
+++ b/crates/http-api-bindings/src/rate_limit.rs
@@ -11,17 +11,12 @@ use futures::stream::BoxStream;
 use ratelimit::Ratelimiter;
 use tabby_inference::{ChatCompletionStream, CompletionOptions, CompletionStream, Embedding};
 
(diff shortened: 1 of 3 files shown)


crates/http-api-bindings/src/rate_limit.rs
```diff
@@ -11,28 +11,23 @@ use futures::stream::BoxStream;
...
         .max_tokens(rpm)
         .initial_available(rpm)
         .build()
-        .map_err(|e| {
-            anyhow::anyhow!(
-                "Failed to create ratelimiter, please check the rate limit configuration: {}",
-                e,
-            )
-        })
+        .expect("Failed to create RateLimiter, please check the HttpModelConfig.rate_limit configuration")
```
zwpaper, line 19:
```suggestion
        .expect("Failed to create RateLimiter, please check the model.*.http.rate_limit.rate_limit configuration")
```

Perhaps we could assist users by specifying the configuration details directly like this?

wsxiaoys: ack.

PR #3490, opened 2024-11-28
fix: code index has no failed chunk count even when embed failed

zwpaper:
Signed-off-by: Wei Zhang <kweizh@tabbyml.com>

### DIFF
--- a/crates/tabby-index/src/code/intelligence.rs
+++ b/crates/tabby-index/src/code/intelligence.rs
@@ -246,7 +246,7 @@ mod metrics {
 }
 
 #[cfg(test)]
-pub mod tests {
+mod tests {
     use serial_test::file_serial;
     use tabby_common::path::set_tabby_root;
     use tracing_test::traced_test;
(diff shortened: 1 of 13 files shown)


crates/tabby-index/src/structured_doc/mod.rs
```diff
@@ -37,7 +38,7 @@ impl IndexAttributeBuilder<StructuredDoc> for StructuredDocBuilder {
     async fn build_chunk_attributes<'a>(
         &self,
         document: &'a StructuredDoc,
-    ) -> BoxStream<'a, JoinHandle<(Vec<String>, serde_json::Value)>> {
+    ) -> BoxStream<'a, JoinHandle<(Result<Vec<String>>, serde_json::Value)>> {
```
wsxiaoys, line 41:
```suggestion
    ) -> BoxStream<'a, JoinHandle<(Result<Vec<String>, serde_json::Value)>>> {
```

in future, `serde_json::Value` might contains a value coming from a fail-able source.

crates/tabby-index/src/code/intelligence.rs
```diff
@@ -246,7 +246,7 @@ mod metrics {
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
```
wsxiaoys, line 249: extract this part to `src/code_testutils.rs`, instead of marking it as public.

PR #3496, opened 2024-12-02
fix(indexer): fan out when index code chunks

zwpaper:


crates/tabby-index/src/indexer.rs
```diff
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::{collections::HashSet, sync::mpsc::sync_channel};
```
wsxiaoys, line 1: need use tokio channel instead of std channel to utilize `async`

PR #3506, opened 2024-12-04
feat(vscode): adding checkbox for features in tabby command palette

Sma1lboy:
before:
image

after:
imagedisconnected
image

### Next Step
Modularizing Different features and configure them


clients/vscode/src/commands/commandPalette.ts
```diff
@@ -16,60 +19,85 @@ export class CommandPalette {
...
 
+    // Features section
+    const validStatuses = ["ready", "readyForAutoTrigger", "readyForManualTrigger"];
+    if (validStatuses.includes(this.client.status.current?.status || "")) {
+      items = items.concat([
+        { label: "enable/disable features", kind: QuickPickItemKind.Separator },
+        {
+          label:
+            (this.config.inlineCompletionTriggerMode === "automatic" ? "" : MENU_ITEM_INDENT_SPACING) +
+            "Code Completions",
```
wsxiaoys, line 43:
```suggestion
            "Code Completion",
```

clients/vscode/src/commands/commandPalette.ts
```diff
@@ -16,60 +19,85 @@ export class CommandPalette {
...
+    if (disconnectStatues.includes(this.client.status.current?.status || "")) {
+      items.push(this.itemForStatus());
+    }
 
-    items.push(this.itemForStatus());
+    // Status section
     this.client.status.on("didChange", () => {
       items[0] = this.itemForStatus();
       quickPick.items = items;
     });
```
icycodes, line 33: This listener function of status changing will always update the first item.  
But the first item may not be status after this PR.

icycodes: I suggest keeping the first item to show the status

clients/vscode/src/commands/commandPalette.ts
```diff
@@ -16,60 +19,85 @@ export class CommandPalette {
...
 
+    // Features section
+    const validStatuses = ["ready", "readyForAutoTrigger", "readyForManualTrigger"];
+    if (validStatuses.includes(this.client.status.current?.status || "")) {
+      items = items.concat([
```
icycodes, line 38: I think using `items.push` and keeping `items` as a const would be better.

clients/vscode/src/commands/commandPalette.ts
```diff
@@ -16,60 +19,85 @@ export class CommandPalette {
...
-        label: "$(plug) Connect to Server...",
+        label: "Connect to Server",
         command: "tabby.connectToServer",
+        iconPath: new ThemeIcon("plug"),
       },
       {
-        label: "$(settings) Settings...",
+        label: "Settings",
         command: "tabby.openSettings",
+        iconPath: new ThemeIcon("gear"),
```
icycodes, line 72: I prefer use icon `settings` than `gear` as VSCode settings page use this icon.

icycodes (approved):

PR #3507, opened 2024-12-04
feat(VSCode): navigate to format `output` log channel

Sma1lboy:
Depends on #3504 

Now we can navigate to specific channel by click the reference item on chat panel

### demo 

https://jam.dev/c/7c970cc9-f958-4725-ac3f-585a1e7615d1

### DIFF
--- a/clients/vscode/src/chat/fileContext.ts
+++ b/clients/vscode/src/chat/fileContext.ts
@@ -1,7 +1,7 @@
 import type { TextEditor, TextDocument } from "vscode";
 import type { FileContext } from "tabby-chat-panel";
 import type { GitProvider } from "../git/GitProvider";
-import { workspace, window, Position, Range, Selection, TextEditorRevealType, Uri, ViewColumn } from "vscode";
+import { workspace, window, Position, Range, Selection, TextEditorRevealType, Uri, ViewColumn, commands } from "vscode";
 import path from "path";
 import { getLogger } from "../logger";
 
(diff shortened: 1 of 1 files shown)


clients/vscode/src/chat/fileContext.ts
```diff
@@ -67,6 +67,11 @@ export async function showFileContext(fileContext: FileContext, gitProvider: Git
   if (!document) {
     throw new Error(`File not found: ${fileContext.filepath}`);
   }
+  if (document.uri.toString().startsWith("output:")) {
+    const channelId = document.uri.fsPath.toString();
+    await commands.executeCommand(`workbench.action.output.show.${channelId}`);
+    return;
+  }
```
icycodes, line 74: Let's move this block to before Line66, and use `fileContext.filepath` instead of  `document.uri.toString()`, as there is no need to try to resolve and open the document.

icycodes (changes requested):

clients/vscode/src/chat/fileContext.ts
icycodes: `Uri.parse` should be used if needed.

ee/tabby-ui/components/chat/question-answer.tsx
```diff
@@ -385,6 +385,7 @@ function AssistantMessageCard(props: AssistantMessageCardProps) {
           }}
           // When onApplyInEditor is null, it means isInEditor === false, thus there's no need to showExternalLink
           showExternalLink={!!onApplyInEditor}
+          isInEditor={!!onApplyInEditor}
```
icycodes, line 388: This change should not be included in this PR.

icycodes (approved):

PR #3551, opened 2024-12-12
fix(chat-panel): improve the implement of lookup symbol.

icycodes:
Test preview:
https://jam.dev/c/efdbde67-37bb-4427-8349-c7e9d4cacc0d

### DIFF
--- a/clients/tabby-chat-panel/src/index.ts
+++ b/clients/tabby-chat-panel/src/index.ts
@@ -252,10 +252,12 @@ export function createClient(target: HTMLIFrameElement, api: ClientApiMethods): 
       refresh: api.refresh,
       onSubmitMessage: api.onSubmitMessage,
       onApplyInEditor: api.onApplyInEditor,
+      onApplyInEditorV2: api.onApplyInEditorV2,
       onLoaded: api.onLoaded,
       onCopy: api.onCopy,
       onKeyboardEvent: api.onKeyboardEvent,
       lookupSymbol: api.lookupSymbol,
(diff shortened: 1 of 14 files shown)


clients/tabby-chat-panel/src/index.ts
```diff
@@ -84,8 +210,20 @@ export interface ClientApiMethods {
...
+   * @returns The symbol information if found, otherwise undefined.
+   */
+  lookupSymbol?: (symbol: string, hints?: LookupSymbolHint[] | undefined) => Promise<SymbolInfo | undefined>
+
+  /**
+   * Open the target file location in the editor.
+   * @param target The target file location to open.
+   * @returns Whether the file location is opened successfully.
+   */
+  openInEditor: (target: FileLocation) => Promise<boolean>
```
Sma1lboy, line 226: Should this be a method implemented by all clients? Or should it only be implemented by those who have this capability? It seems we already have a 'base' method called 'navigate', and 'base' implies it should be handled by all client IDEs.

Sma1lboy (approved):
Otherwise, LGTM

clients/tabby-chat-panel/src/index.ts
icycodes: Yes, this method is intended to replace the existing `navigate` method.

I will add a new method `openExternal(url: string): Promise<boolean>` next, and replace all use of `navigate` with either `openInEditor` or `openExternal`, then deprecate the `navigate` method.

PR #3562, opened 2024-12-13
fix(graphQL): fix error when retry mark notifications read

zwpaper:
if you try to mark a notification as read when it was marked as read before.

Before:
CleanShot 2024-12-14 at 01 49 29@2x

After:
CleanShot 2024-12-14 at 01 49 45@2x

### DIFF
--- a/ee/tabby-db/src/notifications.rs
+++ b/ee/tabby-db/src/notifications.rs
@@ -65,7 +65,7 @@ impl DbConn {
 
         let query = format!(
             r#"
-INSERT OR IGNORE INTO read_notifications (notification_id, user_id)
+INSERT INTO read_notifications (notification_id, user_id)
 SELECT
     notifications.id,
     ?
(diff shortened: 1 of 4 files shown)


wsxiaoys (approved):

ee/tabby-db/src/notifications.rs
```diff
@@ -33,7 +33,13 @@ impl DbConn {
 
     pub async fn mark_notification_read(&self, id: i64, user_id: i64) -> Result<()> {
         query!(
-            "INSERT INTO read_notifications (notification_id, user_id) VALUES (?, ?)",
+            "INSERT INTO read_notifications (notification_id, user_id)
```
wsxiaoys, line 36: you might use `ON CONFLICT` query to handle this situation:

https://www.sqlite.org/lang_upsert.html

liangfung (approved):

PR #3577, opened 2024-12-17
feat(index): add commit to CodeSearchDocument

zwpaper:
this PR does:

1. check code indexes, and reindex if no commit field
2. index code with commit as a field
3. return commit in codeAttachment

commit in response:
CleanShot 2024-12-17 at 19 59 08@2x

### DIFF
--- a/crates/tabby/src/services/code.rs
+++ b/crates/tabby/src/services/code.rs
@@ -185,6 +185,14 @@ fn create_hit(scores: CodeSearchScores, doc: TantivyDocument) -> CodeSearchHit {
             code::fields::CHUNK_GIT_URL,
         )
         .to_owned(),
+        // commit is introduced in v0.23, but it is also a required field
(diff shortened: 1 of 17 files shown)


crates/tabby/src/services/code.rs
```diff
@@ -185,6 +185,14 @@ fn create_hit(scores: CodeSearchScores, doc: TantivyDocument) -> CodeSearchHit {
             code::fields::CHUNK_GIT_URL,
         )
         .to_owned(),
+        // commit is introduced in v0.23, but it is also a required field
+        // so we need to handle the case where it's not present
+        commit: get_json_text_field_or_default(
```
wsxiaoys, line 190: we can mark `commit` field as `optional` by now, and flip it to required after two release.

wsxiaoys:
Please fix tests

PR #3584, opened 2024-12-18
feat(thread): add `code_source_id` field to `Message` struct and upda…

wsxiaoys:
…te `ThreadServiceImpl`


wsxiaoys:
cc @zwpaper

PR #3602, opened 2024-12-20
feat(vscode): add at feature with AT file implementation in VSCode

Sma1lboy:
### not git repo
image

### in git repo
image


clients/vscode/src/chat/utils.ts
```diff
@@ -221,3 +222,49 @@ export function generateLocalNotebookCellUri(notebook: Uri, handle: number): Uri
...
+  if (filepath.kind === "git") {
+    label = filepath.filepath;
+  } else {
+    const workspaceFolder = workspace.getWorkspaceFolder(uri);
+    if (workspaceFolder) {
+      label = path.relative(workspaceFolder.uri.fsPath, uri.fsPath);
+    } else {
+      label = path.basename(uri.fsPath);
+    }
+  }
```
icycodes, line 239: Please try to use method `workspace.asRelativePath`:
```suggestion
  const label = worksapce.asRelativePath(uri);
```

clients/vscode/src/chat/utils.ts
```diff
@@ -221,3 +222,49 @@ export function generateLocalNotebookCellUri(notebook: Uri, handle: number): Uri
...
+
+  if (typeof range.start === "object" && typeof range.end === "object") {
+    const startPos = new VSCodePosition(range.start.line - 1, range.start.character - 1);
+    const endPos = new VSCodePosition(range.end.line - 1, range.end.character - 1);
+    const selectedRange = new VSCodeRange(startPos, endPos);
+    return document.getText(selectedRange);
+  }
+
+  return "";
+}
```
icycodes, line 270: Can we reuse `chatPanelLocationToVSCodeRange` to convert `range: LineRange | PositionRange` to VSCodeRange?
By reusing thiat, there is no need to extract `extractTextFromRange`.

clients/vscode/src/chat/webview.ts
```diff
@@ -456,6 +462,71 @@ export class ChatWebview {
...
+              .map((tab) => ({
+                uri: (tab.input as TabInputText).uri,
+                label: tab.label,
+              })),
+          );
+
+          this.logger.info(`No query provided, listing ${openTabs.length} opened editors.`);
+
+          return openTabs
+            .filter((tab) => tab.uri.scheme === "file")
```
icycodes, line 484: Why is this filter required?

clients/vscode/src/chat/webview.ts
```diff
@@ -456,6 +462,71 @@ export class ChatWebview {
...
+        if (!query) {
+          // open all opened tabs
+          // TODO: check tab file stat, check if exists
+          const openTabs = window.tabGroups.all.flatMap((group) =>
+            group.tabs
+              .filter((tab) => tab.input instanceof TabInputText)
+              .map((tab) => ({
+                uri: (tab.input as TabInputText).uri,
+                label: tab.label,
+              })),
```
icycodes, line 478: I think this map is not required here. You can directly map it with `uriToListFileItem`.

clients/vscode/src/chat/webview.ts
```diff
@@ -456,6 +462,71 @@ export class ChatWebview {
...
+          );
+
+          this.logger.info(`No query provided, listing ${openTabs.length} opened editors.`);
+
+          return openTabs
+            .filter((tab) => tab.uri.scheme === "file")
+            .map((tab) => uriToListFileItem(tab.uri, this.gitProvider));
+        }
+
+        const globPattern = `**/${query}*`;
```
icycodes, line 488: Should we validate the query string before injecting it into the glob pattern to ensure `**/${query}*` is safe? For example, what if the query string contains `*` or other special characters?

clients/vscode/src/chat/webview.ts
```diff
@@ -456,6 +462,71 @@ export class ChatWebview {
...
+          this.logger.info(`Found ${files.length} files.`);
+          return files.map((uri) => uriToListFileItem(uri, this.gitProvider));
+        } catch (error) {
+          this.logger.warn("Failed to find files:", error);
+          return [];
+        }
+      },
+
+      readFileContent: async (info: FileRange): Promise<string | null> => {
+        if (info.range) {
```
icycodes, line 501: There's no need for an `if` branch here. `document.getText()` can accept a range or `undefined` as parameters.

clients/vscode/src/chat/webview.ts
```diff
@@ -456,6 +462,71 @@ export class ChatWebview {
...
+              this.logger.warn(`Could not resolve URI from filepath: ${JSON.stringify(info.filepath)}`);
+              return null;
+            }
+            const document = await workspace.openTextDocument(uri);
+
+            return extractTextFromRange(document, info.range);
+          } catch (error) {
+            this.logger.error("Failed to get file content by range:", error);
+            return null;
+          }
```
icycodes, line 514: No need to catch error, or just rethrow it.

clients/vscode/src/chat/webview.ts
Sma1lboy: Now that I have applied escapeGlobPattern to the query to prevent pattern matching, there is an issue where I cannot search for a file named `*testfile.txt` using the query `*`.

clients/vscode/src/chat/webview.ts
```diff
@@ -456,6 +462,53 @@ export class ChatWebview {
...
+              }
+              return escapeGlobPattern(char);
+            })
+            .join("");
+
+          const globPattern = `**/${caseInsensitivePattern}*`;
+
+          this.logger.info(`Searching files with pattern: ${globPattern}, limit: ${maxResults}`);
+
+          const files = await workspace.findFiles(globPattern, null, maxResults);
```
icycodes, line 493: Should we set exclude to `undefined` to use default file-excludes?
```suggestion
          const files = await workspace.findFiles(globPattern, undefined, maxResults);
```

icycodes: cc @liangfung

PR #3613, opened 2024-12-24
ci(release): always mark release as prerelease (make latest requires …

wsxiaoys:
…manual ops)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -345,17 +345,6 @@ jobs:
       - name: Display structure of created files
         run: ls -R dist
 
-      - name: Determine is stable release
-        run: |
-          if [[ ${{ github.ref_name }} =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -360,8 +360,8 @@ jobs:
         if: github.event_name == 'push'
         with:
           allowUpdates: true
-          prerelease: ${{ env.STABLE_RELEASE == 'false' }}
```
zwpaper, line 363: should we also drop https://github.com/TabbyML/tabby/pull/3613/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L348-L358 since STABLE_RELEASE is no longer used

zwpaper (changes requested):

.github/workflows/release.yml
wsxiaoys: Done

zwpaper (approved):

PR #3640, opened 2025-01-02
feat(job): db job should retain job_runs and user_events for only last three months

zwpaper:
…last three months

tested locally:

CleanShot 2025-01-02 at 20 13 56@2x

and added tests codes

### DIFF
--- a/ee/tabby-db/src/job_runs.rs
+++ b/ee/tabby-db/src/job_runs.rs
@@ -107,6 +107,19 @@ impl DbConn {
         Ok(num_deleted as usize)
     }
 
+    pub async fn delete_jobs_before(&self, before: DateTime<Utc>) -> Result<usize> {
+        let before = before.as_sqlite_datetime();
+        let num_deleted = query!(
+            "delete FROM job_runs WHERE updated_at < ? AND exit_code IS NOT NULL",
+            before,
+        )
+        .execute(&self.pool)
+        .await?
+        .rows_affected();
+
+        Ok(num_deleted as usize)
+    }
+
     pub async fn list_job_runs_with_filter(
         &self,
         ids: Option<Vec<i32>>,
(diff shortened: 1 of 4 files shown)


ee/tabby-db/src/job_runs.rs
```diff
@@ -107,6 +107,19 @@ impl DbConn {
         Ok(num_deleted as usize)
     }
 
+    pub async fn delete_jobs_before(&self, before: DateTime<Utc>) -> Result<usize> {
```
wsxiaoys, line 110:
```suggestion
    pub async fn delete_job_run_before_three_months(&self) -> Result<usize> {
```

ee/tabby-db/src/user_events.rs
```diff
@@ -73,4 +73,14 @@ impl DbConn {
 
         Ok(events)
     }
+
+    pub async fn delete_user_events_before(&self, before: DateTime<Utc>) -> Result<usize> {
```
wsxiaoys, line 77:
```suggestion
    pub async fn delete_user_events_before_three_months(&self) -> Result<usize> {
```

wsxiaoys (approved):
Otherwise LGTM

ee/tabby-webserver/src/service/background_job/db.rs
```diff
@@ -37,4 +38,158 @@ impl DbMaintainanceJob {
             .await?;
         Ok(())
     }
+
+    pub async fn retention(now: DateTime<Utc>, db: DbConn) -> tabby_schema::Result<()> {
```
wsxiaoys, line 42:
```suggestion
    async fn data_retention(now: DateTime<Utc>, db: DbConn) -> tabby_schema::Result<()> {
```

just call it in `cron`

PR #3646, opened 2025-01-06
fix(ui): set the color of popover in Chat side panel properly

liangfung:
### Preview
image

----

image


icycodes (approved):
Looks great!

PR #3653, opened 2025-01-08
feat(notification): send notification to admin when job failed

zwpaper:
This PR:
1. create a job_run in db before running a background job
2. save background error logs using `logkit::warn`
3. send a notification if the job failed and users could redirect to the job detail page

db failed:
CleanShot 2025-01-13 at 15 23 03@2x

click on return will go back to job list:
CleanShot 2025-01-13 at 15 42 45@2x

CleanShot 2025-01-13 at 21 03 01@2x

when click on return:
https://jam.dev/c/dd698d82-0b48-4bbd-b056-9f8d23f218d3

### DIFF
--- a/ee/tabby-webserver/src/service/background_job/db.rs
+++ b/ee/tabby-webserver/src/service/background_job/db.rs
@@ -65,7 +65,7 @@ impl DbMaintainanceJob {
             logkit::warn!("Failed to run data retention job: {}", e);
         }
 
-        if has_error {
+        if !has_error {
             Ok(())
         } else {
             Err(CoreError::Other(anyhow::anyhow!(
@@ -75,28 +75,29 @@ impl DbMaintainanceJob {
     }
 
     async fn data_retention(now: DateTime<Utc>, db: &DbConn) -> tabby_schema::Result<()> {
-        let mut errors = vec![];
+        let mut has_error = false;
 
         if let Err(e) = db.delete_job_run_before_three_months(now).await {
-            errors.push(format!(
+            has_error = true;
+            logkit::warn!(
                 "Failed to clean up and retain only the last 3 months of jobs: {}",
                 e
-            ));
+            );
         }
 
         if let Err(e) = db.delete_user_events_before_three_months(now).await {
-            errors.push(format!(
+            has_error = true;
+            logkit::warn!(
                 "Failed to clean up and retain only the last 3 months of user events: {}",
                 e
-            ));
+            );
         }
 
-        if errors.is_empty() {
+        if !has_error {
             Ok(())
         } else {
             Err(CoreError::Other(anyhow::anyhow!(
-                "{}",
-                errors.join(";\n\n")
+                "Failed to run data retention job"
             )))
         }
     }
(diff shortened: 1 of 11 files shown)


ee/tabby-webserver/src/service/background_job/mod.rs
```diff
@@ -120,46 +128,74 @@ pub async fn start(
...
+                    return;
                 }
             };
+
+            if !result.is_empty() {
+                notification_service
+                    .create(
+                        NotificationRecipient::Admin,
+                        &format!(
+                            r#"Background job failed
```
wsxiaoys, line 190: Please add link to `/jobs/${id}` page for the failed job.

ee/tabby-webserver/src/service/background_job/mod.rs
```diff
@@ -120,46 +128,74 @@ pub async fn start(
...
+                        },
+                        _ => {
+                            logkit::info!(exit_code = 0; "Job completed successfully");
+                            logger.finalize().await;
+                            vec![]
+                        }
+                    }
                 },
                 Some(now) = hourly.next() => {
+                    let mut errors = vec![];
```
wsxiaoys, line 148: i recommend just send individual notifications, instead of collect errors for each hourly / croin jobs.

ee/tabby-webserver/src/service/background_job/db.rs
```diff
@@ -14,23 +15,32 @@ impl Job for DbMaintainanceJob {
     const NAME: &'static str = "db_maintainance";
 }
 
+macro_rules! append_error {
```
wsxiaoys, line 18: No need to append error anymore, just use logkit::warn

wsxiaoys:
It's not necessary to put any error information in notification, please use following text copy for all background job failed notification:

```
Job ${id} has failed. Please check the log at $link to identify the underlying issue.
```

zwpaper:
There is no error information in the notification for jobs shown in UI.

For the database job, since several tasks are executed independently, I append the task name if it fails and only uses one notification.
and db job is not shown in UI, so I can only tell users to check stderr logs

zwpaper:
@wsxiaoys PR updated, PTAL

ee/tabby-webserver/src/service/background_job/db.rs
```diff
@@ -20,16 +20,19 @@ impl DbMaintainanceJob {
         context: Arc<dyn ContextService>,
         db: DbConn,
     ) -> tabby_schema::Result<()> {
-        let mut errors = vec![];
+        let mut exit_code = 0;
```
wsxiaoys, line 23:
```suggestion
        let mut has_error = false;
```

wsxiaoys (changes requested):

ee/tabby-webserver/src/service/background_job/db.rs
```diff
@@ -44,27 +47,29 @@ impl DbMaintainanceJob {
...
         }
 
         if let Err(e) = Self::data_retention(now, &db).await {
-            errors.push(format!("Failed to run data retention job: {}", e));
+            has_error = true;
+            logkit::warn!("Failed to run data retention job: {}", e);
         }
 
-        if errors.is_empty() {
+        if has_error {
```
wsxiaoys, line 68:
```suggestion
        if !has_error {
```

PR #3710, opened 2025-01-16
fix(chat-panel): update typedef of filepath.

icycodes:
This PR helps to resolve the issue of when the workspace is not a git repository, the file path is displayed as an absolute uri. It would be better to show a relative path to the workspace instead.

Screenshot from 2025-01-20 10-13-01


icycodes:
Closed and moved to #3823

PR #3766, opened 2025-01-27
feat(vscode): support attach file context to inline edit

zhanba:
## Feature
- Support attach file context to inline edit #2964 
- demo: https://jam.dev/c/41e820ac-9f92-4736-b113-e21f8f9a7d39

image

image

image

## Need help

- [ ] how we update prompt template to include file content

### DIFF
--- a/clients/tabby-agent/src/chat/inlineEdit.ts
+++ b/clients/tabby-agent/src/chat/inlineEdit.ts
@@ -22,6 +22,10 @@ import cryptoRandomString from "crypto-random-string";
 import { isEmptyRange } from "../utils/range";
 import { readResponseStream, Edit, applyWorkspaceEdit } from "./utils";
 import { initMutexAbortController, mutexAbortController, resetMutexAbortController } from "./global";
+import { readFile } from "fs-extra";
+import { getLogger } from "../logger";
+
+const logger = getLogger("ChatEditProvider");
 
 export class ChatEditProvider implements Feature {
   private lspConnection: Connection | undefined = undefined;
@@ -170,6 +174,20 @@ export class ChatEditProvider implements Feature {
       }
     }
 
+    // get file context
+    const fileContext =
+      (
+        await Promise.all(
+          (params.context ?? []).map(async (item) => {
+            const content = await readFile(item.path, "utf-8");
+            const fileContent = content.toString().substring(0, config.chat.edit.documentMaxChars);
+            return `filePath: "${item.path}" \n ${fileContent} \n`;
+          }),
+        )
+      ).join("\n") ?? "";
+
+    logger.debug(`fileContext: ${fileContext}`);
+
     const messages: { role: "user"; content: string }[] = [
       {
         role: "user",
@@ -189,6 +207,8 @@ export class ChatEditProvider implements Feature {
                 return userCommand;
               case "{{languageId}}":
                 return document.languageId;
+              case "{{fileContext}}":
+                return fileContext;
               default:
                 return "";
             }
@@ -196,6 +216,7 @@ export class ChatEditProvider implements Feature {
         ),
       },
     ];
+    logger.debug(`messages: ${JSON.stringify(messages)}`);
     const readableStream = await this.tabbyApiClient.fetchChatStream(
       {
         messages,
--- a/clients/tabby-agent/src/chat/inlineEdit.ts
+++ b/clients/tabby-agent/src/chat/inlineEdit.ts
@@ -1,5 +1,5 @@
-import type { Connection, CancellationToken } from "vscode-languageserver";
-import type { TextDocument } from "vscode-languageserver-textdocument";
+import type { Connection, CancellationToken, Range, URI } from "vscode-languageserver";
+import { TextDocument } from "vscode-languageserver-textdocument";
 import type { TextDocuments } from "../lsp/textDocuments";
 import type { Feature } from "../feature";
 import type { Configurations } from "../config";
@@ -17,18 +17,22 @@ import {
   ChatEditMutexError,
   ServerCapabilities,
   ChatEditResolveParams,
+  ClientCapabilities,
+  ReadFileParams,
+  ReadFileRequest,
 } from "../protocol";
 import cryptoRandomString from "crypto-random-string";
 import { isEmptyRange } from "../utils/range";
-import { readResponseStream, Edit, applyWorkspaceEdit } from "./utils";
+import { readResponseStream, Edit, applyWorkspaceEdit, truncateFileContent } from "./utils";
 import { initMutexAbortController, mutexAbortController, resetMutexAbortController } from "./global";
 import { readFile } from "fs-extra";
 import { getLogger } from "../logger";
-
-const logger = getLogger("ChatEditProvider");
+import { isBrowser } from "../env";
 
 export class ChatEditProvider implements Feature {
+  private logger = getLogger("ChatEditProvider");
   private lspConnection: Connection | undefined = undefined;
+  private clientCapabilities: ClientCapabilities | undefined = undefined;
   private currentEdit: Edit | undefined = undefined;
 
   constructor(
@@ -37,8 +41,9 @@ export class ChatEditProvider implements Feature {
     private readonly documents: TextDocuments<TextDocument>,
   ) {}
 
-  initialize(connection: Connection): ServerCapabilities {
+  initialize(connection: Connection, clientCapabilities: ClientCapabilities): ServerCapabilities {
     this.lspConnection = connection;
+    this.clientCapabilities = clientCapabilities;
     connection.onRequest(ChatEditCommandRequest.type, async (params) => {
       return this.provideEditCommands(params);
     });
@@ -96,6 +101,44 @@ export class ChatEditProvider implements Feature {
     return result;
   }
 
+  async fetchFileContent(uri: URI, range?: Range, token?: CancellationToken) {
+    this.logger.trace("Prepare to fetch text content...");
+    let text: string | undefined = undefined;
+    const targetDocument = this.documents.get(uri);
+    if (targetDocument) {
+      this.logger.trace("Fetching text content from synced text document.", {
+        uri: targetDocument.uri,
+        range: range,
+      });
+      text = targetDocument.getText(range);
+      this.logger.trace("Fetched text content from synced text document.", { text });
+    } else if (this.clientCapabilities?.tabby?.workspaceFileSystem) {
+      const params: ReadFileParams = {
+        uri: uri,
+        format: "text",
+        range: range
+          ? {
+              start: { line: range.start.line, character: 0 },
+              end: { line: range.end.line, character: range.end.character },
+            }
+          : undefined,
+      };
+      this.logger.trace("Fetching text content from ReadFileRequest.", { params });
+      const result = await this.lspConnection?.sendRequest(ReadFileRequest.type, params, token);
+      this.logger.trace("Fetched text content from ReadFileRequest.", { result });
+      text = result?.text;
+    } else if (!isBrowser) {
+      try {
+        const content = await readFile(uri, "utf-8");
+        const textDocument = TextDocument.create(uri, "text", 0, content);
+        text = textDocument.getText(range);
+      } catch (error) {
+        this.logger.trace("Failed to fetch text content from file system.", { error });
+      }
+    }
+    return text;
+  }
+
   async provideEdit(params: ChatEditParams, token: CancellationToken): Promise<ChatEditToken | null> {
     if (params.format !== "previewChanges") {
       return null;
@@ -134,7 +177,7 @@ export class ChatEditProvider implements Feature {
     if (mutexAbortController && !mutexAbortController.signal.aborted) {
       throw {
         name: "ChatEditMutexError",
-        message: "Another smart edit is already in progress",
+        message: "Another chat edit is already in progress",
       } as ChatEditMutexError;
     }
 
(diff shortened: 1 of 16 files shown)


clients/tabby-agent/src/chat/inlineEdit.ts
```diff
@@ -189,13 +207,16 @@ export class ChatEditProvider implements Feature {
                 return userCommand;
               case "{{languageId}}":
                 return document.languageId;
+              case "{{fileContext}}":
```
zhanba, line 210: @wsxiaoys we should update prompt template to include `{{fileContext}}` part. Can you guide me on that?

icycodes:
Hi @zhanba, Thank you for the PR!

The current design is good, but I have some concerns.  

### Use the same or seperated quick pick widget 
The current design use the same quick pick widget for input commands and selecting files, but  
1. The quick pick widget auto-filters items by the current value. If we include both command list and file list in one quick pick list, all items need to be marked as `alwaysShow` to be visible, which may not be the ideal behavior of a quick pick widget.
2. The modal stack is not clear if we blend the quick pick widget. When the user triggers file selection, accepting (Enter or Click on item) will complete the file selection and continue the input command, but dismissing (Esc) will stop the input command instead of continue. By use a new quick pick widget for file selection, we can temporarily hide the command quick pick widget and show the file quick pick widget. Once it is hidden (by accept or dismiss), we update and show the command quick pick widget again to continue.

So, I suggest using a new quick pick for file selection with pushing modal stack.

### How to trigger file selection

1. Using `@`:
   This keeps the UX consistent with the chat side panel. But, to achieve this in a quick pick widget, we need some tricky processing. Detecting whether the last character is `@` to trigger should be good enough. However, there are some cons:
     - The `@file` is not highlighted to mark it includes a file context.
     - When the user moves the cursor and is not at the end and inputs something, the `@` is not properly triggered.
2. Adding a button to trigger selecting files:
   This avoids a lot of tricky implementations but is not as convenient as using `@`.
3. Using a more customizable editor for inputting command:
   This should solve these problems in a cleaner way but requires a side view instead of using the quick pick widget.

I suggest keeping plan 1 for now.

### The resolved data interface for the promise of showQuickPick

I think the `@file` should be kept in the command, as it can be part of a command. For example:
A command could be `Fix the use of function foo according to @foo.ts`, where `@foo.ts` is included by selecting a file. Then the promise of showQuickPick should resolve as:

```json
{
   "command": "Fix the use of function foo according to @foo.ts",
   "fileContexts": [{
      "referer": "@foo.ts",
      "uri": "file://path/to/foo.ts"
   }]
}
```

---

Let me know if you have any thoughts about these points.
Also cc @wsxiaoys for design review.

zhanba:
@icycodes thanks for your reply with so many details!

> The quick pick widget auto-filters items by the current value. If we include both command list and file list in one quick pick list, all items need to be marked as alwaysShow to be visible, which may not be the ideal behavior of a quick pick widget.

Use `alwaysShow ` and filter items by ourself is acceptable, the problem is that we can't handle the sort of items, so the order of items may be inconsistent in a few case.

> The modal stack is not clear if we blend the quick pick widget. When the user triggers file selection, accepting (Enter or Click on item) will complete the file selection and continue the input command, but dismissing (Esc) will stop the input command instead of continue. By use a new quick pick widget for file selection, we can temporarily hide the command quick pick widget and show the file quick pick widget. Once it is hidden (by accept or dismiss), we update and show the command quick pick widget again to continue.

I'm not sure if  `esc` on file selection stop input would make real trouble for user. By using two quick pick, we can avoid this problem. My concern for  two quick pick is that this may interrupt the user input and cause some context switch between two input which may confusing user. 

On `How to trigger file selection`,  I'm agree with you. Using `@` is convenient but with some limitation due to the limited capability of  quick pick widget.

On `The resolved data interface for the promise of showQuickPick`, I'm agree with you and I will update command to include the `@file` part.

cc @wsxiaoys

zhanba:
> Hi @zhanba, Thank you for the PR!
> 
> The current design is good, but I have some concerns.
> 
> ### Use the same or seperated quick pick widget
> The current design use the same quick pick widget for input commands and selecting files, but
> 
> 1. The quick pick widget auto-filters items by the current value. If we include both command list and file list in one quick pick list, all items need to be marked as `alwaysShow` to be visible, which may not be the ideal behavior of a quick pick widget.
> 2. The modal stack is not clear if we blend the quick pick widget. When the user triggers file selection, accepting (Enter or Click on item) will complete the file selection and continue the input command, but dismissing (Esc) will stop the input command instead of continue. By use a new quick pick widget for file selection, we can temporarily hide the command quick pick widget and show the file quick pick widget. Once it is hidden (by accept or dismiss), we update and show the command quick pick widget again to continue.
> 
> So, I suggest using a new quick pick for file selection with pushing modal stack.
> 
> ### How to trigger file selection
> 1. Using `@`:
>    This keeps the UX consistent with the chat side panel. But, to achieve this in a quick pick widget, we need some tricky processing. Detecting whether the last character is `@` to trigger should be good enough. However, there are some cons:
>    
>    * The `@file` is not highlighted to mark it includes a file context.
>    * When the user moves the cursor and is not at the end and inputs something, the `@` is not properly triggered.
> 2. Adding a button to trigger selecting files:
>    This avoids a lot of tricky implementations but is not as convenient as using `@`.
> 3. Using a more customizable editor for inputting command:
>    This should solve these problems in a cleaner way but requires a side view instead of using the quick pick widget.
> 
> I suggest keeping plan 1 for now.
> 
> ### The resolved data interface for the promise of showQuickPick
> I think the `@file` should be kept in the command, as it can be part of a command. For example: A command could be `Fix the use of function foo according to @foo.ts`, where `@foo.ts` is included by selecting a file. Then the promise of showQuickPick should resolve as:
> 
> ```json
> {
>    "command": "Fix the use of function foo according to @foo.ts",
>    "fileContexts": [{
>       "referer": "@foo.ts",
>       "uri": "file://path/to/foo.ts"
>    }]
> }
> ```
> 
> Let me know if you have any thoughts about these points. Also cc @wsxiaoys for design review.

I will try to use two quick pick widget and see how it works.

zhanba:
two quick pick demo: 
https://jam.dev/c/ce025a95-397d-47f4-8401-d9c41e56f6b8

icycodes:
> two quick pick demo: https://jam.dev/c/ce025a95-397d-47f4-8401-d9c41e56f6b8

Thank you for the quick update. This demo aligns with my thoughts, it looks much better to me.

zhanba:
add smooth file deletion:
https://jam.dev/c/41e820ac-9f92-4736-b113-e21f8f9a7d39

clients/vscode/src/inline-edit/quickPick.ts
```diff
@@ -0,0 +1,321 @@
...
+    return fileList.map((file) => workspace.asRelativePath(file));
+  }
+
+  private findFiles = wrapCancelableFunction(
+    findFiles,
+    (args) => args[1]?.token,
+    (args, token) => [args[0], { ...args[1], token }] as Parameters<typeof findFiles>,
+  );
+
+  private async searchFilelist(query: string) {
```
icycodes, line 277: I think the item list for file quick pick should always return:

- list of current open editors
- separator
- findFile search result (filter out which is in list of open editors)

icycodes, line 277: The different between funciton `searchFilelist` and `searchFileList` is only a `L`. So give it another name.
```suggestion
  private async getFileList(query: string) {
```

clients/vscode/src/inline-edit/util.ts
```diff
@@ -0,0 +1,75 @@
...
+  mentionQuery?: string;
+}
+
+export interface InlineEditCommand extends InlineEditParseResult {
+  command: string;
+}
+
+export const parseUserCommand = (input: string): InlineEditParseResult => {
+  const mentions: string[] = [];
+  const regex = /(?<=\s|^)@(\S*)/g;
```
icycodes, line 21: Could you try to use `\b` `\B`? I think it should be better than `\s` and `\S`.

clients/vscode/src/inline-edit/util.ts
```diff
@@ -0,0 +1,75 @@
...
+   * for `explain @`, mentionQuery is `''`,  we can trigger file pick
+   * for `explain @file`, mentionQuery is `file`,  we know user is editing the mention
+   * for `explain @file to me`, mentionQuery is `undefined`
+   */
+  mentionQuery?: string;
+}
+
+export interface InlineEditCommand extends InlineEditParseResult {
+  command: string;
+}
```
icycodes, line 17: If we save a map of label to file uri in UserCommandQuickpick scope, we will not parse the command again to get the context list.
```suggestion
export interface InlineEditCommand {
  command: string;
  context: {
    referer: string,
    uri: Uri,
  }[];
}
```

And I think `parseUserCommand` could only handle the case of the trailing query, so I could return single string instead of InlineEditParseResult.

clients/vscode/src/inline-edit/quickPick.ts
zhanba: If user has many open editors, it would be hard to select file in search result part quickly. So I think it would be better to filter the open editors by user query too.

clients/vscode/src/inline-edit/util.ts
zhanba: A map of label to file can't make this `parseUserCommand` just return string, because user may delete file mentions in  command.

clients/vscode/src/inline-edit/quickPick.ts
zhanba: In fact, when user input query, the order of items are controlled by quick pick itself. Even if I always return open editors, the quick pick do filter and reorder of items.

image

image

clients/vscode/src/inline-edit/index.ts
```diff
@@ -1,26 +1,21 @@
...
 import { getLogger } from "../logger";
+import { InlineEditCommand } from "./util";
+import { UserCommandQuickpick } from "./quickPick";
 
 export class InlineEditController {
   private readonly logger = getLogger("InlineEditController");
   private readonly editLocation: Location;
 
+  private get workspaceFolder() {
+    return workspace.getWorkspaceFolder(this.editor.document.uri);
```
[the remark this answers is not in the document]
zhanba: - add a map of label to file uri to resolve context
- history command also store file context info

clients/vscode/src/inline-edit/util.ts
zhanba: - added a map of label to file, but still need to do the `parseUserCommand`

clients/vscode/src/inline-edit/quickPick.ts
zhanba: - Data fetch logic have changed to always push open editors + sep + search files
- The quick pick behavior does not change

clients/tabby-agent/src/chat/inlineEdit.ts
```diff
@@ -92,6 +100,45 @@ export class ChatEditProvider implements Feature {
...
+        format: "text",
+        range: range
+          ? {
+              start: { line: range.start.line, character: 0 },
+              end: { line: range.end.line, character: range.end.character },
+            }
+          : undefined,
+      };
+      this.logger.trace("Fetching text content from ReadFileRequest.", { params });
+      const result = await this.lspConnection?.sendRequest(ReadFileRequest.type, params, token);
```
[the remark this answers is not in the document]
zhanba: Add missing `cancellationToken` and it work fine, although  `cancellationToken` is optional.

clients/vscode/src/inline-edit/quickPick.ts
```diff
@@ -0,0 +1,410 @@
...
+      const file = list.find((i) => i.label === label);
+      if (file === undefined) {
+        list.push({
+          label: label,
+          uri: item.toString(),
+        });
+      } else {
+        /**
+         * In multi-root workspace, different files may have same relative path. In this case we direct use their absolute path as label.
+         */
```
icycodes, line 362: I believe `workspace.asRelativePath` is safe to get unique labels for files in different workspace folders. This is vscode api document for `workspace.asRelativePath` :
```
		/**
		 * Returns a path that is relative to the workspace folder or folders.
		 *
		 * When there are no {@link workspace.workspaceFolders workspace folders} or when the path
		 * is not contained in them, the input is returned.
		 *
		 * @param pathOrUri A path or uri. When a uri is given its {@link Uri.fsPath fsPath} is used.
		 * @param includeWorkspaceFolder When `true` and when the given path is contained inside a
		 * workspace folder the name of the workspace is prepended. Defaults to `true` when there are
		 * multiple workspace folders and `false` otherwise.
		 * @returns A path relative to the root or the input.
		 */
```

clients/tabby-agent/src/chat/inlineEdit.ts
```diff
@@ -92,6 +100,45 @@ export class ChatEditProvider implements Feature {
...
+      try {
+        const content = await readFile(uri, "utf-8");
+        const textDocument = TextDocument.create(uri, "text", 0, content);
+        textDocument;
+        text = textDocument.getText(range);
+      } catch (error) {
+        this.logger.trace("Failed to fetch text content from file system.", { error });
+      }
+    }
+    return text;
```
icycodes, line 139: Please add a `if (!isBrowser)` guard to use fs.

clients/tabby-agent/src/protocol.ts
```diff
@@ -439,8 +439,34 @@ export type ChatEditParams = {
...
+export interface ChatEditFileContext {
+  /**
+   * The symbol in the user command that refer to this file context.
+   */
+  referer: string;
+
+  /**
+   * The uri of the file.
+   */
+  path: string;
```
icycodes, line 461: I perfer to rename it to `uri: URI`
```suggestion
  uri: URI;
```

icycodes:
> A map of label to file can't make this parseUserCommand just return string, because user may delete file mentions in command.

Before resolving the InlineEditComment, we can filter the context list by checking if the referer is still a substring included in the command.
This is just my idea. The current implementaion is also good and acceptable.

icycodes:
Otherwise LGTM.  Please rebase on main and resolve confilicts.

clients/vscode/src/inline-edit/index.ts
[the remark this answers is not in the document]
icycodes: OK, I want it to be looks like the vscode `Go to File` (Ctrl+P) picker.  
But it seems that we cannot implement it via the current quickpick api?

Screenshot from 2025-02-11 18-00-41

clients/vscode/src/inline-edit/quickPick.ts
icycodes: OK, I want it to be looks like the vscode `Go to File` (Ctrl+P) picker.  
But it seems that we cannot implement it via the current quickpick api?

Screenshot from 2025-02-11 18-00-41

zhanba: Found a related issue: https://github.com/microsoft/vscode/issues/73904 , and some workaround...
```ts
(this.quickPick as any).sortByLabel = false;
```

image

zhanba:
> Otherwise LGTM. Please rebase on main and resolve confilicts.

Rebased and Fixed

icycodes:
The next task is to update the chat prompt template to include file context.   
I think it can be done in a separate PR.  
cc @wsxiaoys

PR #3791, opened 2025-02-05
chore: log current models upon startup

zwpaper:
- local
  CleanShot 2025-02-05 at 21 07 13@2x
- remote
  CleanShot 2025-02-05 at 21 08 07@2x
- completion model not set
  CleanShot 2025-02-05 at 21 10 03@2x
- Chat model not set
  CleanShot 2025-02-05 at 21 11 09@2x
- with overwrite models
  CleanShot 2025-02-05 at 21 12 45@2x


wsxiaoys:
image

Instead of adding it to log (for remote model), shall we just added it to `System` tab?

PR #3805, opened 2025-02-07
fix(embedding): add retry to fix llama.cpp request got connect reset …

zwpaper:
…when startup

tested locally without initial errors:
CleanShot 2025-02-07 at 15 09 16@2x

### DIFF
--- a/crates/http-api-bindings/src/embedding/llama.rs
+++ b/crates/http-api-bindings/src/embedding/llama.rs
@@ -2,6 +2,10 @@ use anyhow::anyhow;
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use tabby_inference::Embedding;
+use tokio_retry::{
+    strategy::{jitter, ExponentialBackoff},
+    RetryIf,
+};
 use tracing::Instrument;
 
 use crate::{create_reqwest_client, embedding_info_span};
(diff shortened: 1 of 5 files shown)


crates/http-api-bindings/src/embedding/llama.rs
```diff
@@ -46,10 +46,45 @@ impl Embedding for LlamaCppEngine {
...
-            .await?;
+        // Some initial requests to llama.cpp are experiencing issues
+        // would failed with `Connection reset by peer` or `Broken pipe`
+        //
+        // This serves as a temporary solution to attempt the request up to three times.
+        //
+        // Track issue: https://github.com/ggerganov/llama.cpp/issues/11411
+        let mut attempts = 0;
+        let max_attempts = 3;
+        let response = loop {
```
wsxiaoys, line 57: use `ExponentialBackoff` from `tokio_retry`.

wsxiaoys (commented):
<s>Consider handle embedding  input clipping in this PR as well.</s>

crates/http-api-bindings/src/embedding/llama.rs
wsxiaoys: If we're sure that this situation only happens in server start, consider create a timer to disable the retry behavior after server has been up for 2 hours.

zwpaper: No, it occurs consistently after a period without any embedding requests, let me confirm this again later

crates/http-api-bindings/src/embedding/llama.rs
```diff
@@ -37,19 +41,39 @@ struct EmbeddingResponse {
 #[async_trait]
 impl Embedding for LlamaCppEngine {
     async fn embed(&self, prompt: &str) -> anyhow::Result<Vec<f32>> {
-        let request = EmbeddingRequest {
-            content: prompt.to_owned(),
-        };
+        // Some initial requests to llama.cpp are experiencing issues
```
wsxiaoys, line 44:
```suggestion
        // Some initial requests to llama.cpp are experiencing issues
```

Please refine the docs here to reflect the fact that not only "initial request", but from time to time whenever embedding server being vacant for a while.

wsxiaoys (approved):
Otherwise LGTM

PR #3812, opened 2025-02-09
feat(index): add GDScript language support.

faceCutWall:
There seems to be an annoying dependency problem appeared, [tree-sitter-gdscript](https://crates.io/crates/tree-sitter-gdscript) only release a crate that uses `LanguageFn` supported above [v0.23.0](https://github.com/tree-sitter/tree-sitter/releases/tag/v0.23.0) of tree-sitter. The older version however depends on ~0.20.10, which is incompatible with Tabby. I forked [tree-sitter-gdscript](https://github.com/faceCutWall/tree-sitter-gdscript) and bumped tree-sitter to v0.22.6 to make it work. The functionality has been tested, but I'm not sure if it's okay to do so. Looking forward to more suggestions :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5489,6 +5489,7 @@ dependencies = [
  "tree-sitter-c-sharp",
  "tree-sitter-cpp",
  "tree-sitter-elixir",
+ "tree-sitter-gdscript",
  "tree-sitter-go",
  "tree-sitter-java",
  "tree-sitter-kotlin",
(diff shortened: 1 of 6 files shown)


wsxiaoys:
Thanks for the PR.

I opt to remain with an older version of treesitter to ensure optimal compatibility with language-specific Tlibs. Therefore, a fork of gdscript that introduces support for legacy versions is a welcome addition.

wsxiaoys:
cc @lizzieshinkicker

Issue #3694, opened 2025-01-14
Check whether avx2 instruction set is supported for linux binary


wsxiaoys:
**Please describe the feature you want**
Before running llama-server binary.

**Additional context**
Code location: https://github.com/TabbyML/tabby/blob/5aa27b590e23e905813e26c41f3a015cd474a70b/crates/llama-cpp-server/src/supervisor.rs#L32

---
Please reply with a 👍 if you want this feature.

Evalir:
i'd like to take this (replied). What would be the outcome of the check here? any additional info helps :)

wsxiaoys:
> i'd like to take this (replied). What would be the outcome of the check here? any additional info helps :)

On Linux build - when avx2 instruction is not supported, generated a warning messages saying llama-server would likely fail with default binary distribution

Evalir:
awesome. i'll get to it this weekend!

haro3535:
> > i'd like to take this (replied). What would be the outcome of the check here? any additional info helps :)
> 
> On Linux build - when avx2 instruction is not supported, generated a warning messages saying llama-server would likely fail with default binary distribution

Do you think this problem is caused by the CPU or from Linux?

Sma1lboy:
Fixed in #3814

Issue #3512, opened 2024-12-05
WARN llama-server <chat> exited with status code 1


Mte90:
**Describe the bug**
I am trying to run tabby but I get:
```
WARN llama_cpp_server::supervisor: crates/llama-cpp-server/src/supervisor.rs:98: llama-server <chat> exited with status code 1, args: `Command { std: "//tabby_x86_64-manylinux2014-cuda122/llama-server" "-m" "/home/mte90/.tabby/models/TabbyML/Mistral-7B/ggml/model-00001-of-00001.gguf" "--cont-batching" "--port" "30892" "-np" "1" "--log-disable" "--ctx-size" "4096" "-ngl" "9999" "--chat-template" "<s>{% for message in messages %}{% if (message[\'role\'] == \'user\') != (loop.index0 % 2 == 0) %}{{ raise_exception(\'Conversation roles must alternate user/assistant/user/assistant/...\') }}{% endif %}{% if message[\'role\'] == \'user\' %}{{ \'[INST] \' + message[\'content\'] + \' [/INST]\' }}{% elif message[\'role\'] == \'assistant\' %}{{ message[\'content\'] + \'</s> \' }}{% else %}{{ raise_exception(\'Only user and assistant roles are supported!\') }}{% endif %}{% endfor %}", kill_on_drop: true }`
```

**Information about your version**
0.21

Ideally when there is this output tabby should exit and not still trying to start and generate the command to try with the same parameter or atleast a log output so it is possible to investigate as copy and paste that output doesn't work as it is escaped.

gtozzi:
Same here. Also tried 0.20

JamesNewton:
Same. Version 0.23.

wsxiaoys:
it's likely the CPU of the host doesn't support avx2 - thus cause the issue. We shall at least provide proper error log for such case.

Filing https://github.com/TabbyML/tabby/issues/3694

JamesNewton:
> it's likely the CPU of the host doesn't support avx2 - thus cause the issue. We shall at least provide proper error log for such case.

My system has Intel© Core™ i7-4700HQ which Intel says _does_ have the avx2 feature:
https://www.intel.com/content/www/us/en/products/sku/75116/intel-core-i74700hq-processor-6m-cache-up-to-3-40-ghz/specifications.html

Note that this error only appears for me if I try to use the `-device vulkan` option. It works fine without that; on the CPU only.

zwpaper:
[...]

Issue #3056, opened 2024-09-03
llama-server <embedding> exited with status code -1


Gnomesenpai:
**Describe the bug**
llama-server <embedding> exited with status code -1

**Information about your version**
Unable to get version as it will not start. Docker image used:
REPOSITORY                        TAG                                IMAGE ID       CREATED         SIZE
tabbyml/tabby                     latest                             bc5a49b31c6f   6 days ago      2.64GB

**Information about your GPU**
Tue Sep  3 20:27:02 2024
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.54.14              Driver Version: 550.54.14      CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  Tesla P4                       On  |   00000000:13:00.0 Off |                  Off |
| N/A   67C    P0             25W /   75W |     795MiB /   8192MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

**Additional context**
I removed my old Tabby setup and pulled the new container and set a new data folder however it fails with the error:

Starting...2024-09-03T19:25:59.506655Z  WARN llama_cpp_server::supervisor: crates/llama-cpp-server/src/supervisor.rs:98: llama-server <embedding> exited with status code -1, args: `Command { std: "/opt/tabby/bin/llama-server" "-m" "/data/models/TabbyML/Nomic-Embed-Text/ggml/model.gguf" "--cont-batching" "--port" "30888" "-np" "1" "--log-disable" "--ctx-size" "4096" "-ngl" "9999" "--embedding" "--ubatch-size" "4096", kill_on_drop: true }`

wsxiaoys:
Hi - could you also share the command being used to start docker container?

Gnomesenpai:
Sure, its the one in the getting started guide just with a different file path.

`ersion: '3.5'

services:
  tabby:
    restart: always
    image: tabbyml/tabby
    command: serve --model StarCoder-1B --chat-model Qwen2-1.5B-Instruct --device cuda
    volumes:
      - "./data:/data"
    ports:
      - 8080:8080
    deploy:
      resources:
[...]

PR #3814, opened 2025-02-09
feat(llama-cpp-server): implement logging for stderr output with buff…

Sma1lboy:
- provide more detail information when llama cpp server cannot be pull up

cudaMalloc fail
image

### Relate Issues
#3694 
#3512
#3056


crates/llama-cpp-server/src/supervisor.rs
```diff
@@ -145,6 +190,23 @@ impl LlamaCppSupervisor {
...
+        return Some(String::from(
+            "CUDA memory allocation error detected:\n\
+             1. Try using a smaller Model\n\
+             2. Try to reduce GPU memory usage\n",
+        ));
+    }
+    if error_message.contains("Illegal instruction") {
+        return Some(String::from(
+            "Illegal instruction detected: Current binary requires AVX2 instruction set.\n\
+             Solution: Compile llama.cpp with appropriate CPU instruction set flags for your hardware."
```
wsxiaoys, line 204: suggest user acquire binary from https://github.com/ggml-org/llama.cpp/releases directly

crates/llama-cpp-server/src/supervisor.rs
```diff
@@ -145,6 +190,23 @@ impl LlamaCppSupervisor {
...
 
+fn analyze_error_message(error_message: &str) -> Option<String> {
+    if error_message.contains("cudaMalloc") {
+        return Some(String::from(
+            "CUDA memory allocation error detected:\n\
+             1. Try using a smaller Model\n\
+             2. Try to reduce GPU memory usage\n",
+        ));
+    }
+    if error_message.contains("Illegal instruction") {
```
wsxiaoys, line 201: also need do a detection from user system to ensure avx2 is not supported.

crates/llama-cpp-server/src/supervisor.rs
```diff
@@ -99,18 +104,58 @@ impl LlamaCppSupervisor {
...
+
+                    eprintln!("{}", error_message);
+                    eprintln!(
+                        "llama-server <{}> encountered a fatal error. Exiting service. Please check the above logs for details.",
+                        name
+                    );
+
+                    match status_code {
+                        0 => (),
+                        1 | -1 => {
```
wsxiaoys, line 141: add a colorful section (tabby_common::terminal)

crates/llama-cpp-server/src/supervisor.rs
```diff
@@ -87,30 +92,64 @@ impl LlamaCppSupervisor {
...
-                        .map(tokio::io::BufReader::new)
-                        .map(|reader| reader.lines())
-                        .expect("Failed to read stderr");
-
-                    while let Ok(Some(line)) = stderr.next_line().await {
-                        warn!("<{}>: {}", name, line);
-                    }
 
-                    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+                    let error_message = error_lines.into_iter().fold(
```
zwpaper, line 118: print the error message directly later instead of accumulating it into a string?

crates/llama-cpp-server/src/supervisor.rs
```diff
@@ -87,30 +92,64 @@ impl LlamaCppSupervisor {
...
+                                msg.print();
+                            }
+                            eprintln!(
+                                "llama-server <{}> encountered a fatal error. Exiting service. Please check the above logs and suggested solutions for details.",
+                                name
+                            );
+                            std::process::exit(1);
+                        }
+                        _ => {
+                            warn!(
```
zwpaper, line 146: how about printing the first or some of the retried error messages, so we can gather unknown error messages when users experience them or if they are caught in a retry loop.

WDYT @wsxiaoys

crates/llama-cpp-server/src/supervisor.rs
```diff
@@ -87,30 +92,64 @@ impl LlamaCppSupervisor {
...
+                            acc.push_str(&format!("{}\n", line));
+                            acc
+                        },
+                    );
+
+                    match status_code {
+                        0 => (),
+                        1 | -1 => {
+                            eprintln!("{}", error_message);
+                            if let Some(solution) = analyze_error_message(&error_message) {
```
zwpaper, line 130: Analyze the error message while printing it line by line. If a solution is detected, we can halt the analysis directly.

then we also don't have to split the solution into lines later

crates/llama-cpp-server/src/supervisor.rs
```diff
@@ -87,30 +92,64 @@ impl LlamaCppSupervisor {
...
-                        .map(|reader| reader.lines())
-                        .expect("Failed to read stderr");
-
-                    while let Ok(Some(line)) = stderr.next_line().await {
-                        warn!("<{}>: {}", name, line);
-                    }
 
-                    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+                    let error_message = error_lines.into_iter().fold(
+                        String::from("Recent llama-cpp errors:\n"),
```
zwpaper, line 119: we could also print this line with color

zwpaper (changes requested):
Could you also attempt to perform an indexing operation and evaluate whether there is any difference in performance between having logging disabled and the present handling method?

crates/llama-cpp-server/src/supervisor.rs
wsxiaoys: SGTM

crates/llama-cpp-server/src/supervisor.rs
Sma1lboy: I'm not sure if I don't split the solution how could I fit into the InfoMessage new arguments? it's require a lines: &'a [&'a str]

Can I adding a new constructor

PR #3840, opened 2025-02-14
feat(agent): include file context in inline edit prompt.

icycodes:


clients/tabby-agent/src/chat/prompts/include-file-context-item.md
```diff
@@ -0,0 +1,2 @@
+title="{{filepath}}"
+{{content}}
```
zhanba, line 2: - Should the prompt also include the `referer` property? 
- Should we put file content in XML tag like `<CONTEXTDOCUMENT>` to prevent file content pollute the prompt?

clients/tabby-agent/src/chat/inlineEdit.ts
```diff
@@ -217,53 +218,46 @@ export class ChatEditProvider implements Feature {
...
+              content: fileContent,
+            });
           }),
         )
       )
         .filter((item): item is string => item !== undefined)
         .join("\n") ?? "";
 
-    this.logger.debug(`fileContext: ${fileContext}`);
+    const fileContext = formatPlaceholders(fileContextListTemplate, {
```
zhanba, line 241: If there are no valid file context, we could just insert empty string as `fileContext`  to the prompt?

clients/tabby-agent/src/chat/prompts/include-file-context-item.md
icycodes: Fixed

clients/tabby-agent/src/chat/inlineEdit.ts
icycodes: Fixed

PR #3851, opened 2025-02-16
ci(docker): add vulkan image

jjm2473:
supports x86_64 and arm64

### DIFF
--- a/.github/workflows/docker-vulkan.yml
+++ b/.github/workflows/docker-vulkan.yml
@@ -1,0 +1,149 @@
+name: Create and publish docker image (Vulkan)
+
+on:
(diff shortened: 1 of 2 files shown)


wsxiaoys:
Hi - can you share an use case about vulkan docker image? AFAIK, docker support for vulkan has been problematic, thus we recommend binary installation for utilizing vulkan.

jjm2473:
I want to run tabby on ARM platform with libmali-vulkan. There is not any docker image or binary provided. This PR create a base vulkan-based image, so ARM platforms can build image on top of it.

jjm2473:
reopen

PR #3852, opened 2025-02-17
refactor(chat): simplify attachment client code structure

Sma1lboy:
- Remove endline field
- Remove baseDir and instead make all URIs use relative paths when looking up symbols

### DIFF
--- a/ee/tabby-ui/components/message-markdown/index.tsx
+++ b/ee/tabby-ui/components/message-markdown/index.tsx
@@ -226,7 +226,7 @@ export function MessageMarkdown({
           ? {
               kind: 'workspace',
               filepath: code.filepath,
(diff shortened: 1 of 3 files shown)


ee/tabby-ui/components/message-markdown/index.tsx
```diff
@@ -213,35 +213,23 @@ export function MessageMarkdown({
...
               uri: code.filepath
             },
         location:
-          code.startLine && code.endLine
+          code.startLine && code.content
             ? {
                 start: code.startLine,
-                end: code.endLine
+                end: code.startLine + code.content.split('\n').length - 1
               }
```
liangfung, line 233: Consider using the existing method `getRangeFromAttachmentCode`

liangfung (approved):

PR #3878, opened 2025-02-19
fix(agent): skip computing char diff for edit preview that just an insertion/deletion.

icycodes:


zhanba:
> It seems that `DefaultLinesDiffComputer.computeDiff` does not accept empty array input. 
The diff algorithm does not handle the empty string case, as in the VSCode scenario, the code is guaranteed to have at least one line. For Tabby, skipping the diff is reasonable. I will add a high level wrapper in codiff to handle this case lately.

PR #3903, opened 2025-02-26
feat(rate_limit): increase user request limit from 60 to 120 per minute

Sma1lboy:
increase limit from 60 to 120 per min

### DIFF
--- a/ee/tabby-webserver/src/rate_limit.rs
+++ b/ee/tabby-webserver/src/rate_limit.rs
@@ -8,7 +8,7 @@ pub struct UserRateLimiter {
     rate_limiters: Mutex<TimedCache<String, ratelimit::Ratelimiter>>,
 }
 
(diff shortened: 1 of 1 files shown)


ee/tabby-webserver/src/rate_limit.rs
```diff
@@ -8,7 +8,7 @@ pub struct UserRateLimiter {
     rate_limiters: Mutex<TimedCache<String, ratelimit::Ratelimiter>>,
 }
 
-static USER_REQUEST_LIMIT_PER_MINUTE: u64 = 60;
+static USER_REQUEST_LIMIT_PER_MINUTE: u64 = 120;
```
zwpaper, line 11: please also change L16

zwpaper (changes requested):

zwpaper (approved):

PR #3909, opened 2025-02-26
fix(webserver): changed ldap search from onelevel to subtree

ifelsefi:
OneLevel does not do a subtree search which is required for effective use of directory server.  See [ldap3 docs](https://docs.rs/ldap3/latest/ldap3/#synchronous-search).


ifelsefi:
@wsxiaoys please review, thanks!

PR #3943, opened 2025-03-04
feat(context_providers): support indexing commits for git provider

zwpaper:
1. did not halt indexing commits even if a commit index already existed, as there may be fewer than 100 commits when a build previously failed. We should rebuild to achieve the requisite 100 commits.
2. May necessitate a refactor of persync, permitting diverse kinds of structured_doc to determine when to skip indexing.

## Tests
CleanShot 2025-03-04 at 22 11 28@2x


ee/tabby-webserver/src/service/background_job/commit_index.rs
wsxiaoys, line 1: index_commits.rs

wsxiaoys (approved):
Otherwise LGTM

PR #3988, opened 2025-03-12
feat(chat-panel): add getChanges API method and related interface

Sma1lboy:
This interface can help the server obtain the git diff changes from the client side


icycodes (approved):

clients/tabby-chat-panel/src/client.ts
```diff
@@ -229,3 +236,27 @@ export interface ListSymbolItem {
...
+/**
+ * Parameters for retrieving Git changes from the client IDE.
+ * Used with {@link ClientApi.getChanges} method.
+ */
+export interface GetChangesParams {
+  /**
+   * The maximum number of change items to return.
+   * If not provided, all changes will be returned.
+   */
+  limit?: number
```
icycodes, line 249: This limit applies to the number of file diffs.  
I think it should be better to limit the number of characters directly, as our target is to control the token count.

clients/tabby-chat-panel/src/client.ts
```diff
@@ -229,3 +236,27 @@ export interface ListSymbolItem {
...
+/**
+ * Represents a Git change item returned by the {@link ClientApi.getChanges} method.
+ * Contains information about modifications detected in the Git workspace.
+ */
+export interface ChangeItem {
+  /**
+   * The content of the change, formatted in the same Diff format as the `git diff` command.
+   * This includes information about added, modified, and deleted lines in the affected files.
+   */
+  content: string
```
icycodes, line 261: It would be good to add `staged: boolean` here. See https://github.com/TabbyML/tabby/pull/4023/files#r2000678893

clients/tabby-chat-panel/src/client.ts
```diff
@@ -229,3 +236,31 @@ export interface ListSymbolItem {
...
+/**
+ * Parameters for retrieving Git changes from the client IDE.
+ * Used with {@link ClientApi.getChanges} method.
+ */
+export interface GetChangesParams {
+  /**
+   * The maximum number of tokens to return in the change content.
+   * If not provided, all changes will be returned without token limitation.
+   */
+  limitToken?: number
```
icycodes, line 249:
```suggestion
  maxChars?: number
```
This parameter is used to limit the token count. However, the client cannot calculate the token count accurately, and approximate token count calculations should not be performed on the client side either.  
I prefer using `maxChars` here as it provides a precise definition.

icycodes (approved):

PR #3996, opened 2025-03-13
feat(admin): add security option to hide password login in frontend

liangfung:
### 1. Add security option to hide password login in frontend
image

If password login is hidden and LDAP is configured, then only LDAP login is displayed.
image

### 2. Show password with searchParams.
When password login is disabled in settings, it can be forcibly displayed by using the search parameter `passwordSignIn=true`.

image


.changes/unreleased/Features-20250318-131219.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Features
+body: Add security option to hide password login in frontend. When enabled, the password login can be revealed by appending the URL search parameter `password_sign_in=true`.
```
wsxiaoys, line 2:
```suggestion
body: Add security option to hide password login in frontend. When enabled, the password login can be revealed by appending the URL search parameter `passwordSignIn=true`.
```

wsxiaoys (commented):
otherwise LGTM

PR #4000, opened 2025-03-14
chore: return section id before attachment when append section

zwpaper:
appended section:

CleanShot 2025-03-15 at 02 15 34@2x

### DIFF
--- a/ee/tabby-webserver/src/service/page.rs
+++ b/ee/tabby-webserver/src/service/page.rs
@@ -144,6 +144,9 @@ impl PageService for PageServiceImpl {
             let section = db
                 .create_page_section(page_id, &new_section_title)
                 .await?;
+            let section_id = section.id;
+            let page_section: PageSection = section.into();
+            yield Ok(SectionRunItem::PageSectionCreated(page_section));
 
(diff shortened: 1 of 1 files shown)


ee/tabby-webserver/src/service/page.rs
```diff
@@ -144,6 +144,9 @@ impl PageService for PageServiceImpl {
             let section = db
                 .create_page_section(page_id, &new_section_title)
                 .await?;
+            let section_id = section.id;
+            let page_section: PageSection = section.into();
+            yield Ok(SectionRunItem::PageSectionCreated(page_section));
```
wsxiaoys, line 149: https://demo.tabbyml.com/files/github/TabbyML/tabby/-/blob/main/ee/tabby-schema/src/schema/page/types.rs?plain=1#L242

To confirm: is the order of yielding matching its definition in `types.rs` after this PR?

wsxiaoys (approved):

ee/tabby-webserver/src/service/page.rs
zwpaper: yes

PR #4023, opened 2025-03-18
feat(VSCode): implement getChanges API method to retrieve changed files in workspace repositories

Sma1lboy:
VSCode implementation
using gitProvider get  all  changes in current workspace/repositories
image


clients/vscode/src/chat/webview.ts
```diff
@@ -725,6 +727,45 @@ export class ChatWebview extends EventEmitter {
...
+            }
+          }
+          return res;
+        };
+        // get staged first
+        const res: ChangeItem[] = await getRepoChanges(repositories, true, remain);
+        remain -= res.length;
+        // if there are still some left, get unstaged
+        res.push(...(await getRepoChanges(repositories, false, remain)));
+        this.logger.info(`Found ${res.length} changed files.`);
```
icycodes, line 764: If we return all staged and unstaged changes at the same time, each item should be flagged as staged or not.

clients/vscode/src/chat/webview.ts
```diff
@@ -725,6 +727,74 @@ export class ChatWebview extends EventEmitter {
...
+              break;
+            }
+          }
+
+          return res;
+        };
+
+        const stagedChanges: ChangeItem[] = await getRepoChanges(repositories, true, remainingChars);
+
+        const stagedTokenCount = stagedChanges.reduce((count, item) => count + Math.ceil(item.content.length / 4), 0);
```
icycodes, line 786:
```suggestion
        const stagedCharCount = stagedChanges.reduce((count, item) => count + item.content.length, 0);
```

clients/vscode/src/chat/webview.ts
```diff
@@ -725,6 +727,74 @@ export class ChatWebview extends EventEmitter {
...
+          repos: Repository[],
+          staged: boolean,
+          charLimit: number,
+        ): Promise<ChangeItem[]> => {
+          if (charLimit <= 0) {
+            return [];
+          }
+
+          const res: ChangeItem[] = [];
+          let currentTokenCount = 0;
```
icycodes, line 753:
```suggestion
          let currentCharCount = 0;
```

clients/vscode/src/chat/webview.ts
```diff
@@ -725,6 +727,74 @@ export class ChatWebview extends EventEmitter {
           return [];
         }
       },
+      getChanges: async (params: GetChangesParams): Promise<ChangeItem[]> => {
+        if (!this.gitProvider.isApiAvailable()) {
+          return [];
+        }
+
+        const maxChars = params.maxChars || 4096;
```
icycodes, line 735: Please update this logic according to the API definition. 
>  ```
> /**
>   * The maximum number of characters to return in the change content.
>   * If not provided, all changes will be returned without character limitation.
>   */
> maxChars?: number

icycodes (changes requested):

PR #4035, opened 2025-03-19
feat(chat): log chat event in answer engine

zwpaper:
CleanShot 2025-03-19 at 10 31 36@2x

CleanShot 2025-03-19 at 10 31 50@2x


ee/tabby-db/src/user_events.rs
```diff
@@ -16,6 +16,13 @@ pub struct UserEventDAO {
     pub payload: Vec<u8>,
 }
 
+#[derive(FromRow)]
+pub struct UserChatCompletionDailyStatsDAO {
+    pub start: DateTime<Utc>,
+    pub user_id: i64,
+    pub chats: i32,
+}
+
```
wsxiaoys, line 25: this shall goes to `user_chats.rs` (take `user_completions.rs` as example)

ee/tabby-db/src/user_chats.rs
```diff
@@ -0,0 +1,51 @@
...
+                user_id,
+                COUNT(1) as chats
+            FROM user_events
+            WHERE created_at >= ?1
+              AND created_at < ?2
+              AND kind = 'chat_completion'
+              AND ({no_selected_users} OR user_id IN ({users}))
+            GROUP BY 1, 2
+            ORDER BY 1 ASC
+            "#,
```
wsxiaoys, line 41: image
use strftime instead?

ee/tabby-schema/src/schema/thread.rs
```diff
@@ -50,6 +50,7 @@ pub trait ThreadService: Send + Sync {
     /// Create a new thread run
     async fn create_run(
         &self,
+        user_id: &ID,
```
wsxiaoys, line 53: what is this for?

wsxiaoys:
Otherwise LGTM

ee/tabby-schema/src/schema/thread.rs
zwpaper: pass the user_id to answer service, so that logger could utilize it to log the chat user

wsxiaoys: i recommend just pass `user: &User` (and you can omit user_id and policy).

PR #4072, opened 2025-03-25
feat(client): add optional file type property to ListFileItem interface

Sma1lboy:
Adding ListFileItem as optional for client-side backward compatibility. If there is no type passed in previous client versions, it will remain undefined as workspace

### DIFF
--- a/clients/tabby-chat-panel/src/client.ts
+++ b/clients/tabby-chat-panel/src/client.ts
@@ -206,6 +206,12 @@ export interface ListFileItem {
    * The filepath of the file.
    */
   filepath: Filepath
+
+  /**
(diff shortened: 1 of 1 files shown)


clients/tabby-chat-panel/src/client.ts
```diff
@@ -206,6 +206,12 @@ export interface ListFileItem {
    * The filepath of the file.
    */
   filepath: Filepath
+
+  /**
+   * Type of the file, it could be a tab or a file from workspace, if the file didn't pass the type, it will be default from the workspace.
+   * The type is used to determine whether the file is opened in the editor or not.
+   */
+  type?: 'tab' | 'workspace'
```
icycodes, line 214:
```suggestion
  source?: 'openedInEditor' | 'searchResult'
```

PR #4083, opened 2025-03-28
feat(vscode): enhance mention parsing by introducing MentionType and …

Sma1lboy:
still working

### screenshots

image

image

image

### DIFF
--- a/clients/vscode/src/findSymbols.ts
+++ b/clients/vscode/src/findSymbols.ts
@@ -25,21 +25,16 @@ export async function listSymbols(uri: Uri, query: string, limit?: number): Prom
   try {
     const maxResults = limit ?? 50;
     const queryString = query.trim().toLowerCase();
-
-    // Get document symbols from the specified file
     const rawDocumentSymbols =
       (await commands.executeCommand<DocumentSymbol[] | SymbolInformation[]>(
         "vscode.executeDocumentSymbolProvider",
         uri,
       )) || [];
 
-    // Convert DocumentSymbol[] to SymbolInformation[] if needed
     const documentSymbols = convertToSymbolInformation(rawDocumentSymbols);
 
-    // Filter symbols from the current file only
     const filteredSymbols = filterSymbols(documentSymbols, queryString, maxResults);
 
-    // Convert to result format with appropriate icons
     return filteredSymbols.map((symbol) => ({
       name: symbol.name,
       kind: symbol.kind,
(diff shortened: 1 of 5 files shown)


wsxiaoys:
cc @zhanba

clients/vscode/src/findSymbols.ts
```diff
@@ -0,0 +1,197 @@
...
+    if (symbol.children && symbol.children.length > 0) {
+      flattenDocumentSymbols(symbol.children, fullName, uri, result);
+    }
+  }
+}
+
+/**
+ * Maps symbol kinds to appropriate theme icons
+ */
+function getSymbolIcon(kind: SymbolKind): ThemeIcon {
```
icycodes, line 97: Consider using a map:
```
const symbolIconMap = new Map<SymbolKind, string>([
  [SymbolKind.File, "file"],
  // ...
]);
const icon = new ThemeIcon(symbolIconMap.get(key) ?? "symbol-misc");
```

clients/vscode/src/inline-edit/quickPick.ts
```diff
@@ -62,27 +98,139 @@ export class UserCommandQuickpick {
...
+      this.showContextPicker();
     } else {
       this.updateQuickPickList();
       this.updateQuickPickValue(this.quickPick.value);
     }
   }
 
+  private async showContextPicker() {
+    const currentValue = this.quickPick.value;
+    const mentionText = currentValue.substring(currentValue.lastIndexOf("@") + 1).trim();
```
zhanba, line 112: - Is this `mentionText` the same as `mentionQuery` on `handleValueChange` ? 
- Can we extract the ContextPicker to a separate class?

Sma1lboy: > Is this mentionText the same as mentionQuery on handleValueChange ?

yes, we should just use the mentionQuery on inputParseResult

>  Can we extract the ContextPicker to a separate class?

Extracting the ContextPicker to a separate class would likely not be beneficial since it has high functional coupling with `UserCommandQuickpick`, is completely lifecycle-dependent on its parent class, and lacks reuse potential due to its highly specific purpose in the command input context selection flow. It might be a good idea to consider refactoring and splitting the code when there's a future need to reuse this code snippet.

zhanba (commented):
LGTM

icycodes (approved):

PR #4088, opened 2025-03-31
feat(vscode): remove generateBranchName command and add branch check …

Sma1lboy:
https://jam.dev/c/efe8603a-8572-4781-ad59-28baee958612


clients/vscode/src/commands/index.ts
```diff
@@ -422,8 +422,20 @@ export class Commands {
             { repository: selectedRepo.rootUri.toString() },
             token,
           );
+
           if (result && selectedRepo.inputBox) {
             selectedRepo.inputBox.value = result.commitMessage;
+
+            if (this.gitProvider.isApiAvailable()) {
+              const repo = this.gitProvider.getRepository(selectedRepo.rootUri);
```
icycodes, line 430: `this.gitProvider.isApiAvailable()` is not necessary here, `selectedRepo` is already of type Repository, so there is no need to get it again.

PR #4120, opened 2025-04-07
fix(registry): explicitly specify a five-second timeout for downloadi…

zwpaper:
…ng registry

context: https://tabbyml.slack.com/archives/C05CWLZ0Y85/p1743307225964259

### DIFF
--- a/crates/tabby-common/src/registry.rs
+++ b/crates/tabby-common/src/registry.rs
@@ -40,10 +40,9 @@ fn models_json_file(registry: &str) -> PathBuf {
 }
 
 async fn load_remote_registry(registry: &str) -> Result<Vec<ModelInfo>> {
-    // Create a client with custom timeout.
-    // There have been instances where requests did not adhere to the default timeout,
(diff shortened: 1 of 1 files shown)


crates/tabby-common/src/registry.rs
```diff
@@ -40,15 +40,27 @@ fn models_json_file(registry: &str) -> PathBuf {
...
-    ))
-    .await
-    .context("Failed to download")?
-    .json()
-    .await
-    .context("Failed to get JSON")?;
+    // Create a client with custom timeout.
+    // There have been instances where requests did not adhere to the default timeout,
+    // resulting in requests hanging indefinitely.
+    // Therefore, it's necessary to specify a custom timeout.
```
wsxiaoys, line 46:
```suggestion
    // Create an HTTP client with a custom timeout.
    // This is necessary because the default timeout settings can sometimes cause requests to hang indefinitely.
    // To prevent such issues, we specify a custom timeout duration.
```

wsxiaoys (approved):

PR #4267, opened 2025-05-27
(docs)model registry: remove code models from chat list

loleg:
If you try to use the code models for chat completion, you get an error:

```
Message:  called `Result::unwrap()` on an `Err` value: Model validation has failed. For TabbyML models, please consult https://github.com/tabbyml/registry-tabby to locate the appropriate models.

Caused by:
    Model 'Qwen2.5-Coder-7B-Instruct' is not a completion model; it does not have a prompt template.
Location: /root/workspace/crates/tabby-download/src/lib.rs:207
```

These "Coder" models should not be recommended as chat models. I'd be happy to see the list generally updated, with some new recommendations. Is there a good process to do this?


loleg:
Sorry, I think I misunderstood your error message and it's a different issue. 

It would still be helpful if the docs would explain whether a Coder-style model is appropriate for the chat, under which circumstances, and how it affects system requirements.

PR #4285, opened 2025-06-13
feat(ssh-git): Introduce support to pull repositories using ssh keypairs

leophys:
Hi folks! I love what you are building and wanted to contribute back. I see a use case where someone (me?) would like to pull private repositories. As of now, this is supported only for forges which accept authentication mechanisms through HTTP. I wanted to support pulling via ssh using ssh keys.

Doing so, I also refactored the code to drop invoking the `git` binary and instead completely rely on `git2-rs`, which is already a dependency you were pulling in.

Currently, this PR only includes the backend work. There is also the need for a UI change to make this available for end users, and here I stopped. I did because I understand this is a product, and you might want to have a discussion (open or not) on how (whether) to expose this feature to the users.

I designed the change to allow both keys stored in files and as text in the database. This seemed like a sensible and easy thing to do.

There are three new tests:
 - one to test pulling a public repository using the new code path (that leverages only `git2-rs`)
 - two more to try and pull private repository either providing a path to a private key or the private key itself via env variables (these are skipped if the needed environment variables are not set)

I am open for feedback/discussions. I am also keen to do the frontend work, if we agree on a final UX/UI.

### DIFF
--- a/crates/tabby-index/src/code/repository.rs
+++ b/crates/tabby-index/src/code/repository.rs
@@ -9,11 +9,10 @@ use git2::{
     build::{CheckoutBuilder, RepoBuilder},
     Cred, FetchOptions, RemoteCallbacks, Repository,
 };
-use tabby_common::path::repositories_dir;
+use tabby_common::{config::SshKeyPair, path::repositories_dir};
 use tracing::warn;
 
 use super::CodeRepository;
-use tabby_common::config::SshKeyPair;
 
 trait RepositoryExt {
     fn sync(&self) -> anyhow::Result<String>;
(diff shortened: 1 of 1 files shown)


leophys:
I don't know how to ask for a review to a team member, so I am just tagging someone, @gyxlucy

leophys:
Hi there! :wave: 
Is there any bandwidth to deal with this? I would really like to add this feature :pray:

wsxiaoys:
add @zwpaper

mexmer:
question to this, if this will be setup trough trough config.toml, how do we will supply the kyes? only thought environment variable, or do you plan any other way? for example storing in .ssh_keys located besides config.toml

torrefatto:
> question to this, if this will be setup trough trough config.toml, how do we will supply the kyes? only thought environment variable, or do you plan any other way? for example storing in .ssh_keys located besides config.toml

It's quite some time I did this, so forgive me but my memory is fuzzy. IIRC, the idea is that the keys would be stored in the database. What is missing from this PR is an API (and related frontend changes, that I mention in the description) to manage those keys. Again, IIRC, the main API of this application is graphql, and I was waiting for feedback from the team before implementing it.

mexmer:
thanks for answer,
will be actually nice, if there was config.toml support

since you can use ssh uri in config.toml, just not with the keys (or at least i have not figured out how, and there is no mention in documentation either)

Issue #4291, opened 2025-06-18
Localhost mismatch


Ari1009:
**Describe the bug**
The localhost URL shown in the address bar does not match the endpoint URL displayed in the "UI" tab. 

**Information about your version**
 Tabby version: 0.29.0

**Additional context**

![Image](https://github.com/user-attachments/assets/b37d384a-cfe4-48fd-b6ba-faf2b1307734)

PR #4292, opened 2025-06-18
fix: ensure endpoint URL matches browser address

Ari1009:
Fixes: #4291  

- Updated the useExternalURL hook to use the browser's current URL (window.location.origin) by default.
- This fixes issues where the app showed the wrong URL in  the UI.

250618_15h45m25s_screenshot

### DIFF
--- a/ee/tabby-ui/lib/hooks/use-network-setting.tsx
+++ b/ee/tabby-ui/lib/hooks/use-network-setting.tsx
@@ -28,6 +28,7 @@ const useExternalURL = () => {
     return networkSetting?.externalUrl || ''
   }, [networkSetting])
 }
+
 function getOrigin() {
   if (isClientSide()) {
     return new URL(window.location.href).origin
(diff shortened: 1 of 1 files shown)


Ari1009:
@zwpaper can i get a review

zwpaper:
hi @Ari1009, thanks for the contribute, I have confirmed with maintainers, the `window.location.origin` is not always precise, may be wrong in some cases.

for changing this url, please update it here:
CleanShot 2025-06-18 at 23 45 39@2x

and welcome to create a PR for documenting it

Ari1009:
> hi @Ari1009, thanks for the contribute, I have confirmed with maintainers, the `window.location.origin` is not always precise, may be wrong in some cases.
> 
> for changing this url, please update it here: CleanShot 2025-06-18 at 23 45 39@2x
> 
> and welcome to create a PR for documenting it

Ah, I assumed during the initial setup that the external URL would be something to set programmatically, so I missed the fact that it’s configurable directly from the settings UI
Thnx for clarifying... will go ahead and close pr then.Will give a small doc update after a while.
Just had a small question where I should update in the documentation.

zwpaper:
I found a doc talking about this, it turn out to be a failed action making it missed from the doc site, maybe you could help to resolve the action issue here: https://github.com/TabbyML/tabby/actions/runs/15125003196/job/42515346951

and the doc located at: https://github.com/TabbyML/tabby/blob/main/website/docs/administration/general-settings/index.mdx

PR #4310, opened 2025-07-08
fix(website): fix build fail for website

zwpaper:


Ari1009:
@zwpaper  there were some build warnings popping up because of broken anchor links, like the one pointing to #code-source-repositories  it just needed to match the actual header id. Also tweaked the Tally form integration a bit, since using anchor links there was causing warnings during the build (even though everything still worked fine).
No changes to how anything works  just cleaned up the warnings to keep things tidy.

PR #4323, opened 2025-07-17
feat(model): Expose the Flash Attention LLAMA flag as an environment variable

kprinssu:
This PR contains changes to expose the embedded Llama.cpp flash attention flag as an environment variable. The intent behind this PR is to expose fine grain control of how models are loaded and run. 

I am proposing to expose this via a new environment variable called `LLAMA_CPP_FAST_ATTENTION`. The intended behaviour is as follows:

- if not set, then the `fa` flag is not enabled
- if the `LLAMA_CPP_FAST_ATTENTION` is set to any value, then the `fa` flag is enabled

### DIFF
--- a/crates/tabby/src/main.rs
+++ b/crates/tabby/src/main.rs
@@ -100,6 +100,7 @@ fn to_local_config(model: &str, parallelism: u8, device: &Device) -> ModelConfig
     } else {
         0
     };
+    // This only works when the the model is included in cli arguments.
     let enable_fast_attention = Some(std::env::var("LLAMA_CPP_FAST_ATTENTION").is_ok());
 
     ModelConfig::new_local(model, parallelism, num_gpu_layers, enable_fast_attention)


crates/tabby/src/main.rs
```diff
@@ -100,6 +100,7 @@ fn to_local_config(model: &str, parallelism: u8, device: &Device) -> ModelConfig
     } else {
         0
     };
+    let enable_fast_attention = Some(std::env::var("LLAMA_CPP_FAST_ATTENTION").is_ok());
```
zwpaper, line 103:
```suggestion
    // This only works when the the model is included in cli arguments.
    let enable_fast_attention = Some(std::env::var("LLAMA_CPP_FAST_ATTENTION").is_ok());
```

zwpaper (commented):
Thanks for improving this, this env would only works when we use the model in cli args, so please add a comment to remind this.

otherwise LGTM

PR #4331, opened 2025-08-01
chore(ci): fix download llama from upstream and drop cuda11.7 release

zwpaper:


zwpaper:
Linux verified, will verify Windows in RC

Issue #4374, opened 2025-10-04
Support for codestral/embedding API kind


Silzinc:
**Please describe the feature you want**

Mistral AI made an [embedding API](https://docs.mistral.ai/capabilities/embeddings/code_embeddings/) for `codestral-embed` (actually they made [another one](https://docs.mistral.ai/capabilities/embeddings/text_embeddings/) for `mistral-embed`, but I don't expect it to be any relevant compared to `codestral-embed` for this project). This contradicts the [tabby docs](https://tabby.tabbyml.com/docs/references/models-http-api/mistral-ai/) if I'm not mistaken.

Would it be possible to support a new `codestral/embedding` (or `mistral/embedding`) API kind for this?

**Additional context**

I stumbled on this while trying to configure `codestral-embed` in tabby.

Here is my current (failing) config for this model:
```toml
[model.embedding.http]
kind = "openai/embedding"
model_name = "codestral-embed"
api_endpoint = "https://api.mistral.ai/v1"
api_key = "my-key"
```
Here, I have been using the `openai/embedding` API although it is not exectly the one expected by the model. Upon connection, I obtain this error:
```
Error 422: {"detail":[{"type":"model_attributes_type","loc":["body"],"msg":"Input should be a valid dictionary or object to extract fields from","input":"{\"input\":[\"hello Tabby\"],\"model\":\"codestral-embed\"}"}]}
```
Besides, this works perfectly fine in my terminal:
```bash
curl -X POST "https://api.mistral.ai/v1/embeddings" \
           -H "Content-Type: application/json" \
           -H "Authorization: Bearer $MISTRAL_API_KEY" \
           -d "{\"input\": [\"hello Tabby\"], \"model\": \"codestral-embed\"}"
```

---
Thank you in advance!
Please reply with a 👍 if you want this feature.

Silzinc:
Solved in #4380.

PR #4380, opened 2025-10-11
feat(model): add support for codestral/embedding api kind

Silzinc:
Trying to solve #4374. I was able to get it to work with the following config:
```toml
[model.embedding.http]
kind = "codestral/embedding"
model_name = "codestral-embed" # actually ignored, hardcoded
api_endpoint = "https://api.mistral.ai/v1" # actually ignored, same as above
api_key = "my-key"
```
I don't know what test layers do you use internally (I only checked that the model was "connected" on the webapp). I'll just leave this here.

### DIFF
--- a/crates/http-api-bindings/src/embedding/mistral.rs
+++ b/crates/http-api-bindings/src/embedding/mistral.rs
@@ -44,9 +44,10 @@ impl MistralEmbeddingEngine {
     ) -> Box<dyn Embedding> {
         Box::new(Self {
             client: Arc::new(Client::new()),
-            api_endpoint: api_endpoint
-                .unwrap_or("https://api.mistral.ai/v1/embeddings")
-                .to_owned(),
+            api_endpoint: format!(
+                "{}/embeddings",
+                api_endpoint.unwrap_or("https://api.mistral.ai/v1")
+            ),
             model_name: model_name.unwrap_or("codestral-embed").to_owned(),
             api_key: api_key.unwrap_or_default().to_owned(),
         })


crates/http-api-bindings/src/embedding/codestral.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use anyhow::Result;
+use async_trait::async_trait;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use tabby_inference::Embedding;
+
+/// `AzureEmbeddingEngine` is responsible for interacting with Azure's Embedding API.
+///
+/// **Note**: Currently, this implementation only supports the OpenAI API and specific API versions.
```
zwpaper: should be MistralEmbeddingEngine

crates/http-api-bindings/src/embedding/codestral.rs
```diff
@@ -0,0 +1,85 @@
...
+            api_key: api_key.unwrap_or_default().to_owned(),
+        })
+    }
+}
+
+#[async_trait]
+impl Embedding for CodestralEmbeddingEngine {
+    async fn embed(&self, prompt: &str) -> Result<Vec<f32>> {
+        // Clone all necessary fields to ensure thread safety across await points
+        let api_endpoint = "https://api.mistral.ai/v1/embeddings".to_string();
```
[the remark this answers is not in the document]
Silzinc: I'll do that for both the \api_endpoint` and the `model_name`.

crates/http-api-bindings/src/embedding/codestral.rs
Silzinc: I admit that file was copy pasted. I will fix that.

Silzinc:
Thank you @zwpaper for your review. I tried to implement the changes you suggested in a second commit. May you have a look.

Also, I made a mistake in the config, the api_endpoint should be "https://api.mistral.ai/v1/embeddings" now that it isn't hardcoded anymore.

zwpaper:
Hi @Silzinc, thanks for the update!
the PR now LGTM mostly except one nit.

we use only the url prefix(without the `embeddings` suffix) in other model providers, like OpenAI embedding:

https://github.com/TabbyML/tabby/blob/01f8bb72de75221ebaa18247a6a727f494421d05/crates/http-api-bindings/src/embedding/openai.rs#L26

for consistence, please also construct the full url during the creation of MistralEmbeddingEngine.

Silzinc:
Hello @zwpaper,
I think it should be solved now.

PR #4425, opened 2026-01-20
feat: sunset knowledge and chat features

liangfung:
## Summary
- Disable chat feature flag in useIsChatEnabled
- Remove Pages/Knowledge and Search Pages from feature list
- Redirect /pages and /search to home
- Remove Chat statistics from DailyCharts
- Remove ThreadFeeds from home page
- Remove New Page option from UserPanel

## Test plan
- Verify chat is disabled
- Verify /pages and /search redirect to home
- Verify UI elements related to chat and knowledge are removed

## Screenshot
image

🤖 Generated with [Pochi](https://getpochi.com)

### DIFF
--- a/ee/tabby-ui/app/(dashboard)/experiments/components/feature-list.tsx
+++ b/ee/tabby-ui/app/(dashboard)/experiments/components/feature-list.tsx
@@ -1,10 +1,17 @@
 'use client'
 
-import { useEnableDeveloperMode } from '@/lib/experiment-flags'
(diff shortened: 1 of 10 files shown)


zwpaper:
the pr LGTM, only have one question, should we add a flag to toggle those feature so that the original users can gain the previous experience if their want to?

But it may require extra effort to maintain compatibility.

WDYT @wsxiaoys

PR #4481, opened 2026-04-10
fix: propagate error instead of panicking on non-UTF-8 reference name in checkout

kuishou68:
## Problem

In `crates/tabby-index/src/code/repository.rs`, the call to `reference.name().unwrap()` can panic at runtime. `reference.name()` returns `Option<&str>`, which is `None` when the reference name contains non-UTF-8 bytes. Since `git2` exposes raw bytes for ref names, this is a valid (if rare) code path that causes an unhandled panic.

## Fix

Replace `.unwrap()` with `.ok_or_else()` to convert the `Option` into a `Result` and propagate the error through `anyhow`, which the function already returns:

```rust
// Before
repo.set_head(reference.name().unwrap())?;

// After
let ref_name = reference
    .name()
    .ok_or_else(|| anyhow::anyhow!("Reference name is not valid UTF-8"))?;
repo.set_head(ref_name)?;
```

This avoids the panic and allows callers to handle the error gracefully.

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

### DIFF
--- a/crates/tabby-index/src/code/repository.rs
+++ b/crates/tabby-index/src/code/repository.rs
@@ -109,7 +109,10 @@ pub fn checkout(repository: &CodeRepository, branch: &str) -> anyhow::Result<()>
     let mut checkout_builder = git2::build::CheckoutBuilder::new();
     checkout_builder.force();
 
-    repo.set_head(reference.name().unwrap())?;
+    let ref_name = reference
+        .name()
+        .ok_or_else(|| anyhow::anyhow!("Reference name is not valid UTF-8"))?;
+    repo.set_head(ref_name)?;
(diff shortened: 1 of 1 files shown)


kortez-io:
Replacing `unwrap()` with proper error propagation on `reference.name()` is good Rust practice — a panic in index building would crash the whole repository indexing job rather than just skipping the invalid reference. Using `ok_or(...)` to convert the `Option` to a `Result` and propagating with `?` is the right pattern. Out of curiosity, is non-UTF8 in reference names something that comes up with specific git hosting providers, or is it more of an edge case with local repos?

PR #4487, opened 2026-04-28
fix(tabby-agent): only send tabby/chat feature registration to VS Code LSP Clients

### LINKED ISSUE #4486: Neovim LSP Client Crashing on unhandled tabby/chat feature registration
**Describe the bug**
Neovim 0.12.0+ has been failing with recent versions of tabby-agent. 

```
     Error  08:16:00 AM msg_show.echoerr LSP[tabby]: Error SERVER_REQUEST_HANDLER_ERROR: "...wrapped-0.12.1/share/nvim/runtime/lua/vim/lsp/client.lua:969: table index is nil"  
```

Which caused vim-tabby / autocompletion to stop working. 

Please provide output of `tabby --version`

This is an issue inside the tabby-agent TS LSP server code, not the core rust application. I was seeing the issue with the latest `1.8.0` npx `tabby-agent` package. 

The issue looks to be related to the ChatFeature functionality which is specific to VSCode clients.

There could be an argument made that some of this responsibility falls on neovim to not crash for an unknown handler? I'm not well versed on the politics.

**Additional context**
I have a local branch with a solution. It involves exposing `clientInfo` from the `tabbyApiClient` and an extracted `getClientType` utility from `telemetry.ts`. Then inside ChatFeature, only send `connection.client.register(ChatFeatures.type)` when the client type is "vscode". I am working my way through the contribution guidelines. 

Edit: Forgot to call out the corresponding issue someone else opened in [vim-tabby](https://github.com/TabbyML/vim-tabby/issues/40).

datGryphon:
## Summary

Fixes Neovim 0.12 crash where `tabby/chat` registration sent via `client/registerCapability` caused a LuaJIT panic (`table index is nil`) because it is a VS Code-specific method not recognized by Neovim's LSP client.

## Changes

- `src/utils/clientType.ts` - Extracted `getClientType(clientInfo)` from `telemetry.ts` into a shared utility with a `ClientType` union type (`"vscode" | "intellij" | "vim" | "unknown"`)

- `src/telemetry.ts` - Use the shared `getClientType` instead of the private method

- `src/http/tabbyApiClient.ts` - Store and expose `clientInfo` via `getClientInfo()` getter (previously only used it to build `userAgentString`)

- `src/chat/index.ts` - `syncFeatureRegistration()` gates `tabby/chat` registration behind `getClientType(...) === "vscode"`

## Verification

- Builds, all 84 tests pass, lint clean
- Tested locally with Neovim 0.12, completions working again through vim-tabby

## Related

Fixes #4486, vim-tabby[#40](https://github.com/TabbyML/vim-tabby/issues/40)

### DIFF
--- a/.changes/unreleased/Fixed and Improvements-20260428-092408.yaml
+++ b/.changes/unreleased/Fixed and Improvements-20260428-092408.yaml
@@ -1,0 +1,3 @@
+kind: Fixed and Improvements
+body: Do not send tabby/chat LSP Feature to clients that cannot support it
+time: 2026-04-28T09:24:08.662286918-04:00
--- a/clients/tabby-agent/src/chat/index.ts
+++ b/clients/tabby-agent/src/chat/index.ts
@@ -4,6 +4,7 @@ import type { ServerCapabilities } from "../protocol";
 import type { Feature } from "../feature";
 import type { TabbyApiClient } from "../http/tabbyApiClient";
 import { ChatFeatures } from "../protocol";
+import { getClientType } from "../utils/clientType";
 
 export class ChatFeature extends EventEmitter implements Feature {
   private isApiAvailable = false;
@@ -41,7 +42,7 @@ export class ChatFeature extends EventEmitter implements Feature {
   }
 
   private async syncFeatureRegistration(connection: Connection) {
-    if (this.isApiAvailable) {
+    if (this.isApiAvailable && getClientType(this.tabbyApiClient.getClientInfo()) === "vscode") {
       if (!this.featureRegistration) {
         try {
           this.featureRegistration = await connection.client.register(ChatFeatures.type);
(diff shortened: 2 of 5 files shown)


datGryphon:
I'm also looking at if/how this should be addressed at the neovim lsp client level. I'm still investigating the source of the regression on that end. 

Anyway it feels like a two sided problem. Something is being sent that is unsupported, and something unsupported is not being handled properly. ¯\\\_(ツ)_/¯

datGryphon:
Bumping this with the added detail that this was addressed on the [neovim side](https://github.com/neovim/neovim/issues/39491) after I submitted an issue, but the potential for this to be reintroduced will persist while tabby-agent continues to send `tabby/chat` to lsp clients which are known to be unable to support it. 

Edit: clarification, grammar

wsxiaoys:
Thanks for the PR - LG from my side

rudiservo:
seems ok, when can we roll this out?

PR #4497, opened 2026-06-02
fix: handle Windows file:// URL with three slashes in resolve_dir

### LINKED ISSUE #4104: Context provider sheduled jobs failing on Windows
## Bug description
I'm using `Git` Context provider with `Git URL` parameter set to local folder as `file://C:\repos\myproject`
Initial indexing job finished with success. Manual job execution via Run button in ui also goes well.
But sheduled indexing jobs that executes every hour - all failing with same error.
> <time> [INFO]: Building source code index: file:///C:/repos/myproject
> <time> [WARN]: Failed to refresh code index: Directory /C:/repos/myproject does not exist
> <time> [INFO]: Job failed Directory /C:/repos/myproject does not exist

### Info about my tabby version and setup
**Tabby:**
- tabby version : `tabby 0.26.0`
- tabby release asset: `tabby_x86_x64-windows-msvc-cuda124.zip`

**GPU:**
- gpu name: `NVIDEA GeForce RTX 4050 Laptop GPU`
- gpu driver version: `566.26`
- cuda version: `12.7`

**OS:**
- os edition: `Windows 10 Enterprise LTSC`
- os version: `21H2`
- os build: `19044.5608`

## Additional context
I think there's some kind of windows\unix path conflict...
The error appears in this file `crates/tabby-index/src/code/repository.rs` at line 86.
But since initial and manual runs goes well I think there are error in sheduling module or somewhere else that handles repo path as unix path starts with `/`.
I'm not a rust dev and don't have much free time to dive into your codebase so I hope you'll be able to fix this issue.

In my opinion for linux file path should be `file:///home/repos/myproject` or so and for windows it should be `file://C:/repos/myproject` or `file://C:\repos\myproject` without `/` after `file://`

I also tried to init git context provider using next strings as `Git URL` 
- `file://C:\repos\myproject`
- `file://C:\\repos\\myproject` 
- `file://C:/repos/myproject`

And I remember correctly(I did it like a week ago) all worked with manual runs but `/` transorms into `\\` and `\\` transofms into `\\\\` in tabby logs while `\` transforms into `/` ...

1795771535y-cell:
## Problem

When a file:// URL is canonicalized by RepositoryConfig::canonicalize_url(), it becomes file:///C:/path on Windows (three slashes per RFC 8089). The fallback strip_prefix("file://") in resolve_dir() leaves a leading slash, producing /C:/repos/... which is invalid on Windows. This causes scheduled index jobs to fail with "Directory does not exist".

**Steps to reproduce:**
1. Configure a Git context provider with file://C:/repos/project on Windows
2. Initial/manual indexing works fine
3. Scheduled hourly indexing fails because the path becomes /C:/repos/project

## Fix

Try strip_prefix("file:///") first (three slashes for absolute paths), then fall back to strip_prefix("file://") (two slashes for relative/host-based paths). This correctly handles both canonicalized and raw Windows file:// URLs.

## Testing

The existing test suite covers both formats. No test changes needed as existing ile:///C:/ test cases already produce the correct PathBuf.

Closes #4104

### DIFF
--- a/crates/tabby-common/src/config.rs
+++ b/crates/tabby-common/src/config.rs
@@ -201,7 +201,7 @@ impl RepositoryConfig {
                 .ok()
                 .and_then(|url| url.to_file_path().ok())
                 .unwrap_or_else(|| {
-                    let path = git_url.strip_prefix("file://").unwrap_or(git_url);
+                    let path = git_url.strip_prefix("file:///").unwrap_or_else(|| { git_url.strip_prefix("file://").unwrap_or(git_url) });
                     PathBuf::from(path)
                 })
         } else {


1795771535y-cell:
Closing: no maintainer response after extended wait. Thanks for the opportunity.

PR #4532, opened 2026-08-22
fix(vim): pass position encoding to Neovim

### LINKED ISSUE #4527: Neovim 0.11 warning because position encoding is omitted from make_position_params
**Describe the bug**

  When vim-tabby requests an inline completion on Neovim 0.11, Neovim displays the following warning:

  `position_encoding param is required in vim.lsp.util.make_position_params. Defaulting to position encoding of the first client.`

  The warning appears because `lua/tabby/lsp/nvim_lsp.lua` constructs the inline-completion parameters without specifying a position encoding:

  `local inline_completion_params = vim.lsp.util.make_position_params()`

  Besides producing the warning, defaulting to the first LSP client could select the wrong encoding when multiple clients are attached to the buffer.

  tabby 0.32.0

  vim-tabby commit:

  5e30c429354b8e0fd5bbc7634be4abac5518bb80

  Neovim version:

  NVIM v0.11.7
  Build type: Release
  LuaJIT 2.1.1741730670

  GPU: NVIDIA GeForce RTX 3080 Ti
  VRAM: 12 GB
  NVIDIA driver: 610.43.03
  CUDA version: 13.3

tandede:
## Summary

- fetch the active Tabby LSP client before building inline-completion parameters
- pass that client offset encoding to vim.lsp.util.make_position_params
- return early when the client is no longer available

This removes the Neovim 0.11 position-encoding warning and avoids selecting the encoding of an unrelated first client when several LSP clients are attached.

## Test plan

- luac -p clients/vim/lua/tabby/lsp/nvim_lsp.lua
- isolated Lua regression scenarios for an active and a missing client
- verified with Neovim 0.10.4 and 0.11.7

Fixes #4527


rudiservo:
this works on my local setup, the only thing I would point out is that I seem to get no error if the config.toml is not configured, but that is maybe another issue.

rudiservo (approved):
It works, seems ok, the major difference is the
`make_position_params(0, client.offset_encoding)`

Code seems cleaner like this.

